PR #137, opened 2017-11-28
added support for byte conversion

rohanmars:


runtime/core/src/main/java/io/atlasmap/converters/StringConverter.java
```diff
@@ -50,15 +49,23 @@ public Boolean convertToBoolean(String value) throws AtlasConversionException {
...
     public Byte convertToByte(String value) throws AtlasConversionException {
         if (value == null) {
             return null;
         }
-        throw new AtlasConversionException(new AtlasUnsupportedException("String to Byte conversion is not supported"));
+        if (value.length() == 1 && value.charAt(0) <= Byte.MAX_VALUE) {
+            return (byte) value.charAt(0);
+        } else {
+            if (value.length() > 1) {
+                throw new AtlasConversionException("String is greater then one character");
```
apupier: typo then -> than

runtime/core/src/test/java/io/atlasmap/converters/BooleanConverterTest.java
```diff
@@ -58,9 +57,12 @@ public void convertToBooleanNull() throws Exception {
         assertNull(t2);
     }
 
-    @Test(expected = AtlasConversionException.class)
+    @Test
     public void convertToByte() throws Exception {
-        converter.convertToByte(Boolean.TRUE);
+        Byte trueValue = (byte) 1;
+        assertEquals(trueValue, converter.convertToByte(Boolean.TRUE));
+        Byte falseValue = (byte) 0;
```
apupier: maybe worth adding more complex byte such as (byte) 0.1, Given the current code it should return True (and seems to be the expected result from classical Java implementations) What do you think?

Issue #132, opened 2017-11-27
Support multiple source/target Document


igarashitm:
Support multiple source/target Document at runtime side. As it currently assumes both of source and target are single object everywhere in the code, it involves major refactoring in the `DefaultAtlasContext` and modules. `AtlasModule` SPI needs to be reconstructed to consider per field level invocation as both of source and target could be a mixup from multiple modules.

PR #145, opened 2017-12-04
feat: Support multiple source/target Document

igarashitm:
Fixes: #132


igarashitm:
Hi @atlasmap/runtime ,

This is adding multiple source/target support into atlasmap/runtime, which involves major refactoring around `DefaultAtlasSession`, `DefaultAtlasContext` and `AtlasModule`s.

I pulled up the SEPARATE/COMBINE handling from each modules into `DefaultAtlasContext` in order to support a mixup from different modules for a single mapping entry. Also this introduces `AtlasInternalSession`, which holds `AtlasFieldReader`/`AtlasFieldWriter` and `Head`.

`Head` is a reading head pointing to the mapping entry, so each module can trace which pair of `SourceField` and `TargetField` are being processed via session.

Any feedback would be appreciated.

igarashitm:
It was assuming each of source/target module is single item, i.e. only supports same format like multiple java source. And the module were shared across all the source documents, you can't cache FieldReader there. It never worked if there're java source and json source.

Issue #141, opened 2017-11-30
XmlFieldReader test fails with the method really in use


igarashitm:
There're some problems here
* `XmlFieldReader.readNew()` is the method really in use by xml module, others are just invoked from unit tests
  * Plus there's no unit test running against `XmlFieldReader.readNew()`...
* following tests fail with `XmlFieldReader.readNew()`, so works only with the method which is not really in use
  * testReadDocumentElementWithMultipleNamespaceComplex
  * testReadDocumentWithMultipleNamespaceComplex
  * testReadDocumentElementWithMultipleNamespaceComplexConstructorArg

This popped up while I'm working on #132, I'll add `@Ignore` on these tests for now so we can address later.

igarashitm:
`testReadDocumentWithMultipleNamespaceComplex` passed, but other 2 are still failing.

igarashitm:
@mattrpav would you assign somebody for this?

Issue #11, opened 2017-11-15
xml: Migrate XmlPathCoordinate to use AtlasPath


igarashitm:
@mattrpav commented on [Thu Jul 27 2017](https://github.com/atlasmap/atlasmap-runtime/issues/111)

The Xml module uses XmlPathCoordinate and should be migrated to the common AtlasPath handler, or an extension of the AtlasPath class that supports namespaces

mattrpav:
@igarashitm please review and approve the suggested requirements. I've ramped up Sunil this morning on AtlasMap and this task specifically. He'll be ready to start working it shortly.

1. XmlPath class extends from AtlasPath
2. Adds support for an optional namespace modifier on a path segment (no other formats have a concept of a segment modifier at this time)
3. Refactor Xml Module / Service(s) to use XmlPath class

igarashitm:
@mattrpav sounds good! do you think is it worth aligning JSON module as well from future maintenance cost point of view?

mattrpav:
@igarashitm good deal, I'll work on transitioning this task to Sunil w/ the requirements as described. 

Regarding JSON module-- yes, a quick look it appears it is using the legacy PathUtil class instead of AtlasPath. I'll create another ticket and assign it to Sunil to do after this one.

Thanks!

PR #146, opened 2017-12-05
Issues 11.135,136,141

sunilmeddr:


igarashitm:
Hi @sunilmeddr ,

Can you describe what has been done specifically in the commit log so we can trace later? Also you can put `Fixes: #11` to close the issue automatically when it's merged. If you still want to keep some of them open, then just mention with hash. like this (just an example, I haven't checked if it resolves #141 or not):
```
fix: Migrate XmlPathCoordinate and PathUtil into AtlasPath, added XmlPath with extending AtlasPath
fix: something else you've done...

Fixes: #11
Fixes: #135
Fixes: #136
Also Addresses: #141
```

mattrpav:
Changes LGTM

+1 update the commit comments and I think its good to merge

@igarashitm thoughts?

sunilmeddr:
Commit message is updated

runtime/core/src/main/java/io/atlasmap/core/PathUtil.java
```diff
@@ -1,483 +0,0 @@
...
-
-    public static boolean isMapSegment(String segment) {
-        return isCollectionSegment(segment) && segment.contains(PATH_MAP_START);
-    }
-
-    public static String findNextSegment(String segment, String path) {
-        PathUtil pathUtil = new PathUtil(path);
-        String parentSegment = null;
-        for (String currentSegment : pathUtil.segments) {
-            if (parentSegment == null || !segment.equals(parentSegment)) {
```
igarashitm: Yeah I think Codacy is right here, this loop looks super weird. @sunilmeddr I know it was just copied, but can you turn over the condition and just return in 'if'?
https://www.codacy.com/app/atlasmapio/atlasmap/pullRequest?prid=1131462

runtime/core/src/main/java/io/atlasmap/core/AtlasPath.java
```diff
@@ -230,44 +317,170 @@ public String getCollectionSegment() {
...
+
+    public static boolean isMapSegment(String segment) {
+        return isCollectionSegment(segment) && segment.contains(PATH_MAP_START);
+    }
+
+    public static String findNextSegment(String segment, String path) {
+        AtlasPath pathUtil = new AtlasPath(path);
+        String parentSegment = null;
+        for (String currentSegment : pathUtil.segments) {
+            if (parentSegment == null || !segment.equals(parentSegment)) {
```
igarashitm: Yeah I think Codacy is right here, this loop looks super weird. @sunilmeddr I know it was just copied, but can you turn over the condition and just return in 'if'?
https://www.codacy.com/app/atlasmapio/atlasmap/pullRequest?prid=1131462

igarashitm:
Ah you put the commit log message onto follow-up commit - you may want to squash into single commit then? GitHub Blame is a powerful tool to trace the code later and it relies on appropriate commit message.

sunilmeddr:
Method findNextsegment has no references,So I deleted the method.

igarashitm:
I meant the commit `Updated for issues 11.135,136,141` contains major part of your work for this, which should have an appropriate commit log. I would rather squash into a single commit though.

igarashitm:
Wait, please don't include other commit which has no relation with the topic, path handling. `StringConverter should be more forgiving for conversion to Boolean` is totally different topic, so it must be a separate pull request.

PR #150, opened 2017-12-06
byte support in reference mappings

### LINKED ISSUE #7: Update BYTE conversion behavior
@igarashitm commented on [Wed Nov 08 2017](https://github.com/atlasmap/atlasmap-runtime/issues/279)

It looks Byte related converters are all marked as UNSUPPORTED but shouldn't be, easy to implement. Also NUMBER type conversion must be provided OOTB as it often appears in the JSON schema.

BYTE -> Short 
BYTE -> String (string representation of the number)
...

-127 to 127
Int, Short, Long -> BYTE (range concern)
String -> BYTE (range concern and type concern)
Decimal -> BYTE (precision concern.. if 1.2 comes in as a source value throw an exception)

rohanmars:
fixes #7 

references mapping now make use of byte field conversion
updated reference mapping to include char and boolean
they also include "happy path" testing, i.e. testing form each primitive type to each other.

includes fixes for fieldType support on XML and Json readers

***** Boolean support has been changed for numerical values, such that 0 is false and all other numbers true.


igarashitm:
Oops, I thought you are now working on NUMBER converters... This will hit a big confliction with #145.

I think TypeConverter tests shouldn't be added to the itests. We already have 
https://github.com/atlasmap/atlasmap/tree/master/runtime/core/src/test/java/io/atlasmap/converters

And there, you can even add unit tests using DefaultAtlasConversionService like this
https://github.com/atlasmap/atlasmap/blob/master/runtime/core/src/test/java/io/atlasmap/core/DefaultAtlasConversionServiceTest.java

rohanmars:
@igarashitm why would reference mapping need to change for your fix? in very basic primitive integration tests? I haven't fully analyzed you changes yet but I think you have the order of the mappings now out of sync. I spent 3 days adding in proper byte,char,boolean support. meh. this sets me back to the beginning.

@mattrpav. thoughts.

igarashitm:
Because so many validation errors were buried and it popped up once I enabled runtime validation. That's why I raise #154

rohanmars:
What type of errors? I spent a lot of time making sure the basic primitive inter-operable tests where now working. Arguably they were not covering all corners. I still don't get why basic itest primitive mapping files would need to change. Also, I saw you have added byte in some places, or mapped boolean to boolean in others. I'm confused.

mattrpav:
@igarashitm the tests Rohan is referencing look and feel like integration tests to me. The original intent was to capture all combinations of various features across and _**between**_ the formats. The tests build on each other as they start with basic functionality and then move to cross-format testing. I agree, straight conversion test should be unit test-- and they are.. and increasing test coverage on those is handled by Joey's unit test coverage tasks.

Rohan's BYTE and char work complete the primitive handling and these integration tests provide clean examples of many combinations of scenarios demonstrating a cross-cutting of features.

Same formats:
Java: auto-type detection (mapping definitions do not contain FieldType entries), auto-conversion, etc
Json: Rooted vs Unrooted, auto-detection, auto-conversion, etc
Xml: Namespaced, non-namespaced, auto-detection, auto-conversion, etc

Across formats:
Java to/from Json, Xml
Json to/from Java, xml
Xml to/from Java, Json

igarashitm:
Do you think it's impossible to split them into
* type converter in field reader
* type converter in mapping
* type converter in field writer

These exhaustive field type tests in the integration tests are making it much harder to maintain. It would always conflict with other like in this case.

igarashitm:
btw @rohanmars can you file an issue with details and let us know before you start working? instead of opening closed issue.

rohanmars:
yes, I did just that on the call last week when I asked if I should open a new issue, but you mentioned that I should just open this one back up? was that not correct?

igarashitm:
Sorry, it didn't sound like this big. Then better reopening it before start working and put a comment about what is left.

igarashitm:
OK then let's think how we can merge this... basically we've done
* same bug fixes around the field reader and field writer, but can't say if all same or not
* I removed *->BOOLEAN conversions in the reference-mappings since it threw an Exception as expected after bug fixes

It looks the former is more tough to solve. There's no easy way. @rohanmars if you can't handle it then let me know, I'll take over.

rohanmars:
I've had to make some changes to how char and string converters to work to boolean for example. For this I created a new conversion concern "CONVENTION"

I've added back the reference mappings so that they convert between all the base primitive types.

Issue #151, opened 2017-12-06
Implement Number-related p0 field actions


jpav:
Sub feature of #55

jpav:
Fixed by PR #152

PR #152, opened 2017-12-06
Issue #151: Implement Number-related p0 field actions

jpav:


jpav:
@igarashitm, Can we change the checkstyle rules so that all instance members exist after static members (including constructors)?

igarashitm:
Sounds good to me, you can add that rule

jpav:
Well, looks like checkstyle only allows me to ignore where constructors show up, which doesn't seem good, so I just moved it to where it wanted it.

runtime/core/src/main/java/io/atlasmap/actions/NumberFieldActions.java
```diff
@@ -238,207 +293,206 @@ public static Number convertVolumeUnit(Action action, Number input) {
...
-            inputs = dinList;
+            return dinList;
         } else if (input instanceof float[]) {
             float[] fin = (float[]) input;
             List<Float> finList = new ArrayList<>(fin.length);
             for (float e : fin) {
                 finList.add(e);
             }
-            if (numberType == null) {
-                numberType = NumberType.FLOAT;
```
igarashitm: Is this field type filling causing any problem?

igarashitm: ah. just noticed it's now in the another method...

Issue #176, opened 2017-12-09
ClassInspectionService is returning NOT_FOUND status for the COMPLEX fields of twitter4j.Status


igarashitm:
This happens when running via atlasmap-maven-plugin. There's a `ComplexClassInspectTest` inspects `SourceOrder`, which has `COMPLEX` field successfully inspected though.
```json
{
    "JavaClass": {
        "jsonType": "io.atlasmap.java.v2.JavaClass",
        "modifiers": {
            "modifier": ["ABSTRACT", "INTERFACE", "PUBLIC"]
        },
        "className": "twitter4j.Status",
        "primitive": false,
        "synthetic": false,
        "javaEnumFields": {
            "javaEnumField": []
        },
        "javaFields": {
            "javaField": [{
                "jsonType": "io.atlasmap.java.v2.JavaField",
                "arrayDimensions": 1,
                "collectionType": "ARRAY",
                "path": "withheldInCountries",
                "status": "SUPPORTED",
                "fieldType": "STRING",
                "modifiers": {
                    "modifier": []
                },
                "name": "withheldInCountries",
                "className": "java.lang.String",
                "getMethod": "getWithheldInCountries",
                "primitive": true,
                "synthetic": false
            }, {
                "jsonType": "io.atlasmap.java.v2.JavaField",
                "path": "quotedStatusId",
                "status": "SUPPORTED",
                "fieldType": "LONG",
                "modifiers": {
                    "modifier": []
                },
                "name": "quotedStatusId",
                "className": "long",
                "getMethod": "getQuotedStatusId",
                "primitive": true,
                "synthetic": false
            }, {
                "jsonType": "io.atlasmap.java.v2.JavaField",
                "path": "text",
                "status": "SUPPORTED",
                "fieldType": "STRING",
                "modifiers": {
                    "modifier": []
                },
                "name": "text",
                "className": "java.lang.String",
                "getMethod": "getText",
                "primitive": true,
                "synthetic": false
            }, {
                "jsonType": "io.atlasmap.java.v2.JavaField",
                "path": "source",
                "status": "SUPPORTED",
                "fieldType": "STRING",
                "modifiers": {
                    "modifier": []
                },
                "name": "source",
                "className": "java.lang.String",
                "getMethod": "getSource",
                "primitive": true,
                "synthetic": false
            }, {
                "jsonType": "io.atlasmap.java.v2.JavaClass",
                "path": "scopes",
                "status": "NOT_FOUND",
[...]

PR #178, opened 2017-12-09
fix: ClassInspectionService is returning NOT_FOUND status for the COM…

igarashitm:
…PLEX fields of twitter4j.Status

Fixes: #176

### DIFF
--- a/runtime/atlasmap-maven-plugin/src/main/java/io/atlasmap/maven/GenerateInspectionsMojo.java
+++ b/runtime/atlasmap-maven-plugin/src/main/java/io/atlasmap/maven/GenerateInspectionsMojo.java
@@ -182,15 +182,12 @@ public class GenerateInspectionsMojo extends AbstractMojo {
             JavaClass c = null;
             // Not even this plugin will be available on this new URLClassLoader
             try (URLClassLoader loader = new URLClassLoader(urls.toArray(new URL[urls.size()]), origTccl)) {
-                Thread.currentThread().setContextClassLoader(loader);
                 clazz = loader.loadClass(className);
                 ClassInspectionService classInspectionService = new ClassInspectionService();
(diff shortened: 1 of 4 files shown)


igarashitm:
@rhuss , @lburgazzoli , FYI, we need this one to be merged prior to release to get twitter4j.Status inspection work via maven plugin.

igarashitm:
@atlasmap/runtime Could anybody make a review on this? It's just allowing to pass ClassLoader into ClassInspectionService. This is required for maven plugin.

runtime/atlasmap-maven-plugin/src/main/java/io/atlasmap/maven/GenerateInspectionsMojo.java
```diff
@@ -182,15 +182,12 @@ private void generateInspection(List<String> artifacts, Collection<String> class
...
-                Thread.currentThread().setContextClassLoader(loader);
                 clazz = loader.loadClass(className);
                 ClassInspectionService classInspectionService = new ClassInspectionService();
                 classInspectionService.setConversionService(DefaultAtlasConversionService.getInstance());
-                c = classInspectionService.inspectClass(clazz);
+                c = classInspectionService.inspectClass(loader, clazz);
             } catch (ClassNotFoundException | IOException e) {
                 throw new MojoExecutionException(e.getMessage(), e);
-            } finally {
-                Thread.currentThread().setContextClassLoader(origTccl);
```
jpav: Doesn't seem like you need this finally block since you never set the TCCL to anything before this anymore (i.e., since you deleted line 185)

igarashitm: Yep, so I'm removing it. So I'm seeing L191-L192 is also red. Do you see something different?

Issue #180, opened 2017-12-10
Add twitter4j.Status as a mock Source Document


igarashitm:
ATM twitter4j.Status is a most important Document. Also add twitter4j deps into itests profile of runtime.

PR #181, opened 2017-12-10
chore: Add twitter4j.Status as a mock Source Document

igarashitm:
Fixes: #180

chore: Reduce surefire fork count, expecting to avoid recent VM termination in CircleCI


igarashitm:
@atlasmap/runtime Could anybody review on this?

Issue #13, opened 2017-11-15
Java field inspection doesn't support Enum fields


igarashitm:
I suspect Enum field without getter/setter is not handled here. Add a unit test to inspect a class which has a Enum field without getter/setter, and see if it's in the inspection result. If not, implement it.

https://github.com/atlasmap/atlasmap/blob/943cab3fe8e28250a27ed1e302f3dae8378fecee/runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java#L597-L599

igarashitm:
w/r/t not handling `JavaEnumField` in `inspectGetMethod()`, Let's talk here as it seems to be related - so as far as I can see from debugger, Enum inspection doesn't hit this method. I checked with `TargetTestClass` which has `StateEnumClassLong` Enum, and it looks successfully inspected and handled through UI. Enum might be handled somewhere else.
If so, `ClassInspectionService` code might be fishy, like entering method anyway but return without doing anything if it's Enum. It might need to be refactored.

sunilmeddr:
Yes. I will fix the method inspection for Enum Fields along with the field
inspection.



On Thu, Dec 7, 2017 at 10:08 AM, Tomohisa Igarashi <notifications@github.com

igarashitm:
The first step would be to add an unit test verifying field Enum inspection (i.e. inspection succeed without getter method for the Enum field), if it works, maybe you can just go over to next task... I don't yet have concrete idea how it should be refactored.

sunilmeddr:
ok

On Thu, Dec 7, 2017 at 10:17 AM, Tomohisa Igarashi <notifications@github.com

igarashitm:
@sunilmeddr of course if you have a good idea how to refactor `ClassInspectionService`, that would be awesome to do it!

sunilmeddr:
I am planning to add code for the inspection of the Enum fields and in the
get and set method inspection,I will check the method parameter and return
type,if they are of the type Enum,I will update the JavaEnumField,otherwise
I will update the JavaField .

On Thu, Dec 7, 2017 at 10:21 AM, Tomohisa Igarashi <notifications@github.com

mattrpav:
Let's clarify communication and hop on a call.. will send out some times.

sunilmeddr:
I tested the Class Inspection by hitting the java service
http://localhost:8585/v2/atlas/java/class?className=io.atlasmap.java.test.
[...]

PR #182, opened 2017-12-10
fix: Java field inspection doesn't support Enum fields

sunilmeddr:
Fixes: #13


runtime/modules/java/core/src/test/java/io/atlasmap/java/inspect/ClassInspectionServiceTest.java
```diff
@@ -110,9 +110,10 @@ public void testClasspathToList() {
     @Test
     public void testDateTimeViaField() {
         JavaClass javaClass = classInspectionService.inspectClass(DateTimeField.class);
-        // FIXME java.time.Month is Enum - https://github.com/atlasmap/atlasmap-runtime/issues/251
+        // FIXME java.time.Month is Enum -
```
igarashitm: So you can just remove 113-115 right? :-)

sunilmeddr:
Removed the unnecessary comments.

On Sun, Dec 10, 2017 at 5:13 PM, Tomohisa Igarashi <notifications@github.com

Issue #162, opened 2017-12-06
Java inspection to support Map as a root-level object


mattrpav:
1. Investigate how long it would take to support
2. If level of effort < 1 day (a few hours), go ahead and implement

Initial investigation revealed that the current model doesn't work once we add Collection.MAP. We don't have appropriate properties in the Java inspection result model to carry parameterized types for the Map. For example, with `Map<String, Object>`, we need to store field type and field class for both of `String` and `Object`. Note that for other Collection types, we are populating field type and field class with a parameterized type read from Collection class, as it only has one for values. That doesn't work for Map without model change as it has two.

sunilmeddr:
I captured the Map key and values in JavaClass as JavaFields.JavaField[n],JavaFields.JavaField[n+1]. Is there any better way,we can capture the Map key and values.

igarashitm:
@sunilmeddr ah it doesn't need to inspect into Map, at design time we have no guarantee to have a complete set of Map keys/values, UI would just prefill with the keys provided from outside. Java inspection would just return it with saying it's a Map, and that's it.

PR #188, opened 2017-12-11
fix: Java inspection to support Map as a root-level object

sunilmeddr:
Fixes: #162

### DIFF
--- a/runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
+++ b/runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
@@ -55,12 +55,15 @@ public class ClassInspectionService {
     public static final int MAX_ARRAY_DIM_LIMIT = 256; // JVM specification
 
     private static final Logger LOG = LoggerFactory.getLogger(ClassInspectionService.class);
-                                                       // limit
+    // limit
 
     private List<String> listClasses = new ArrayList<String>(
             Arrays.asList("java.util.List", "java.util.ArrayList", "java.util.LinkedList", "java.util.Vector",
                     "java.util.Stack", "java.util.AbstractList", "java.util.AbstractSequentialList"));
-
+    private List<String> mapClasses = new ArrayList<String>(Arrays.asList("java.util.Map", "java.util.HashMap",
+            "java.util.TreeMap", "java.util.Hashtable", "java.util.IdentityHashMap", "java.util.LinkedHashMap",
+            "java.util.LinkedHashMap", "java.util.SortedMap", "java.util.WeakHashMap", "java.util.Properties",
+            "java.util.concurrent.ConcurrentHashMap", "java.util.concurrent.ConcurrentMap"));
     private AtlasConversionService atlasConversionService = null;
     private List<String> fieldBlacklist = new ArrayList<String>(Arrays.asList("serialVersionUID"));
     private List<String> classNameBlacklist = new ArrayList<String>();
@@ -69,6 +72,10 @@ public class ClassInspectionService {
     private Boolean disablePrivateOnlyFields = false;
     private Boolean disablePublicOnlyFields = false;
     private Boolean disablePublicGetterSetterFields = false;
+
+    public List<String> getMapClasses() {
+        return this.mapClasses;
+    }
 
     public List<String> getListClasses() {
         return this.listClasses;
@@ -244,6 +251,10 @@ public class ClassInspectionService {
             clz = clazz;
         }
 
+        if (isMapList(clz.getCanonicalName())) {
+            javaClass.setCollectionType(CollectionType.MAP);
+        }
+
         javaClass.setClassName(clz.getCanonicalName());
         javaClass.setPackageName((clz.getPackage() != null ? clz.getPackage().getName() : null));
         javaClass.setAnnotation(clz.isAnnotation());
@@ -330,8 +341,8 @@ public class ClassInspectionService {
         field.setName(StringUtil.removeGetterAndLowercaseFirstLetter(m.getName()));
 
         if (pathPrefix != null && pathPrefix.length() > 0) {
-            field.setPath(
-                    pathPrefix + AtlasPath.PATH_SEPARATOR + StringUtil.removeGetterAndLowercaseFirstLetter(m.getName()));
+            field.setPath(pathPrefix + AtlasPath.PATH_SEPARATOR
+                    + StringUtil.removeGetterAndLowercaseFirstLetter(m.getName()));
         } else {
             field.setPath(StringUtil.removeGetterAndLowercaseFirstLetter(m.getName()));
         }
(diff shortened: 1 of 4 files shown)


mattrpav:
@igarashitm I was thinking we could add the keys as fields. Using AtlasPath for coordinates. It is similar to using an index in arrays and lists. 

Field1: atlasPath: /{key1}
Field2: atlasPath: /{key2}

Thoughts?

igarashitm:
Hmm I understand that would work for mapping definition, but does it make sense on inspection? UI can only provides prefilled keys, but user would add more, and there're even more at runtime. I don't see any benefit to let inspection handle only that prefilled keys.

igarashitm:
And I thought inspection doesn't handle any index. does it? At the time of performing inspection, we don't have any info about the index right?

mattrpav:
Right, inspection doesn’t know about indexes and populated values. I was speaking in terms of how the UI (syndesis) might inject the data for mapping purposes. 

Conceptually, I think Maps could work like user defined fields in XML and JSON. Syndesis may provide some keys, and the user may provide some keys. 

Data model wise, I think we have a couple approaches:

Option 1:
1. defining the Map as a collection of type Map

2. ‘Keys’ are child JavaFields of the Map JavaField and use the AtlasPath expression describe above. (Treating keys as JavaField has the bonus that all the type info is available from and including primitive to complex)

3. Values could be set in the ‘value’ field of the JavaField definition for each ‘key’ 

In the UI, you support adding a child field to the Map, just like in XML. 

Option 2:
The other approach might be to link the Map JavaField collection field to a LookupTable. The downside there is that LookupTable is modeled Essentially like Entry<Primitive, Primitive> so it’d be tough to add <String, Object> and other complex map support.

igarashitm:
Yep, and we decided to go with option1 IINM. But I don't understand how that topic fall into the scope of this pull request and #162. It's all about Java inspection.

mattrpav:
Apologies, looks like inside tracked the convo. 

Is this the current question at hand?

Should ClassInspectionService support inspecting a Map Object or Map Class?

igarashitm:
@mattrpav whoa, you filed that issue (#162) ;-)
so the intention of the issue was to take a look at the Java class inspection and see if we want to do something for java.util.Map class inspection as a top level Document type. Actual entries are out of scope.

sunilmeddr:
The attachment contains the inspection results when the Map is a root and field
[Inspection results.zip](https://github.com/atlasmap/atlasmap/files/1556188/Inspection.results.zip)

igarashitm:
Just curious, for the Map as a field, how does it determine `fieldType` as `STRING`?

So still my question is open, how can we deliver the type parameters for Map (`String` and `Object` for Map<String,Object>)? 
Other collections only have one type parameter, which is a type of the entry. How do we carry 2 parameters for Map?

igarashitm:
As first target is `Map<String, String>`, it would be always `String` and `String`. But still worth to address it now so we don't have to rewrite everything later.

sunilmeddr:
Map Key information(fieldType and className) will be captured in a  java
field of type Map and the the Map's value type will be captured in the Java
field value. Please check the attachments for the inspection results.
Please refer below for the Javafield section of a class inspection ,where
class contain a field of type Map<String,BaseOrder>

 "javaFields": {
            "javaField": [
                {
                    "jsonType": "io.atlasmap.java.v2.JavaField",
                    "value": "io.atlasmap.java.test.BaseOrder",
                    "collectionType": "MAP",
                    "path": "orders",
                    "fieldType": "STRING",
                    "modifiers": {
                        "modifier": [
                            "PRIVATE"
                        ]
                    },
                    "name": "orders",
                    "className": "java.lang.String",
                    "getMethod": "getOrders",
                    "setMethod": "setOrders",
                    "primitive": true
                }
            ]



On Wed, Dec 13, 2017 at 2:04 PM, Tomohisa Igarashi <notifications@github.com

mattrpav:
I don’t think we capture parameterized types yet during inspection. We’d need to extend the JavaClass and JavaField model types to support a place to put it.

sunilmeddr:
With the Data Model we have , we can't capture a  parameterized types in a
JavaField. We have to update the data model by creating a new class by
extending the JavaField. The proposed solution is the alternative way to
capture the parameterized types with the available data model.

sunilmeddr:
Please let me know, if we are not required to capture the parameterized
types until we create a new data model class

igarashitm:
Existing model properties should not be used for different purpose.

This revealed that the current model doesn't work once we add Collection.MAP. Note that for other Collection types, we are populating field type and field class with a parameterized type read from Collection class. That doesn't work for Map without model change as it has two.

sunilmeddr:
I will revert the changes to capture the Parameterized types, when the Map
is a field.

On Thu, Dec 14, 2017 at 9:34 AM, Tomohisa Igarashi <notifications@github.com

sunilmeddr:
I have reverted the changes to capture the parameterized type in a field. The only changes made are adding the collectiontype when the Java Field or Class is of type Map.

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -51,16 +51,21 @@
 
 public class ClassInspectionService {
 
+    public static final String MAP_KEY = "key";
+    public static final String MAP_VALUE = "value";
```
igarashitm: So then these 2 constants are no longer required?

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -240,10 +249,16 @@ protected void inspectClass(Class<?> clazz, JavaClass javaClass, Set<String> cac
             javaClass.setArrayDimensions(detectArrayDimensions(clazz));
             javaClass.setCollectionType(CollectionType.ARRAY);
             clz = detectArrayClass(clazz);
-        } else {
+        }
+
+        else {
```
igarashitm: weird line feed?

igarashitm:
Also can you change `Fixes:` header to `Relates to:` or something? I don't want to close #162 until we support parameterized types, including model changes.

sunilmeddr:
Do we have an issue number to add to Related to:

igarashitm:
It's still #162 , but `Relates to:` won't close the issue, which I want

sunilmeddr:
Please share your thoughts on how to capture the parametrized types. We
have the following 2 use cases.

1) Map as root :  We can't capture the Parameterized types. We can capture
the Parmeterized types only for the super classes or for the fields
2)  Map as a field:  Parameterized types are captured in the following
property in the Java Field. If we want to capture more details in a
separate Data model(like JavaMapField that extends JavaClass), please share
your thoughts on the fields in the Java class

parameterizedTypes": {
                        "string": [
                            "java.lang.String",
                            "io.atlasmap.java.test.BaseOrder"
                        ]
                    }

Please note we already updated our code to identify a field as Map and the
changes not merged to our code.

Issue #126, opened 2017-11-22
Create community release job for jenkins


igarashitm:
Here is the build script for monorepo, it builds everything including atlasmap/ui
https://github.com/atlasmap/atlasmap/blob/master/build.sh

igarashitm:
@iocanel if you could help for this one that would be awesome!

igarashitm:
maven profiles from atlasmap-runtime old job - https://jenkins-syndesis-ci.b6ff.rh-idev.openshiftapps.com/job/atlasmap-release/configure
```
fabric8,ci,release,community-release
```

same from camel-atlasmap old job - https://jenkins-syndesis-ci.b6ff.rh-idev.openshiftapps.com/job/camel-atlasmap-release/configure
```
fabric8,ci,release
```

igarashitm:
Hi @iocanel , can you look into this? it seems we need a release soon, to have #176 for the atlasmap-maven-plugin used in syndesis pre-computed inspection.

igarashitm:
I just noticed we need to move those release related profiles to top level pom, as we'll need to release atlasmap/parent as well. I'll try adjusting POMs and add `--release` option into build.sh which adds release specific profiles.

rhuss:
@igarashitm Is this still a blocker for TP3 ?

igarashitm:
@rhuss just removed the blocker tag, as I succeeded to make a 1.32.2 release manually.

PR #190, opened 2017-12-11
fix: Moved release profiles into root pom and added release targets to build.sh

igarashitm:
Relates to: #126

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,7 +10,7 @@ ARGS="$@"
 
 # Display a help message.
 function displayHelp() {
-    echo "This script helps you build the syndesis monorepo."
+    echo "This script helps you build the AtlasMap monorepo."
     echo "The available options are:"
     echo " --skip-tests            Skips the test execution."
     echo " --skip-image-builds     Skips image builds."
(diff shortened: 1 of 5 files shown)


igarashitm:
Hi @rhuss , @iocanel ,
I added `release.sh` to perform `mvn release:prepare release:perform`, and moved release related profiles into root POM. Mainly because I thought release should run only if `build.sh` successfully complete. So we can run `build.sh` to build and then `release.sh` for release from Jenkins. WDYT?

@cunningt ,
I'm also wondering if this change affects product build or not (GitHub wisely suggested you as a reviewer :-). Can you take a look?

rhuss:
> I added release.sh to perform mvn release:prepare release:perform, and moved release related profiles into root POM. Mainly because I thought release should run only if build.sh successfully complete. So we can run build.sh to build and then release.sh for release from Jenkins. WDYT?

Normally a release _includes_ a full build, so I actually would not separate them (as it also probably shares common functionality). Maybe adding an option to build.sh to enable the profile and then call the proper mvn targets ?

igarashitm:
@rhuss thanks for the comment, OK that works for me.
I added `--release-version` and `--development-version` into build.sh, and if `--release-version` is specified, it invokes `release:prepare release:perform release:clean` targets at last.

igarashitm:
Alright, @iocanel is this enough to get you work on adding Jenkins job for AtlasMap release?

PR #193, opened 2017-12-13
fix: Make [None] disappear on focus

### LINKED ISSUE #94: Make [None] disappear on focus
|</img> @rhuss | [2017-10-06](https://github.com/atlasmap/atlasmap-ui/issues/98) | bug  |
|-|-|-|

When creating a new mapping via typing into the right hand colum fields "[None]" should disappear on focus. Now I have to manually delete it, char-by-char to when creating a new entry

image

akieling:
Fixes #94

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.detail.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.detail.component.ts
@@ -14,7 +14,7 @@
     limitations under the License.
 */
 
-import { Component, Input } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
 
 import { ConfigModel } from '../../models/config.model';
@@ -35,8 +35,8 @@ import { MappingModel, FieldMappingPair, MappedField } from '../../models/mappin
         <ng-template #tolTemplate>
             <div class="fieldDetailTooltip">
                 <label class="parentObjectName" *ngIf="displayParentObject()">
-                    <i [attr.class]="getSourceIconCSSClass()"></i>
-                    {{ getParentObjectName() }}
+                    <i [ngClass]="sourceIconCSSClass"></i>
+                    {{ parentObjectName }}
                 </label>
                 <label>{{ getFieldPath() }}</label>
                 <label *ngIf="displayParentObject() && mappedField.field.type">({{ mappedField.field.type }})</label>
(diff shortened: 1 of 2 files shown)


akieling:
Do you have a better idea?

jpav:
Or maybe just "Search", grayed out.  That seems to be what you'd most likely see in search fields on other websites.

igarashitm:
Yep, sounds fine, my text was too long for that short input field ;-)

akieling:
The problem is that the tooltip always show up on hover and it looks broken when the input field is empty. I tried to disable the tooltip when the input field is empty but didn't succeed on that and decided to avoid spending too much time on that. It's a kind of rabbit hole. I personally think the tooltip should be removed but that's something the UI team needs to review. A quick fix would be removing or selecting (what I did) the input field text when it's showing the [None] text and the user clicks on it.

igarashitm:
Again, can you file an another issue for that? I can't accept burying bugs for any purpose.

akieling:
Created #197

jpav:
Also, to again imitate what I see on other websites, assuming we put "Search" in the field, I think the expectation would be that on focus, either a) the "Search" text automatically disappears and reappears on focus loss if the field remains empty, or b) the "Search" text toggles depending on when text is typed or the field is empty (just like the search field works at the top of this GitHub page)

akieling:
I agree but doing that without removing the tooltip creates another bug: you see a tooltip with no content. That's why I think it should be done as part of #197.

akieling:
Pushed some changes more in line with what you expect. When adding a new mapping, the fields are empty and the tooltip doesn't show up. When there is a mapping, an information icon is placed after the label name and the tooltip shows up when hovering this icon.

ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.detail.component.ts
```diff
@@ -44,17 +44,18 @@ import { MappingModel, FieldMappingPair, MappedField } from '../../models/mappin
...
-        <div class='fieldDetail' style="margin-bottom:5px;" *ngIf="mappedField"
-            [tooltip]="tolTemplate" placement="left">
+        <div class='fieldDetail' style="margin-bottom:5px;" *ngIf="mappedField">
             <label class="parentObjectName" *ngIf="displayParentObject()">
                 <i [attr.class]="getSourceIconCSSClass()"></i>
                 {{ getParentObjectName() }}
+                <span class="pficon pficon-info" [tooltip]="tolTemplate" placement="left"></span>
             </label>
             <div style="width:100%;">
                 <input type="text" id="input-{{isSource?'source':'target'}}-{{mappedField.field.getFieldLabel(false)}}"
```
deeleman: Regarding that `{{mappedField.field.getFieldLabel(false)}}` binding, function calls inside templates have a negative impact on performance down the line, since the Angular template engine will re-execute that function on each pass of the `ChangeDetection` cycle.

deeleman: The same applies to the entire block actually. `getSourceIconCSSClass()` and `getParentObjectName()` should be moved to the component controller and exposed by means of component fields, or even class getters.

akieling:
Finished changes requested by @deeleman

ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.detail.component.ts
```diff
@@ -35,45 +35,54 @@ import { MappingModel, FieldMappingPair, MappedField } from '../../models/mappin
...
                 </label>
                 <label>{{ getFieldPath() }}</label>
                 <label *ngIf="displayParentObject() && mappedField.field.type">({{ mappedField.field.type }})</label>
                 <div class="clear"></div>
             </div>
         </ng-template>
 
-        <div class='fieldDetail' style="margin-bottom:5px;" *ngIf="mappedField"
-            [tooltip]="tolTemplate" placement="left">
+        <div class='fieldDetail' style="margin-bottom:5px;" *ngIf="mappedField">
```
deeleman: Quotes are not consistent here. The convention for element attributes will be double quotes.

ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.detail.component.ts
```diff
@@ -35,45 +35,54 @@ import { MappingModel, FieldMappingPair, MappedField } from '../../models/mappin
...
             <label class="parentObjectName" *ngIf="displayParentObject()">
-                <i [attr.class]="getSourceIconCSSClass()"></i>
-                {{ getParentObjectName() }}
+                <i [attr.class]="sourceIconCSSClass"></i>
```
deeleman: `<i [ngClass]="sourceIconCSSClass"></i>`would be more appropriate here

ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.detail.component.ts
```diff
@@ -35,45 +35,54 @@ import { MappingModel, FieldMappingPair, MappedField } from '../../models/mappin
...
+                {{ parentObjectName }}
+                <span class="pficon pficon-info" [tooltip]="tolTemplate" placement="left"></span>
             </label>
             <div style="width:100%;">
-                <input type="text" id="input-{{isSource?'source':'target'}}-{{mappedField.field.getFieldLabel(false)}}"
+                <input type="text" id="{{inputId}}"
```
deeleman: just out of curiosity, we are we binding interpolated values to an `id` attribute? I'm just curious...

igarashitm: @deeleman , I almost overlooked before your comment - so the id is required to capture input fields in the HTML for the e2e test framework. @mmelko , isn't this change breaking what we want for e2e framework?

akieling: @deeleman I don't know why we do this. It's existing code. It looks odd to me. @igarashitm I didn't change any logic related to how we build the id value with the goal of not breaking any e2e tests, in case we depend on it.

deeleman: Now it makes sense. Thanks for the explanation guys! ;)

mmelko: @deeleman as it was mentioned before, It's used in syndesis e2e tests. @akieling I agree not the nicest but it works. Thanks for keeping it ;) We might want to change logic of building id in the future though.

PR #194, opened 2017-12-13
fix: datamapper buttons are links

akieling:
Fixes: #97


igarashitm:

PR #195, opened 2017-12-13
fix: No focus on text field when search is selected

akieling:
Fixes: #99

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/common/focus.directive.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/common/focus.directive.ts
@@ -1,0 +1,20 @@
+import { Directive, ElementRef, Input, AfterContentChecked } from '@angular/core';
+
+@Directive({
+  selector: '[focus]'
+})
+export class FocusDirective implements AfterContentChecked {
+
+  @Input() focus: boolean;
+  private hasFocus = false;
+
+  constructor(private elementRef: ElementRef) {
+  }
+
+  ngAfterContentChecked() {
+    if (this.focus && !this.hasFocus) {
+      this.elementRef.nativeElement.focus();
+      this.hasFocus = true;
+    }
+  }
+}
--- a/ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
@@ -47,7 +47,7 @@ import { ModalWindowComponent } from './modal.window.component';
                 <div *ngIf="searchMode" class="searchBox">
                     <input type="text" #searchFilterBox
                         id="search-filter-box" (keyup)="search(searchFilterBox.value)" placeholder="Search"
-                        [(ngModel)]="searchFilter" />
+                        [(ngModel)]="searchFilter" [focus]="true" />
                     <i class="fa fa-close searchBoxCloseIcon link" (click)="toggleSearch()"></i>
                     <div class="clear"></div>
                 </div>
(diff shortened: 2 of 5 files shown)


igarashitm:

ui/src/app/lib/atlasmap-data-mapper/common/focus.directive.ts
```diff
@@ -0,0 +1,20 @@
...
+  constructor(private elementRef: ElementRef) {
+  }
+
+  ngAfterContentChecked() {
+    if (this.focus && !this.hasFocus) {
+      this.elementRef.nativeElement.focus();
+      this.hasFocus = true;
+    }
+  }
+}
```
deeleman: What happens when the user clicks again the search toggle after the first focus event? I might be wrong but I believe this implementation will work just once, since `hasFocus` is never reset to `false` whenever we toggle the search again.

akieling: A new instance of FocusDirective is created every time the input field is shown. It's probably due to the fact that the input field is added/removed from the DOM, right?

deeleman: If we're toggling that input with a structural directive such as `*ngIf`, then yes. It is usually advised to avoid this approach since it entails a performance cost for repainting that piece of DOM and mostly because it breaks the reusability of that directive, forcing its host element to be removed off the DOM in order to trigger the focus behavior. 

Might I suggest this other approach? :)

```typescript 
import { Directive, ElementRef, Input, AfterContentChecked, HostListener } from '@angular/core';

@Directive({
  selector: '[focus]'
})
export class FocusDirective implements AfterContentChecked {
  @Input() focus: boolean;
  private hasFocus = false;

  constructor(private elementRef: ElementRef) { }
  
  @HostListener('blur')
  resetFocus() {
    this.hasFocus = false;
  }
  
  @HostListener('focus')
  enableFocus() {
    this.hasFocus = true;
  }

  ngAfterContentChecked() {
    if (this.focus && !this.hasFocus) {
      this.elementRef.nativeElement.focus();
    }
  }
}

```

akieling: Your implementation looks great but I got the following error after clicking on the search icons on both panels:
```
DocumentDefinitionComponent.html:16 ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'true'. Current value: 'false'.
    at viewDebugError (core.js:9515)
    at expressionChangedAfterItHasBeenCheckedError (core.js:9493)
    at checkBindingNoChanges (core.js:9662)
    at checkNoChangesNodeInline (core.js:13673)
    at checkNoChangesNode (core.js:13647)
    at debugCheckNoChangesNode (core.js:14455)
    at debugCheckRenderNodeFn (core.js:14395)
    at Object.eval [as updateRenderer] (DocumentDefinitionComponent.html:16)
    at Object.debugUpdateRenderer [as updateRenderer] (core.js:14377)
    at checkNoChangesView (core.js:13489)
```

deeleman: Darn! That pesky error has chased me forever. We're updating the host element right after it has been checked already by the changeDetection machinery, so the component does not feature the same state in the check step of the change detection cycle that it had in the `DoCheck` step. Good news is that you'll only see this annoyance in dev mode tho. Nevermind, although a bit hacky, we can either move the logic to the `OnInit` step of the lifecycle process or just defer the execution of the logic inside `ngAfterContentChecked ` to the next tick by means of a `setTimeout(() => { ... }, 0)` expression.

Otherwise let's get away with the directive as is, but bear in mind that we'll only be able to use it when the host element is meant to be repainted on blur, and this constrains its purpose quite a bit.

akieling: Isn't the current behavior in line with the `autofocus` attribute of the HTML `input` element? It only triggers once when the view is rendered.

akieling: I tried using `setTimeout` inside `ngAfterContentChecked` and it didn't work well.

igarashitm:
@akieling is this one ready to ask @deeleman 's review again?

akieling:
@igarashitm yes

igarashitm:
Hi @deeleman , would you have a look and approve if it looks OK? thanks!

PR #196, opened 2017-12-13
fix: Table view shows wrong icon for the validation error

akieling:
Fixes: #103

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.list.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.list.component.ts
@@ -127,8 +127,9 @@ export class MappingListFieldComponent {
                                     <div class="clear"></div>
                                 </div>
                                 <div class="error">
-                                    <span class="pficon pficon-error-circle-o"
-                                    *ngIf="mapping.validationErrors.length"></span>
+                                    <span class="pficon"
+                                          [ngClass]="mapping.validationErrors | toErrorIconClass"
+                                          *ngIf="mapping.validationErrors.length"></span>
                                 </div>
                                 <div class="clear"></div>
                             </div>
--- a/ui/src/app/lib/atlasmap-data-mapper/common/to-error-icon-class.pipe.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/common/to-error-icon-class.pipe.ts
@@ -1,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { ErrorInfo, ErrorLevel } from '../models/error.model';
+
+@Pipe({ name: 'toErrorIconClass' })
+export class ToErrorIconClassPipe implements PipeTransform {
+
+  transform(value: ErrorInfo[]): string {
+    if (value.some(e => e.level >= ErrorLevel.ERROR)) {
+      return 'pficon pficon-error-circle-o';
+    } else if (value.some(e => e.level === ErrorLevel.WARN)) {
+      return 'pficon pficon-warning-triangle-o';
+    } else {
+      return '';
+    }
+  }
+
+}
--- a/ui/src/app/lib/atlasmap-data-mapper/data.mapper.module.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/data.mapper.module.ts
@@ -63,6 +63,8 @@ export { ConfigModel } from './models/config.model';
 export { MappingModel } from './models/mapping.model';
 export { MappingSerializer } from './services/mapping.serializer';
 
+import { ToErrorIconClassPipe } from './common/to-error-icon-class.pipe';
+
 export { DataMapperAppComponent } from './components/data.mapper.app.component';
 
 @NgModule({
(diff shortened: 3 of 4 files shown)


igarashitm:
-1

This is just showing an warning icon anyway, for every validation results, which contains error/warn/info.

akieling:
Did I misunderstand what needs to be done?

igarashitm:
You need to check error level and choose the right icon, red circle for error, orange triangle for warn, ignore info.

akieling:
Any tips on creating mappings with errors and warnings?

igarashitm:
I'm always clicking around the example fields, easily find the validation error/warn in 10 secs...

akieling:
Ready to be reviewed.

jpav:
LGTM too, but I do have a complaint about the way the icons look.  Is it just me, or does the error icon look more like a remove/delete icon?  Seems like it would look better, and be more consistent with other apps, if the warning/error (maybe info) icons had solid colored background with white/black text for 'x' or '!' symbol. Assuming anyone else agrees, I'll log this as a separate issue.

igarashitm:

ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.list.component.ts
```diff
@@ -127,8 +127,10 @@ export class MappingListFieldComponent {
                                     <div class="clear"></div>
                                 </div>
                                 <div class="error">
-                                    <span class="pficon pficon-error-circle-o"
-                                    *ngIf="mapping.validationErrors.length"></span>
+                                    <span class="pficon"
+                                          [ngClass]="{'pficon-error-circle-o': mapping.hasValidationErrors(),
+                                                      'pficon-warning-triangle-o': mapping.hasValidationWarnings()}"
+                                          *ngIf="mapping.validationErrors.length"></span>
```
deeleman: On every cycle of Angular's ChangeDetector mechanism, you'll be re-executing `mapping.hasValidationErrors()` and `mapping.hasValidationWarnings()`, which will execute subsequent functions inside the `MappingModel` class such as `getValidationErrors ()` and `getValidationWarnings()` which will also execute the `this.validationErrors.filter` functional lambda and the like... These are too many executions happening on every tick. Now multiply this for an application with literally hundreds of components (which AtlasMapper might become one day) and you'll bump into a severe performance bottleneck down the track.

Let me suggest a more convenient and Angular-friendly approach for this: Use a pure`@Pipe` instead. Something like this: 

```
<span class="pficon" [class.pficon-error-circle-o]="mapping.validationErrors | hasErrors" ...> 
``` 

Where `hasErrors` is a pipe that will receive `mapping.validationErrors` and will inspect if there is any actual error by returning `mapping.validationErrors.some(e => e.level >= ErrorLevel.ERROR)`.

The only caveat is that you'll want to alter the memory reference for `mapping.validationErrors` every time you update it, so it becomes immutable:

```
addValidationError(message: string) {
  const e = new ErrorInfo(message, ErrorLevel.VALIDATION_ERROR);
  this.validationErrors = [... this.validationErrors, e];
}
``` 

BTW, since `public` is implied in TypeScript entities, the industry convention is to never annotate class members with it   (which unfortunately we do not honor here)

akieling: Does your suggestion reduce the number of times we evaluate the `validationErrors` array or does it just reduce the method calls? To improve performance, wouldn't be better to create an array of view objects with only the data we need to show on that table? We would have a simple string field with the class name already set on each object.

deeleman: My proposal aims to minimize function calls, which subsequently will reduce the amount of times we evaluate `validationErrors` under the hood. 

Taking this approach will make our application operate all the way around. The template and its component controller will not evaluate `MappingModel.getValidationErrors()` anymore. On the contrary they will evaluate just nothing after content is checked. In the very moment that the ChangeDetection system detects that `MappingModel.validationErrors` has changed its type reference, an OnPush operation will be triggered throughout the component tree and will force the pipe to return a new value.

Encapsulating this conditional logic in a pipe also has an additional advantage: we can reuse this  logic elsewhere at zero cost.

akieling: OK, got it. By referencing `MappingModel.validationErrors` directly instead of a method, Angular can detect when it changes and evaluate it only at that point, correct?

deeleman: >By referencing `MappingModel.validationErrors` directly instead of a method, Angular can detect when it changes

Yes, but here I'm proposing to funnel the conditional validation through a pipe, and this requires taking something into consideration: Unless defined as IMPURE, pipes are only affected by the ChangeDetector system when the input argument (`validationErrors`) changes its type reference. This is perfectly fine when such input is a primitive type (strings, numbers) but `validationErrors` is an array, and adding or removing items will not change its type reference. It will become the same instance object and therefore ChangeDetection will not happen. That's why I pointed out to create a new object by means of ES6's spread operator.

igarashitm:
@akieling it seems you still have lint error on this one?

akieling:
@igarashitm fixed

igarashitm:
Hi @deeleman , would you have a look and approve if it looks OK? thanks!

igarashitm:
oops, it has a merge conflict - @akieling can you rebase this against current upstream master?

akieling:
@igarashitm done

PR #198, opened 2017-12-14
fix: StringConverter should be more forgiving for conversion to Boolean

sunilmeddr:
Fixes: #51


igarashitm:
Hi @rohanmars , can you make sure it's not breaking consistency in the reference-mappings?

rohanmars:
@sunilmeddr I don't think should cause the reference mappings to change, unless necessary. Additionally, we should return null if he input is null I believe (else change it this behavior or all converters).

sunilmeddr:
@rohanmars, I will update the code to return null,if the input is null

sunilmeddr:
I updated the data used in the reference mappings referred in the test cases as test cases are failing.

rohanmars:
@sunilmeddr why the change's in the reference mapping's values? I think these should be reverted back to the originals.

sunilmeddr:
@rohanmars the following method was used earlier to convert a string to boolean and it will not properly convert the string to the boolean. 
ex:   if the input to the method is 8, method returns true a return value,even though it has to return false
After fixing the issue,  several test cases are failed as the test data in the mapping is added as per the old functionality. So I updated the test data in the mappings.

 public Boolean convertToBoolean(String value) throws AtlasConversionException {
        if (value == null) {
            return null;
        }
        if ("0".equals(value) || "f".equalsIgnoreCase(value) || "false".equalsIgnoreCase(value)) {
            return Boolean.FALSE;
        } else if ("1".equals(value) || "t".equalsIgnoreCase(value) || "true".equalsIgnoreCase(value)) {
            return Boolean.TRUE;
        }
        // fall back
        if ("".equals(value)) {
            return Boolean.FALSE;
        } else {
            return Boolean.TRUE;
        }

rohanmars:
I think the value "8" should return true, it doesn't confirm to a convention so it merely tests for a non-empty string. It's arbitrary whether this is really true or false. However, the converter should test for known values t/f/true/false/0/1 etc. However all other values should be converted to true for a non empty string IMHO. Think of it like testing for a non-empty string if it doesn't convert using standard values.

sunilmeddr:
I updated the code as per the below comment from the @jpav . Do we just need to update the if condition so that the method return true if the input is "true" and return false if the input is "false"

https://github.com/atlasmap/atlasmap/issues/51

We should make the action work like the java Boolean wrapper in that values like "true", "1", "t", "y", and "yes" result in a Boolean value of TRUE, and all else result in FALSE, with nothing returning a conversion error.

jpav:
The code in this commit seems like the correct behavior to me. @rohanmars, is there a precedent behind wanting values that aren't clearly "true" values to be treated as true?

Issue #186, opened 2017-12-10
Review JavaValidationService methods visibility and remove unused


igarashitm:
See Epic #183

Issue #185, opened 2017-12-10
Review ClassFinder/StringUtil/ClassHelper utility methods and remove unused


igarashitm:
As ClassFinder is just a single utility method, I prefer migrating into ClassHelper. Then review ClassHelper methods if they are really in use and remove unused.
StringUtil also looks like Java inspection specific utility methods, could be migrated into ClassHelper.

This is a side story of Epic #183

Issue #184, opened 2017-12-10
Review JavaConstructService methods visibility and remove unused


igarashitm:
See Epic #183

Issue #170, opened 2017-12-07
Review visibility of ClassInspectionService methods


igarashitm:
It looks internal subroutines are exposed as public or protected. And some of them are even test only method, which must be removed. If so, change to private. Testing should be handled via public methods which really need to be exposed for the real usage, but not only for test.

PR #201, opened 2017-12-15
fix: Fixed method visibility of ClassInspectionService, JavaConstructService, JavaValidationService,StringUtil,ClassHelper and migrated ClassFinder into ClassHelper

sunilmeddr:
fix: Review visibility of ClassInspectionService methods
fix: Review JavaConstructService methods visibility and remove unused
fix: Review ClassFinder/StringUtil/ClassHelper utility methods and remove unused
fix: Review JavaValidationService methods visibility and remove unused

Fixes: #170
Fixes: #184
Fixes: #185
Fixes: #186


runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/StringUtil.java
```diff
@@ -24,7 +24,7 @@ public static boolean isEmpty(String s) {
         return s == null || "".equals(s.trim());
```
igarashitm: For me this is also a Java specific utilities and could be migrated into ClassHelper, WDYT @sunilmeddr ?

runtime/modules/java/core/src/test/java/io/atlasmap/java/inspect/ClassInspectionServiceTest.java
```diff
@@ -76,37 +75,6 @@ public void testDetectArrayClass() {
         assertEquals(Map.class, classInspectionService.detectArrayClass(Map[][][][].class));
     }
 
-    @Test
```
igarashitm: I think these test cases should be covered from somewhere else. Same tests should be run via public methods.

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
```diff
@@ -86,11 +86,6 @@ public void init() {
         versionMap.put("1.1", 45);
     }
 
-    public void destroy() {
```
igarashitm: why removing destroy()?

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
```diff
@@ -106,14 +101,15 @@ protected String getModuleFieldName(JavaField field) {
...
     }
 
-    protected void validateSourceAndTargetTypes(String mappingId, Field inputField, Field outField, List<Validation> validations) {
+    protected void validateSourceAndTargetTypes(String mappingId, Field inputField, Field outField,
+            List<Validation> validations) {
         if ((inputField instanceof JavaField && outField instanceof JavaField)
                 && ((inputField.getFieldType() == null || outField.getFieldType() == null)
-                    || (inputField.getFieldType().compareTo(FieldType.COMPLEX) == 0
-                    || outField.getFieldType().compareTo(FieldType.COMPLEX) == 0))) {
+                        || (inputField.getFieldType().compareTo(FieldType.COMPLEX) == 0
```
igarashitm: I know it's not your code, but doesn't it look weird to use `compareTo()` here? I would use just `==` for the Enum

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
```diff
@@ -201,7 +199,7 @@ private void validateClass(String mappingId, JavaField field, List<Validation> v
         }
     }
 
-    protected Integer detectClassVersion(String className) {
+    Integer detectClassVersion(String className) {
```
igarashitm: should it be package private? for what purpose? It just looks like a private routine though.

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -777,15 +699,15 @@ protected Integer detectArrayDimensions(Class<?> clazz) {
...
+    private Class<?> detectListClass(ClassLoader classLoader, Field field) throws ClassNotFoundException {
         List<String> types = detectParameterizedTypes(field, true);
         if (types != null && !types.isEmpty()) {
             return classLoader.loadClass(types.get(0));
         }
         return null;
     }
 
-    protected Class<?> detectArrayClass(Class<?> clazz) {
+    Class<?> detectArrayClass(Class<?> clazz) {
```
igarashitm: why package private?

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -55,143 +52,71 @@
...
+    private Boolean getDisablePackagePrivateOnlyFields() {
         return disablePackagePrivateOnlyFields;
     }
 
-    public void setDisablePackagePrivateOnlyFields(Boolean disablePackagePrivateOnlyFields) {
-        this.disablePackagePrivateOnlyFields = disablePackagePrivateOnlyFields;
-    }
-
-    public Boolean getDisablePrivateOnlyFields() {
+    private Boolean getDisablePrivateOnlyFields() {
```
igarashitm: getter/setter should be public

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -55,143 +52,71 @@
...
+    private Boolean getDisableProtectedOnlyFields() {
         return disableProtectedOnlyFields;
     }
 
     public void setDisableProtectedOnlyFields(Boolean disableProtectedOnlyFields) {
         this.disableProtectedOnlyFields = disableProtectedOnlyFields;
     }
 
-    public Boolean getDisablePackagePrivateOnlyFields() {
+    private Boolean getDisablePackagePrivateOnlyFields() {
```
igarashitm: getter/setter should be public

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -55,143 +52,71 @@
...
-    public List<String> getClassNameBlacklist() {
-        return this.classNameBlacklist;
-    }
-
     public List<String> getFieldBlacklist() {
         return this.fieldBlacklist;
     }
 
-    public Boolean getDisableProtectedOnlyFields() {
+    private Boolean getDisableProtectedOnlyFields() {
```
igarashitm: getter/setter should be public

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -55,143 +52,71 @@
...
     private Boolean disablePrivateOnlyFields = false;
     private Boolean disablePublicOnlyFields = false;
     private Boolean disablePublicGetterSetterFields = false;
 
-    public List<String> getListClasses() {
+    private List<String> getListClasses() {
         return this.listClasses;
     }
 
-    public List<String> getClassNameBlacklist() {
```
igarashitm: I don't think we have to remove getters even if it's not used today...

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -55,143 +52,71 @@
...
     private List<String> fieldBlacklist = new ArrayList<String>(Arrays.asList("serialVersionUID"));
-    private List<String> classNameBlacklist = new ArrayList<String>();
     private Boolean disablePackagePrivateOnlyFields = false;
     private Boolean disableProtectedOnlyFields = false;
     private Boolean disablePrivateOnlyFields = false;
     private Boolean disablePublicOnlyFields = false;
     private Boolean disablePublicGetterSetterFields = false;
 
-    public List<String> getListClasses() {
+    private List<String> getListClasses() {
```
igarashitm: getter/setter should be public

igarashitm:
And there is a conflict, please rebase on top of upstream master

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
sunilmeddr: It has no references

igarashitm: I would think this kind of lifecycle methods are added for future use...

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
sunilmeddr: out this class it is referenced only in the testcases. Shall I use the protected in these kind of cases

runtime/modules/java/core/src/test/java/io/atlasmap/java/inspect/ClassInspectionServiceTest.java
sunilmeddr: Can you please elaborate

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
sunilmeddr: out this class this method  is referenced only in the testcases. Shall I continue using the protected

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
sunilmeddr: Class variable classNameBlacklist was removed as it has no references. So getClassNameBlacklist also removed.

sunilmeddr: Class variable classNameBlacklist was removed as it has no references,so corresponding getxxx method also removed

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
igarashitm: Then this should be private and test should be modified to run from public methods... that's the primary goal for this issues...

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/StringUtil.java
sunilmeddr: All the string based utility methods are in the classStringUtil.java . This is one of the utility method.

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
sunilmeddr: Its not look like a framework method as this method is not in the parent class BaseModuleValidationService.java. I will add it back to make sure we dont get any issues

igarashitm:
Also, can you describe about the commit **in the first line**, instead of `This commit contain multiple fixes` and having following lines, like this
```
fix: Fixed method visibility of ClassInspectionService, JavaConstructService, JavaValidationService and migrated all utilities into ClassHelper

Fixes: #...
```

runtime/modules/java/module/src/main/java/io/atlasmap/java/module/JavaValidationService.java
sunilmeddr: Then I think,in this case we can remove the testcase

sunilmeddr:
Commits from another PR updated the same file,that this pull request tried to merge. I have resolved the issue,by merging the 2 files.

sunilmeddr:
Can somebody review the changes

sunilmeddr:
Can anyone review the changes

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassHelper.java
```diff
@@ -26,22 +28,26 @@
...
         opts.add(isMethodNameFromFieldName(fieldName));
         return opts;
     }
 
-    public static String setMethodNameFromFieldName(String fieldName) {
-        return "set" + StringUtil.capitalizeFirstLetter(fieldName);
-    }
-
-    public static String getMethodNameFromFieldName(String fieldName) {
+    private static String getMethodNameFromFieldName(String fieldName) {
```
rohanmars: why are these private static? seems strange.

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
rohanmars: agreed, should be public

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/StringUtil.java
```diff
@@ -24,7 +24,7 @@ public static boolean isEmpty(String s) {
...
     }
 
-    public static String capitalizeFirstLetter(String sentence) {
+    static String capitalizeFirstLetter(String sentence) {
```
rohanmars: why the removal of public keyword for these methods?

sunilmeddr:
@rohanmars , you may be looking at the wrong commit. These issues are already fixed.

https://github.com/atlasmap/atlasmap/pull/201/commits/3e8641a10d5e57de9bfe8cf16faf9a1858bd2d34

sunilmeddr:
I am not able to squash, as in between the commits, I merged the code from
the master due the conflicts( Some other pull request modified the same
file)

rohanmars:
okay. In the future do a rebase from master to your branch instead. This way it will keep your commits "on top" and any merge conflicts will be in a new commit (on top) which you can then squash into your work.

rohanmars:
LGTM :+1:

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -66,7 +63,6 @@
             "java.util.concurrent.ConcurrentHashMap", "java.util.concurrent.ConcurrentMap"));
     private AtlasConversionService atlasConversionService = null;
     private List<String> fieldBlacklist = new ArrayList<String>(Arrays.asList("serialVersionUID"));
-    private List<String> classNameBlacklist = new ArrayList<String>();
```
igarashitm: I thought we agreed several times with @mattrpav that the class name black list is a place holder for the future use and we shouldn't remove it.

runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
```diff
@@ -101,10 +93,6 @@ public Boolean getDisablePackagePrivateOnlyFields() {
         return disablePackagePrivateOnlyFields;
     }
 
-    public void setDisablePackagePrivateOnlyFields(Boolean disablePackagePrivateOnlyFields) {
```
igarashitm: setter

runtime/modules/java/core/src/test/java/io/atlasmap/java/inspect/ClassInspectionServiceTest.java
```diff
@@ -51,62 +46,6 @@ public void tearDown() {
         classInspectionService = null;
     }
 
-    @Test
```
igarashitm: Yet again, these test cases needs to be tested via `inspectClass()`

mattrpav:
@igarashitm yes, I asked Rohan to make a first pass review as my schedule did not allow this week as originally planned. I’m working with Sunil to update the PR to include those changes.

mattrpav:
@igarashitm those two changes have been incorporated

Issue #16, opened 2017-11-15
Increate unit test coverage of atlas-core


igarashitm:
@mattrpav commented on [Wed Jul 19 2017](https://github.com/atlasmap/atlasmap-runtime/issues/36)

- [x] DefaultAtlasCombineStrategy (100%)
- [ ] DefaultAtlasContext
- [ ] DefaultAtlasContextFactory
- [ ] DefaultAtlasConversionService (74%)
- [x] DefaultAtlasPropertyStrategy (84%)
- [x] DefaultAtlasSeparateStrategy (100%)
- [x] DefaultAtlasSession (100%)
- [x] DefaultAtlasValidationService (88%)
- [x] StringDelimiter (86%)

- [x] io.atlasmap.actions (85%)
- [x] io.atlasmap.converters (99%)
- [x] io.atlasmap.validators (70%)

igarashitm:
This was unexpectedly closed as a side effect of migration - reopening

rohanmars:
closed by #204

PR #204, opened 2017-12-18
chore: added test cases to increase unit test coverage of atlas-core

joeylimd:
chore: added test cases to increase unit test coverage of atlas-core

Relates to: #16

### DIFF
--- a/runtime/core/pom.xml
+++ b/runtime/core/pom.xml
@@ -72,4 +72,35 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    
+    <build>
+    	<plugins>
+    		<plugin>
+				<artifactId>maven-dependency-plugin</artifactId>
+				<executions>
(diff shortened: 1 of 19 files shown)


runtime/core/pom.xml
```diff
@@ -72,4 +72,35 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    
+    <build>
```
igarashitm: Why is this required? atlas-mode is already in dependency

joeylimd: Hi Tomo,
A jar file is needed to test AtlasUtil.findClassesFromJar method. copied atlas-model jar into target folder for Unit testing. Thanks

igarashitm: Ah OK... not really ideal to make POM fat with test only configuration though, I'd say OK for now.

Issue #207, opened 2017-12-20
AtlasEndpoint : There's no source document with docId...


igarashitm:
Just an eyesore. It should map source document with appropriate docId
```
mentions2todo-2-zltv9 mentions2todo 2017-12-20 13:10:56.736  WARN 1 --- [line://MENTIONS] o.a.c.component.atlasmap.AtlasEndpoint   : There's no source document with docId='twitter4j.Status', returning default: docId='null', path='null'
mentions2todo-2-zltv9 mentions2todo 2017-12-20 13:10:57.300  WARN 1 --- [line://MENTIONS] o.a.c.component.atlasmap.AtlasEndpoint   : There's no source document with docId='twitter4j.Status', returning default: docId='null', path='null'
mentions2todo-2-zltv9 mentions2todo 2017-12-20 13:10:57.318  WARN 1 --- [line://MENTIONS] o.a.c.component.atlasmap.AtlasEndpoint   : There's no source document with docId='twitter4j.Status', returning default: docId='null', path='null'
mentions2todo-2-zltv9 mentions2todo 2017-12-20 13:10:57.331  WARN 1 --- [line://MENTIONS] o.a.c.component.atlasmap.AtlasEndpoint   : There's no source document with docId='twitter4j.Status', returning default: docId='null', path='null'
mentions2todo-2-zltv9 mentions2todo 2017-12-20 13:10:57.407  WARN 1 --- [line://MENTIONS] o.a.c.component.atlasmap.AtlasEndpoint   : There's no source document with docId='twitter4j.Status', returning default: docId='null', path='null'
mentions2todo-2-zltv9 mentions2todo 2017-12-20 13:10:57.428  WARN 1 --- [line://MENTIONS] o.a.c.component.atlasmap.AtlasEndpoint   : There's no source document with docId='twitter4j.Status', returning default: docId='null', path='null'
mentions2todo-2-zltv9 mentions2todo 2017-12-20 13:10:57.459  WARN 1 --- [line://MENTIONS] o.a.c.component.atlasmap.AtlasEndpoint   : There's no source document with docId='twitter4j.Status', returning default: docId='null', path='null'
```

KurtStam:
@igarashitm with this fix can I close https://github.com/syndesisio/syndesis/issues/792?

igarashitm:
@KurtStam yep, 1.32.2 released today contains this fix.

PR #211, opened 2017-12-21
fix: AtlasEndpoint : There's no source document with docId...

igarashitm:
Fixes: #207


igarashitm:
@atlasmap/runtime could anybody review this? thanks!

PR #217, opened 2017-12-29
chore: Increase unit test code coverage of Xml FieldReader and Xml FieldWriter

joeylimd:
Relates to: #203

### DIFF
--- a/runtime/modules/xml/core/src/test/java/io/atlasmap/xml/core/XmlFieldReaderTest.java
+++ b/runtime/modules/xml/core/src/test/java/io/atlasmap/xml/core/XmlFieldReaderTest.java
@@ -16,6 +16,7 @@
 package io.atlasmap.xml.core;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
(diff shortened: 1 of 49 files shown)


runtime/modules/xml/core/src/test/java/io/atlasmap/xml/core/XmlFieldReaderTest.java
```diff
@@ -359,4 +366,673 @@ private String getDocumentString(String uri) throws IOException {
         fis.close();
         return new String(buf);
     }
+
+    @Test
+    public void testXmlFieldDoubleMax() throws Exception {
+        String fieldPath = "/primitive/value";
```
rohanmars: Can we make some private/protected methods for repetitious code? There seems to a few standard code patterns in use in the file. Make some methods which parametrize differing values.

joeylimd: Done. extracted private method for repetitive codes. Please review. Thanks.

rohanmars:
LGTM :+1:

rohanmars:
Joey, changes look good. I wonder if it's worth adding some tests to validate negative scenarios of the read and s etDocument methods. For example, if incorrect xml is passed. The idea would be test situations which should generate exceptions. What do you think?

rohanmars:
@joeylimd , the changes look good and meet the requirements in @203.

LGTM :+1:

Issue #218, opened 2017-12-29
ActionsJsonDeserializer: in case of "Add", it should return new Add();"Concatentate" should be "Concatenate"


joeylimd:
in class ActionsJsonDeserializer.java, 
1. in case "Add", line 69 "return new Average();" should "return new Add();"
2. line 82: a misspelling - "Concatenate" should be "Concatenate"

PR #219, opened 2017-12-29
fix: ActionsJsonDeserializer: in case of "Add", it should return new Add();"Concatentate" should be "Concatenate"

joeylimd:
Fixes: #218


rohanmars:
LGTM :+1:

PR #223, opened 2018-01-04
fix: XmlFieldReader test fails with the method really in use

joeylimd:
Fixes: #141

### DIFF
--- a/runtime/modules/xml/core/src/test/java/io/atlasmap/xml/core/XmlFieldReaderTest.java
+++ b/runtime/modules/xml/core/src/test/java/io/atlasmap/xml/core/XmlFieldReaderTest.java
@@ -29,7 +29,6 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.Map;
 
-import org.junit.Ignore;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


runtime/modules/xml/core/src/test/java/io/atlasmap/xml/core/XmlFieldReaderTest.java
```diff
@@ -223,19 +216,20 @@ public void testReadDocumentElementWithMultipleNamespaceComplexConstructorArg()
         namespaces.put("http://www.example.com/x/", "");
 
         reader = new XmlFieldReader(DefaultAtlasConversionService.getInstance(), namespaces);
```
rohanmars: This should probably use a different instance, not "reader", as other tests reuse that instance and could causes unintended consequences down the road. It appears that you just want a reader for the purposes of multiple namespaces. Additionally, the setDocument() is at the start of the function but it is reset after a new reader is created.

joeylimd: Done. Updated this file with feedback. Please review again. Thanks.

rohanmars:
LGTM :+1:

Issue #189, opened 2017-12-11
Correct indentation of all the XML/JSON/yaml and etc. files


igarashitm:
pom.xml, log4j2.xml, mapping files, test data, etc. Right now it's mixup of 2-spaces and 4-spaces. sometimes even `<TAB>` is used.

@mattrpav summarzing comments up to 12/19

1. Update the atlas runtime build to use the mojo xml check to validate indentation
 - http://www.mojohaus.org/xml-maven-plugin/check-format-mojo.html
2. Update the EditorConfig to enforce indentation and spacing of non-Java files
 - ref: [EditorConfig reference](http://editorconfig.org/)

igarashitm:
I would vote on 2-spaces, but OK with 4-spaces. No <TAB>. @atlasmap/runtime thoughts?

rhuss:
👍 for 2 spaces in XML, 4 in Java (that's what we use in syndesis)

jpav:
I'd vote similarly: 4 in Java, 2 in everything else.  No tabs.

igarashitm:
OK, so the indent size is:
* Java: 4-spaces, no TAB
* Others: 2-spaces, no TAB
  * XML, JSON, yaml, what else?

I think I fixed Java indentation to 4-spaces while back, but might be good to add checkstyle rule to enforce.

igarashitm:
To everybody in @atlasmap/runtime ,

As we are working concurrently, fixing indentation at once would cause conflicts. Rather I'd like to ask you all to correct indentation when you have a chance to edit the file.

* Java: 4-spaces, no TAB
* Others: 2-spaces, no TAB
  * XML, JSON, yaml, sh, etc

However, 
### please do it in dedicated commit
You can put multiple commits in one pull request. Please separate indentation commit from anything else. That helps reviewer to see what is in that pull request, and check the indentation and the real fix separately.

igarashitm:
It seems 4-spaces is used for shell script in Syndesis. I'd go with it.

mattrpav:
In an effort to align dev tools-- which maven plugin or tool does Syndesis utilize to perform the spacing?

At one point Hiram had created an EditorConfig file to help with enforcement across IDE's. It might be good to revisit that to provide enforcement at development time. Thoughts?

igarashitm:
We're already using checkstyle and Google showed this for XML - http://www.mojohaus.org/xml-maven-plugin/check-format-mojo.html

[...]

PR #236, opened 2018-01-17
fix: Correct indentation of all the XML/JSON/yaml and etc. files

joeylimd:
Fixes: #189


runtime/itests/reference-mappings/src/test/java/io/atlasmap/reference/java_to_java/JavaJavaComplexListTest.java
```diff
@@ -102,7 +102,7 @@ public void testGenerateListMappingBasic() throws Exception {
         AtlasMappingService atlasMappingService = new AtlasMappingService(
                 Arrays.asList("io.atlasmap.v2", "io.atlasmap.java.v2", "io.atlasmap.xml.v2"));
         atlasMappingService.saveMappingAsFile(a,
-                new File("src/test/resources/javaToJava/atlasmapping-complex-list-autodetect-base.xml"));
+                new File("target/atlasmapping-complex-list-autodetect-base.xml"));
```
rohanmars: why have these paths changed?

joeylimd: 1. In the java-generated xml files, the indentation is 4 spaces, which does not align with the new standard - 2 spaces
2. Those are existing xml files. Unit tests should not modify pre-defined testing files under /src/test folder. All new output files should be saved under target folder.
3. If the path is not changed, the new generated xml files with 4-space indentation will break code format validation. That's why the path got changed.

droneware: The tests should not be overriding source files in the first place. The paths should keep the original output structure but within the target directory. I.e. target/reference-mappings/javaToJava/......xml

rohanmars: The tests should not be overriding source files in the first place. The paths should keep the original output structure but within the target directory. I.e. target/reference-mappings/javaToJava/......xml

joeylimd: Done. Generating output files within target sub directory. Please review.

rohanmars:
@joeylimd, can you rebase this branch and resolve any conflicts.

rohanmars:
LGTM.  :+1: 

* only a random selection of reference mappings was preformed due to the number of files.

rohanmars:
@igarashitm would you want the xml-maven-plugin to error out if there are errors in the xml files?  Or just a warning and move on?

igarashitm:
Could it easily exclude the intentionally invalid XML files for test purposes?

igarashitm:
It might be better exclude everything in src/test/resources

joeylimd:
1. to exclude a specific new invalid XML files, a new entry needs to be added in runtime POM.xml every time, e.g. <exclude>pathToFile/foo.ff</exclude>. 
2. the issue is caused by file atlasmap\runtime\service\src\test\resources\mappings\bar.xml and foo.xml although they are excluded. xml-mven-plugin still reads these 2 files, then throws Exception Premature end of file, which breaks build as per config failOnFormatViolation is true. from this thread, https://stackoverflow.com/questions/10022796/why-am-i-getting-this-error-premature-end-of-file, sounds it's a bug of this plugin.
3. If failOnFormatViolation is false, it generate warning messages without breaking build.

igarashitm:
That sounds like something we don't want to enable at all. Even warnings are annoying.

rohanmars:
Maybe put the plugin declaration in new maven profile in the pom.xml. That way it will only be enabled when specifically activated.

igarashitm:
I'm OK with the profile. But can you also completely exclude any src/test/resources? I think we don't care if test resources are well formatted, and adding each intentional error files into POM is really annoying.

joeylimd:
1. in profile, excluded files in src/test/resources, and adding each intentional error files into POM is not needed.
2. The roadblock with xml-maven-plugin is that XML can't be excluded, but all other type of files can be excluded. With profile, if needed, it can be enabled with **mvn -Pxml-format**.
Is it fine to move on? Thanks again.

igarashitm:
Yes that sounds reasonable. Also please don't forget to rebase against current upstream master :-)

igarashitm:
Ah hold on please, I want to rollback the version number to `1.33-SNAPSHOT` in the POM.

joeylimd:
@igarashitm sure. notice there are many conflicts in POM file. if rollback to 1.33-snapshot, that will save a lot of time. Sweet... when done, please post it. Thanks.

igarashitm:
Ha, look at this! conflicts were just that version number :-)

joeylimd:
@igarashitm Thanks a bunch again. It's ready to get merged.
1. Totally 484 files got formatted. No files are deleted this time.
2. To check xml format with profile xml-format, run **mvn -Pxml-format clean install** 
3. To not check xml format, normally run mvn clean install
4. To generate warning messages only, instead of format-check failure, please set **failOnFormatViolation** as **false** in **atlas-runtime POM.xml**

Issue #6, opened 2017-11-15
TypeConverter other than for primitives are not supported yet


igarashitm:
@igarashitm commented on [Mon Oct 30 2017](https://github.com/atlasmap/atlasmap-runtime/issues/248)

https://github.com/atlasmap/atlasmap-runtime/blob/ce498616965254723428641ab5c03ed80cff4d40/atlas-core/src/main/java/io/atlasmap/core/DefaultAtlasConversionService.java#L279-L306

Yes it's definitely TODO. Even [DateConverter](https://github.com/atlasmap/atlasmap-runtime/blob/ce498616965254723428641ab5c03ed80cff4d40/atlas-core/src/main/java/io/atlasmap/converters/DateConverter.java) is not available.

rohanmars:
@igarashitm is this issue only focused on adding DateConverter support?

igarashitm:
@rohanmars nope, this is to implement exactly this `TODO`
https://github.com/atlasmap/atlasmap-runtime/blob/ce498616965254723428641ab5c03ed80cff4d40/atlas-core/src/main/java/io/atlasmap/core/DefaultAtlasConversionService.java#L304-L305

`DateConverter` is just an example (well, right now it's the only one though...). All the OOTB converters other than primitives should be invoked from here.

rohanmars:
@igarashitm I have implemented an approach that uses the annotation (AtlastConversionInfo) values to determine which conversion method to actually use in the conversion service. Is this an okay approach with you? It will simplify the solution I believe and help on #114 

Basically, when the conversion service loads the converter classes it now builds a map of converter methods using reflection, keyed by source/target field types. Then when the converterType method is called it looks up which method it should invoke from the map and uses the saved Method reference to invoke using refection. That way new conversion methods can be added simply by adding methods to the AtlasConverter class and annotating them accordingly. No update to the converter service would then be required.

I might need to slightly change the key object include the class names also (if they are specified on the annotation) to support multiple use cases that map to the COMPLEX fileldType but having different concrete impl classes. This is currently the case for the existing DateConverter, however, I think that this class needs to be improved anyway so that might not be required for that specific converter but probably should be supported anyway.

I was thinking also to remove the AtlasPrimativeConverter interface, I think that is unnecessary now. Do you have any thoughts as to why we should keep it if I change to use the reflection based approach?

rohanmars:
@mattrpav what you think about my proposed changes?

[...]

Issue #205, opened 2017-12-18
Fix date/time related runtime code so fields show up correctly in UI


igarashitm:
Part of epic #248 

See #200 

While field reader/writer can use individual field types like `FieldType.DATE_TIME_TZ`, it's annoying creating all the overload methods for all of those date/time related field types. Introduce `FieldType.ANY_DATE` which works as an umbrella field type so TypeConverter and FieldAction can consume/produce all those fields.

igarashitm:
@jpav can you take a look at this?

PR #237, opened 2018-01-18
fix: date converter support, boolean converter fix, non-primative con…

rohanmars:
…verter support

### DIFF
--- a/runtime/api/src/main/java/io/atlasmap/api/AtlasConversionService.java
+++ b/runtime/api/src/main/java/io/atlasmap/api/AtlasConversionService.java
@@ -15,7 +15,6 @@
  */
 package io.atlasmap.api;
 
-import java.lang.reflect.Method;
 import java.util.Optional;
 
 import io.atlasmap.v2.FieldType;
@@ -26,13 +25,15 @@ public interface AtlasConversionService {
 
     Optional<AtlasConverter<?>> findMatchingConverter(String sourceClassName, String targetClassName);
 
-    Optional<Method> findMatchingMethod(FieldType source, FieldType target, AtlasConverter<?> customConverter);
-
     Object copyPrimitive(Object sourceValue);
 
-    Object convertType(Object sourceValue, FieldType sourceType, FieldType targetType) throws AtlasConversionException;
+    Object convertType(Object sourceValue, FieldType origSourceType, FieldType targetType)
+            throws AtlasConversionException;
 
-    Object convertType(Object sourceValue, FieldType sourceType, FieldType targetType, String customClassName)
+    Object convertType(Object sourceValue, String sourceFormat, FieldType targetType, String targetFormat)
+            throws AtlasConversionException;
+
+    Object convertType(Object sourceValue, String sourceFormat, Class targetType, String targetFormat)
             throws AtlasConversionException;
 
     Class<?> boxOrUnboxPrimitive(Class<?> clazz);
--- a/runtime/api/src/main/java/io/atlasmap/spi/AtlasConversionConcern.java
+++ b/runtime/api/src/main/java/io/atlasmap/spi/AtlasConversionConcern.java
@@ -18,11 +18,11 @@ package io.atlasmap.spi;
 /**
  */
 public enum AtlasConversionConcern {
-    NONE("none", "Conversion from '%s' to '%s' is supported"),
-    RANGE("range", "Conversion from '%s' to '%s' can cause out of range exceptions"),
-    FORMAT("format", "Conversion from '%s' to '%s' can cause numeric format exceptions"),
-    CONVENTION("format", "Conversion from '%s' to '%s' uses a default convention for values"),
-    UNSUPPORTED("unsupported", "Conversions from '%s' to '%s' is not supported");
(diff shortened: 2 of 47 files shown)


igarashitm:
I thought we talked about that introducing `format` thing is too complex for the TypeConverter? At least can you file an another issue and describe what is it going to achieve? And like I said, I don't think we want to merge it before GA, while fixing #6 bug is still important.

BTW I guess DateConverter is eventually blocked by the #205 right? it's not gonna be triggered until #205 is addressed ...

rohanmars:
Talking with @mattrpav  he agreed that support a format attribute is essential to date converter support. I think we need to lay out the case for it a bit clearer.

I don't think this PR or #6 is blocked by #205. Probably the other way around. The important change, IMHO, of this PR is that it introduces general non primitive type conversion support. Prior, the converters where hard coded. I changed this now to be a dynamic map of methods, keyed by source and target java types. So simply declaring an converter (via annotation) will cause it to be picked up. Which is necessary for the number conversion to. Then at run time these are dynamically invoked upon. 

How FieldTypes map to java (internal) objects is a separate question. One which I struggled with and is not full addressed in this PR. However, for the java to java use case this is okay now. The issues are more on the xml/json side I think, specifically when type is determined in the reader.

I think this PR will be necessary for #205.

mattrpav:
@igarashitm I believe it is essential for supporting date and time formats in non-Java document types (Json, xml, etc). 

When parsing for input or writing for output a date format expression is going to be required, since the data is effectively “just a string” in both xml and Json. 

We’d be happy to hop on a call and discuss further as well

igarashitm:
I meant,

> support a format attribute is essential to date converter support. I think we need to lay out the case for it a bit clearer.

I want to separate this one from #6 before discuss. Too bigger topic to inline within #6 bug fix. Note that we don't yet support `format` property in the JSON schema, which is required to be taken into account for date/time format handling.

> I don't think this PR or #6 is blocked by #205. Probably the other way around. The important change, IMHO, of this PR is that it introduces general non primitive type conversion support. Prior, the converters where hard coded. I changed this now to be a dynamic map of methods, keyed by source and target java types. So simply declaring an converter (via annotation) will cause it to be picked up. Which is necessary for the number conversion to. Then at run time these are dynamically invoked upon.

I think this is the one we expect for #6.

BTW I meant *DateConverter* is blocked by the #205, so it can't be utilized from the mapping generated via UI, unless #205 is addressed.

igarashitm:
We might not have introduced date/time related atlasmap primitives even...

igarashitm:
Also note that the format thing involves UX design, which makes it even larger. I don't like introduce runtime only thing at this stage.

mattrpav:
@igarashitm given the timing, may I suggest that a ticket is created to address requested changes from this approach instead?  This set of changes is working and addresses the original requirements. If there is a desire to change direction, I believe it would be more efficient to address in a new issue vs rework what’s here. Also, redoing this work will take away from the requested component efforts. 

Be happy to hop on a call to discuss.

igarashitm:
@mattrpav are you saying that you guys can even address the UXD and UI code as well for the format thing?

mattrpav:
@igarashitm. No, I’m suggesting that supporting this in the runtime should have no impact to the UI. If the decision is to not support dates and times in Json and Xml for GA, no further dev should be needed. 

If the desire is to support dates and times for xml and Json for GA, than the UI would need to be updated anyway.

igarashitm:
That's exactly why I want to have a separate thread, as it doesn't complete by itself

rohanmars:
@igarashitm question: do you agree with the notion of a format attribute on an input/output field or not?

igarashitm:
I'm not sure at this moment, and I do want to avoid introducing configuration into TypeConverter as far as we can. Mainly it introduces complexity in the UI. For me the date/time format thing can be addressed with the source side field action. Adding it in UI doesn't look so hard. But I don't yet know which is better eventually.

igarashitm:
Just to make sure, I'm not saying this PR introduces complexity in the UI, because it hasn't addressed UI thing. But this definitely requires UI work which introduces complexity.

rohanmars:
I'm not convinced that the field action approach will be sufficient. Specifically as the readers directly call the convertType method, thus supporting unmarshaling structured string values into specific types. I view the format attribute as a marshalling/unmarshalling hint. Interestingly enough, the writers should probably be extended to use the conversion service convertType method similar to the readers.

igarashitm:
Do you have an example? Assuming Java-Java mapping, Do you mean having date/time format String field with `FieldType.DATE_TIME_TZ`? IMO that shouldn't happen. It should be `FieldType.STRING`, then source side field action would be `sourceType=FieldType.STRING` and `targetType=FieldType.DATE_TIME_TZ`. Then mapping to target field type `FieldType.DATE_TIME_TZ` doesn't require any converter. Java inspection never adds `FieldType.DATE_TIME_TZ` on a String field.

igarashitm:
As per recent discussion we're going to merge this - @rohanmars can you rebase on top of upstream master?

rohanmars:
Fixes #6 
Fixes #51

Issue #225, opened 2018-01-08
No field actions for Boolean, Short, Date 


mmelko:
Currently there is not possible to add transformation to Boolean, Short, Date fields. I'm not sure if it's intentional but I believe (at least) short should have similar transformation possibilities as integer.

igarashitm:
Hi @mmelko , are you seeing #107 ? This is a UI bug for sure, as we already have some transformations (aka field actions) which should show up for the `NUMBER` field types. And those `NUMBER` field actions should show up for the `Short` field as well.

@jpav do we have Boolean field actions?

mmelko:
yeah looks same, however in #107 it's about Number>Number transformations in general isn't ? I can see transformations for all number types except short.  Is it possible that Short isn't a NUMBER type for atlasmap ?

igarashitm:
It sounds like a variation of #107 bug. Short should be treated as a `NUMBER` for sure.

@rohanmars , can you take a look at this to see if it's easier for you to address this as a part of #107 ?

pleacu:
Hey - since this is my first crack at AtlasMap the more eyes on this the better!

The inital issue was that the 'Add Transformation' button wasn't showing up for some field types.
The MappingFieldActionComponent class has a method to return an array of applicable field actions
for the selected field pair types.  It walks all of the field actions as defined in the TransitionModel
and allows it to determine if we should add the field action to the array.  A non-zero array triggers
the 'Add Transformation' button.

The issue is with the FieldActionConfig class/ appliesToField method in the transition model.
It's basically only set up to work with 'STRING' and 'NUMBER' field actions.  It does go through
some effort to relate generic field action method 'STRING' and 'NUMBER' types to their
sources/targets field types but no other types.

What I did:

* Removed the "field" argument to the appliesToField() method - the field mapping pair was already
  being supplied.

* Made sure the field pair was already fully mapped.

* Checked for string and number types.

* Allowed non-primitive types if the selected field pair types matched the field action types.

Please see the test PR for details.

thkx!

mmelko:
Reopening because there are still no transformations when Boolean is involved (Boolean to String) into mapping. 
Also no transformation is possible when conversion (String<--> numbers) is necessary. I believe this is regression, it was possible to add transformation before.

pleacu:
Hey @mmelko , I relaxed the code to check target types when considering field actions.  There are no actions for target boolean but your  (Boolean to String)  should work now (and others).  Please reopen if it fails.

[...]

PR #240, opened 2018-01-18
i225 - test PR - fix add transformation button for some types

pleacu:
fix: Missing 'Add Transformation' button for certain field types.
Fixes: #225, #107

Signed-off-by: Paul Leacu <pleacu@redhat.com>
Test PR until reviewed

Hey - since this is my first crack at AtlasMap the more eyes on this the better!

The inital issue was that the 'Add Transformation' button wasn't showing up for some field types.
The MappingFieldActionComponent class has a method to return an array of applicable field actions
for the selected field pair types.  It walks all of the field actions as defined in the TransitionModel
and allows it to determine if we should add the field action to the array.  A non-zero array triggers
the 'Add Transformation' button.

The issue is with the FieldActionConfig class/ appliesToField method in the transition model.
It's basically only set up to work with 'STRING' and 'NUMBER' field actions.  It does go through
some effort to relate generic field action method 'STRING' and 'NUMBER' types to their
sources/targets field types but no other types.

What I did:

* Removed the "field" argument to the appliesToField() method - the field mapping pair was already
  being supplied.

* Made sure the field pair was already fully mapped.

* Checked for string and number types.

* Allowed non-primitive types if the selected field pair types matched the field action types.

Please see the test PR for details.

thkx!

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.action.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping.field.action.component.ts
@@ -77,8 +77,12 @@ export class MappingFieldActionComponent {
 
     public getActionConfigs(): FieldActionConfig[] {
         const configs: FieldActionConfig[] = [];
+
+        // Start with the complete list of field actions.
         for (const config of TransitionModel.actionConfigs) {
-            if (config.appliesToField(this.mappedField.field, this.fieldPair)) {
+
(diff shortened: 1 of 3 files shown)


igarashitm:
Hi @pleacu , thanks for the quick fix! can you put the `Fixes: #225, #107` line into git commit log? In that way we can reach the linked issue even from commit history. Other than that LGTM, could anyone from @atlasmap/ui have a quick look?

Issue #241, opened 2018-01-19
Support multiple documents in the Message map held as an Exchange property


igarashitm:
See POC in Syndesis: syndesisio/syndesis#1115 / syndesisio/syndesis#1143

syndesisio/syndesis#1110 is introducing multiple document support for the datamapper, and is supposed to carry a message map in a Exchange property. camel-atlasmap should extract that message map from Exchange property and assign available documents to/from AtlasSession appropriately.

PR #246, opened 2018-01-23
feat: Support multiple documents in the Message map held as an Exchan…

igarashitm:
…ge property

Fixes: #241


igarashitm:
@atlasmap/runtime could anyone review this one? thanks!

igarashitm:
Yep, I think the existing test cases are basically assuming single source/target and covering basic cases, but there might be something missing...

Issue #229, opened 2018-01-10
Default value of input fields is set to '0' in all Transformation attributes  


mmelko:
<img width="400" alt="screen shot 2018-01-10 at 15 37 20" src="https://user-images.githubusercontent.com/6814482/34777992-46daf4a2-f61c-11e7-84fa-edf20b833caf.png">

igarashitm:
Empty is far better than `0`, not sure if there's any field action which want to distinguish `not-specified` from `empty`. @jpav any idea?

jpav:
The only distinction I'd expect to see is one between null and empty. The 0 doesn't make sense to me either.

igarashitm:
@jpav distinguishing `not-specified` from `empty` might require yet another trick in UI, to capture user's intension that they put an `empty` parameter. @atlasmap/ui does anyone know if there's a standard way to do that in Web UI?

jpav:
I agree. Maybe we could display a grayed-out value in angle brackets containing "<null>" or "<empty>" or something? But maybe that would look weird, not sure.

igarashitm:
Also, @jpav do you have specific field actions which require distinguishing null and empty? This sounds like a gray-zone if we want to do before GA vs. defer until we hit fair amount of use cases.

igarashitm:
Let's separate the discussion, I would +1 to focus on, by default, leaving the field empty in UI here, then keep discussing for the empty vs. null in #243.

jpav:
Well, it's not so much about the particular field actions I don't think as much as the expected output. Null values vs. empty values can definitely be treated differently by whatever applications end up consuming XML documents. I don't know specific use-cases, but I've seen this distinction many times in the past in various documents, and the distinction is addressed both in the XSD spec as well as the XSD spec primer I believe. Having said that, it's probably okay for now to defer until we have more customer feedback/use-cases.

igarashitm:
FYI
https://github.com/atlasmap/atlasmap/pull/247/files/a3bcf4edd329f6fbfd7b3d3ca75dbe6d352a0ff3#diff-65f196fbf9fa14c7643a45c3b3141e4fL126

pleacu:
I changes the input field default to the empty string.  I also added a one-line correction to reset the field action argument list on transformation selection.

PR #247, opened 2018-01-24
fix: Do not lose user specified argument values between sessions.

pleacu:
Fixes: #110


ui/src/app/lib/atlasmap-data-mapper/models/transition.model.ts
```diff
@@ -121,9 +121,13 @@ export class FieldActionConfig {
     public populateFieldAction(action: FieldAction): void {
         action.name = this.name;
         action.config = this;
-        action.argumentValues = [];
-        for (const arg of this.arguments) {
-            action.setArgumentValue(arg.name, '0');
```
igarashitm: Sorry to hijack, but just a FYI for #229 taker - this line might be a root cause of #229

Issue #242, opened 2018-01-19
Update atlasmap config object to accept step IDs, labels, etc


gashcrumb:
On the UI side of things we've this issue -> https://github.com/syndesisio/syndesis/issues/1137 (data structure not finalized), this issue is for updating the datamapper API to accept this data structure in some form.

Right now the syndesis UI goes and makes function calls to add input and output documents, we could either update those functions or consider migrating to one function that accepts an object containing the input and output documents to maybe simplify the interface between components.

igarashitm:
@gashcrumb how about this? Having a single model object (just Documents?) holding a list of source & target docs.

```json
{
  "sourceDocuments": [
    {
      "name": "Salesforce (Step 1)",
      "id": "<generated>",
      "kind": "json-schema",
      "specification": "...", // JSON Schema payload for the Salesforce object
    },
    ...
  ],
  "targetDocuments": [
    {
       "name": "Twitter (Step 2)",
       "id": "<generated>",
       "kind": "java",
       "specification": "twitter4j.Status",
       "inspectionOutcome": "...", // inspection result content which is computed ahead via maven plugin. In Syndesis we are holding it in the DataShape.specification for now.
    },
    ...
  ]
}
```
Syndesis right now doesn't need multiple target documents, but AtlasMap supports it, so let's add it. Syndesis can just pass a single target document.

A small difference in the term, I want to avoid putting inspection results into specification like in Syndesis today. Instead added a `inspectionOutcome` to hold pre-computed inspection result. Then specification is defined as a "Source of inspection" in AtlasMap. The property name might be even "inspectionSource".

igarashitm:
Noticed that the `ConfigModel` already holds lists of sourceDocs/targetDocs. Might be better to refactor a bit to map between Syndesis DataShape smoothly.

igarashitm:
weird, auto close didn't work. closing

PR #249, opened 2018-01-24
fix: Refactor Document initialization model (#242)

igarashitm:
Introduced a `DocumentInitializationModel` which is a subset of `DocumentDefinition`
```TypeScript
export class DocumentInitializationModel {
    public id: string; // Document ID. Syndesis would put Step UUID.
    public type: DocumentType;
    public shortName: string; // Document label. Syndesis would put "Salesforce Contact (Step 1)" here
    public fullName: string; // Shown as a tooltip. Put a longer description for the document
    public isSource: boolean;
    public inspectionType: InspectionType;
    public inspectionSource: string;
    public inspectionResult: string;
}
```
and let `ConfigModel` accept it via
```TypeScript
addDocument(docInitModel: DocumentInitializationModel): DocumentDefinition
```
and
```TypeScript
addDocuments(docModels: DocumentInitializationModel[]): DocumentDefinition[]
``` 

Also
fix: Show correct document type (#88)

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
@@ -14,7 +14,7 @@
     limitations under the License.
 */
 
-import { Component, Input, ViewChildren, ElementRef, QueryList, ViewChild } from '@angular/core';
+import { Component, Input, ViewChildren, ElementRef, QueryList, ViewChild, OnInit } from '@angular/core';
 
 import { DocumentType, ConfigModel } from '../models/config.model';
 import { Field } from '../models/field.model';
@@ -37,7 +37,7 @@ import { ModalWindowComponent } from './modal.window.component';
                     <h2 class="card-pf-title">
                         <div class="docName">
                             <i class="fa {{ isSource ? 'fa-hdd-o' : 'fa-download' }}"></i>
-                            <label>{{ getSourcesTargetsLabel() }}</label>
+                            <label>{{ sourcesTargetsLabel }}</label>
                         </div>
                         <i (click)="toggleSearch()" [attr.class]="getSearchIconCSSClass()"></i>
                         <div class="clear"></div>
@@ -53,7 +53,7 @@ import { ModalWindowComponent } from './modal.window.component';
                 </div>
                 <div [attr.class]="searchMode ? 'fieldListSearchOpen' : 'fieldList'" style="overflow:auto;"
                     (scroll)="handleScroll($event)">
-                    <div *ngFor="let docDef of cfg.getDocs(isSource)" #docDetail class="docIdentifier" [attr.id]='docDef.shortName'>
+                    <div *ngFor="let docDef of documents" #docDetail class="docIdentifier" [attr.id]='docDef.shortName'>
                         <div class="card-pf-title documentHeader" tooltip="{{ docDef.fullName }}" placement="bottom"
                             *ngIf="isDocNameVisible(docDef)" (click)="toggleFieldVisibility(docDef)">
                             <div style="float:left">
@@ -71,7 +71,7 @@ import { ModalWindowComponent } from './modal.window.component';
                             <document-field-detail #fieldDetail *ngFor="let f of docDef.fields" [modalWindow]="modalWindow"
                                 [field]="f" [cfg]="cfg" [lineMachine]="lineMachine"></document-field-detail>
                             <div class="FieldDetail"
-                                *ngIf="!searchMode && docDef.isPropertyOrConstant() && !docDef.fields.length">
+                                *ngIf="!searchMode && docDef.isPropertyOrConstant && (!docDef.fields || !docDef.fields.length)">
                                 <label style="width:100%; padding:5px 16px; margin:0">
                                     No {{ docDef.type == 'Property' ? 'properties' : 'constants' }} exist.
                                 </label>
@@ -88,7 +88,7 @@ import { ModalWindowComponent } from './modal.window.component';
     `,
 })
 
-export class DocumentDefinitionComponent {
+export class DocumentDefinitionComponent implements OnInit {
     @Input() cfg: ConfigModel;
     @Input() isSource = false;
     @Input() lineMachine: LineMachineComponent;
@@ -102,6 +102,18 @@ export class DocumentDefinitionComponent {
     private searchFilter = '';
     private scrollTop = 0;
     private searchResultsExist = false;
+
+    private sourcesTargetsLabel: String;
+    private documents: DocumentDefinition[];
+
+    ngOnInit(): void {
+        if (this.isSource) {
+            this.sourcesTargetsLabel = (this.cfg.sourceDocs.length > 1) ? 'Sources' : 'Source';
+        } else {
+            this.sourcesTargetsLabel = (this.cfg.targetDocs.length > 1) ? 'Targets' : 'Target';
+        }
+        this.documents = this.cfg.getDocs(this.isSource);
+    }
 
     getDocDefElementPosition(docDef: DocumentDefinition): any {
         for (const c of this.docElements.toArray()) {
@@ -165,14 +177,6 @@ export class DocumentDefinitionComponent {
         return this.searchMode ? (cssClass + ' selectedIcon') : cssClass;
     }
 
-    getSourcesTargetsLabel(): string {
-        if (this.isSource) {
-            return (this.cfg.sourceDocs.length > 1) ? 'Sources' : 'Source';
-        } else {
-            return (this.cfg.targetDocs.length > 1) ? 'Targets' : 'Target';
-        }
-    }
-
     getFieldCount(): number {
         let count = 0;
         for (const docDef of this.cfg.getDocs(this.isSource)) {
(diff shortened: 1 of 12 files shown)


igarashitm:
@gashcrumb I expect data-mapper-host.component.ts would repack the `DocumentInitializationModel` from a list of steps which contains corresponding DataShape (syndesisio/syndesis#1137). If you could take a look that would be cool!

igarashitm:
ugh it's breaking something. debugging now. But outline is not gonna change.

igarashitm:
OK, now it's good to go.

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
```diff
@@ -53,8 +53,8 @@ import { ModalWindowComponent } from './modal.window.component';
                 </div>
                 <div [attr.class]="searchMode ? 'fieldListSearchOpen' : 'fieldList'" style="overflow:auto;"
                     (scroll)="handleScroll($event)">
-                    <div *ngFor="let docDef of cfg.getDocs(isSource)" #docDetail class="docIdentifier" [attr.id]='docDef.name'>
-                        <div class="card-pf-title documentHeader" tooltip="{{ docDef.fullyQualifiedName }}" placement="bottom"
+                    <div *ngFor="let docDef of cfg.getDocs(isSource)" #docDetail class="docIdentifier" [attr.id]='docDef.shortName'>
```
deeleman: The Angular change detector will evaluate this line multiple times in a second, thus executing `cfg.getDocs(isSource)` on each pass. I do not know the internals of that function but I'd move it to a `docDefs` property in the component class and populate it at `ngOnInit()`. Otherwise you'll wind up running this method every few milliseconds.

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
```diff
@@ -71,9 +71,9 @@ import { ModalWindowComponent } from './modal.window.component';
                             <document-field-detail #fieldDetail *ngFor="let f of docDef.fields" [modalWindow]="modalWindow"
                                 [field]="f" [cfg]="cfg" [lineMachine]="lineMachine"></document-field-detail>
                             <div class="FieldDetail"
-                                *ngIf="!searchMode && docDef.initCfg.type.isPropertyOrConstant() && !docDef.fields.length">
+                                *ngIf="!searchMode && docDef.isPropertyOrConstant() && !docDef.fields.length">
```
deeleman: Same as my previous comment, you do not want to attach functions execution to an structural directive (such as `NgIf` here) since those will be executed on each each pass of the change detector.

Fix: move `docDef.isPropertyOrConstant()` execution to an assignation expression inside `OnInit` in the component controller that binds its value to a `isPropertyOrConstant: boolean` class field and then bind that field inside the `*ngIf` conditional expression.

Regarding `!docDef.fields.length`, that will raise an exception if `docDef.fields` does not exist. I do not know the  internals of its implementation but bear that in mind. Perhaps a more type-safety way to check the whole conditional would be (following up with my previous comment above):

`*ngIf="!searchMode && docDef.isPropertyOrConstant && (!docDef.fields || !docDef.fields.length)"`

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
```diff
@@ -197,8 +197,8 @@ export class DocumentDefinitionComponent {
         const self: DocumentDefinitionComponent = this;
         this.modalWindow.reset();
         this.modalWindow.confirmButtonText = 'Save';
-        const isProperty: boolean = docDef.initCfg.type.isProperty();
-        const isConstant: boolean = docDef.initCfg.type.isConstant();
+        const isProperty: boolean = docDef.type == DocumentType.PROPERTY;
+        const isConstant: boolean = docDef.type == DocumentType.CONSTANT;
```
deeleman: This is just a quick note regarding styling, but nothing really mandatory at all:

TypeScript is smart enough to infer types out of the assigned expression of a class field. In such case as the above, you can safely remove the `boolean` type annotation from the const expression. 

Feel free to remove it here or leave it, and bear in mind this hint for the future. ;)

ui/src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts
```diff
@@ -167,14 +129,24 @@ export class DocumentDefinition {
     }
 
     public getName(includeType: boolean): string {
-        let name: string = this.name;
-        if (ConfigModel.getConfig().showTypes && !this.initCfg.type.isPropertyOrConstant()) {
-            const type: string = this.initCfg.type.isJava() ? ' (Java)' : ' (XML)';
-            name += type;
+        let name: string = this.shortName;
+        if (ConfigModel.getConfig().showTypes && !this.isPropertyOrConstant()) {
+            const type: string = this.type;
```
deeleman: Here, both `name` and `type` are assgined right after being declared. Then you do not need any type annotation at all.

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
igarashitm: Thanks for the review!
There are 2 types of docs here, and we are retrieving sourceDocs if isSource=true, otherwise targetDocs. Do we want to add 2 `*ngIf` here to add this `<div>` conditionally?

deeleman: If `isSource === false`, that <div> will not even show up so the conditional would become redundant.

igarashitm: I think cfg.getDocs(false) returns target documents, am I missing something?
https://github.com/atlasmap/atlasmap/blob/master/ui/src/app/lib/atlasmap-data-mapper/models/config.model.ts#L174-L181
btw these method stink...

igarashitm: Ah never mind, `isSource` is never changed once initialized, so `ngOnInit()` can decide which is required. Same for `Sources`/`Targets` label.

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
igarashitm: Oh good to know. This is a bad habit Java has been enforcing on me :-)

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
igarashitm: This required to hide the `type` property behind the setter/getter, to capture `isPropertyOrConstant` when `type` is set.

igarashitm:
OK fixed, added a follow-up commit.

ui/src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts
```diff
@@ -56,65 +56,28 @@ export class NamespaceModel {
...
+    public id: string;
+    public _type: DocumentType;
+    public shortName: string;
+    public fullName: string;
+    public uri: string;
+    public inspectionType: InspectionType;
+    public inspectionSource: string;
+    public inspectionResult: string;
+    public isSource: boolean;
+    public isPropertyOrConstant: boolean;
```
deeleman: Just a note for the future: Class members are `public` by default, so the standard convention is NOT to annotate public fields with the `public` accessor. This is just a remark, so please leave this file as it is since adding the `public` accessor is not wrong either (although redundant).

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
```diff
@@ -103,9 +103,21 @@ export class DocumentDefinitionComponent {
     private scrollTop = 0;
     private searchResultsExist = false;
 
+    private sourcesTargetsLabel: String;
```
deeleman: `String` is not a valid TypeScript type, but a reference to JavaScript's underlying class type for... strings. That is why the compiler can honor this annotation although not correct. You should use `string` (all in lowercase) instead.

ui/src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts
igarashitm: Thanks, let's cleanup before I forget ;-)

ui/src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
igarashitm: Ugh good catch!! my hand is cursed by Java...

igarashitm:
OK fixed, thanks for the review again!
BTW, I wonder if we can enforce that not to have verbose type declare at left side and public modifier. Does lint support it?

igarashitm:
Found `member-access` with `no-public`
https://palantir.github.io/tslint/rules/member-access/

igarashitm:
hmm... I would expect `no-inferrable-type` does that, but apparently not...
https://palantir.github.io/tslint/rules/no-inferrable-types/

PR #252, opened 2018-01-26
fix: make java class inspection put inspected class in the map of cac…

### LINKED ISSUE #81: Search and Autocomplete are broken
|</img> @rhuss | [2017-09-21](https://github.com/atlasmap/atlasmap-ui/issues/83) | UX Design  |
|-|-|-|

When using a Twitter.Status object in the atlasmap-ui as input and I search for "user" I get the following result:

image

However, there is a top-level "user" element in the tree which I expect to be included, too:

image

This happened also for "screen" input where I expected "user.screenName" in the autocompletion.

image

After clearing the field and trying to search again, "user.screenName" appeared:

image

rohanmars:
Fixes #81

### DIFF
--- a/runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
+++ b/runtime/modules/java/core/src/main/java/io/atlasmap/java/inspect/ClassInspectionService.java
@@ -184,7 +184,9 @@ public class ClassInspectionService {
         }
 
         JavaClass javaClass = AtlasJavaModelFactory.createJavaClass();
-        inspectClass(classLoader, clazz, javaClass, new HashSet<String>(), null);
+        Set<String> cachedClasses = new HashSet<String>();
+        cachedClasses.add(clazz.getCanonicalName()); // we cache ourself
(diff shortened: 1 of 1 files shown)


rohanmars:
This fix will cause objects to be primed at their first definition/use and most importantly excludes the recursive usecase which fixes the specific issue #81.

We should probably create a new issue for a deeper priming of fields. For example, when you have two address fields, shipping and billing. In that case if you search for addressLine1, it will only come up first in shipping (if that is listed first). It gets complicated as very complex class hierarchies could have performance issues on the UI, but that could be mitigated by having a cap on automatically primed cache entries.

igarashitm:
Hi @rohanmars ,
Can you put the `Fixes: #81` line as a 2nd line of git commit log so we can have a issue link from GitHub history?

igarashitm:
I'll try running with UI later

igarashitm:
verified searching by `screen` shows `/user/screenName`, and `/retweetedStatus/user/screenName` disappeared, which is even better for UX.

Then we finally reached to the "right" issue here - #255

@rohanmars thank you for the progress on this, it was really a great catch!!

PR #262, opened 2018-01-30
feature: Support embedding rest services in other JAXRS applications

chirino:
Did a bit of cleanup and made the JAXRS services defined by Atlasmap a little more self contained so that they can more easily be embedded in other JAXRS applications like Syndesis rest.


mattrpav:
Good stuff!

Issue #266, opened 2018-02-01
Support multiple namespaces on a XML Document


igarashitm:
XML SchemaInspector collects target namespaces from each XSSchema
https://github.com/atlasmap/atlasmap/blob/ecc3770b0dcfac2215d35a58924e5f21f2bbcc98/runtime/modules/xml/core/src/main/java/io/atlasmap/xml/inspect/SchemaInspector.java#L144-L150

Then those namespaces are stored in DataSource definition in the mapping file, which is eventually consumed in XmlFieldWriter
https://github.com/atlasmap/atlasmap/blob/ecc3770b0dcfac2215d35a58924e5f21f2bbcc98/runtime/modules/xml/core/src/main/java/io/atlasmap/xml/core/XmlFieldWriter.java#L128

So missing pieces are
* UI to pass multiple schema from DataShape.specification in Syndesis through AtlasMap UI inspection request 
* XML inspection service to consume an array of schema files
* In XML SchemaInspector, feed multiple XSD files into parser and have multiple XSSchema there
* Test

igarashitm:
@zregvart had a great idea to eliminate UI work - we can define a umbrella schema for the multiple ones like this and define as a XML inspection request payload. It might require some try&error though, worth a shot:
```xml
<atlas:SchemaSet xmlns:atlas="http://atlasmap.io/java/v2/schemaset" 
  xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:schema>
    <xs:import namespace="ns1"/>
    <xs:import namespace="ns2"/>
    ...
  </xs:schema>
  <atlas:AdditionalSchemas>
    <xs:schema targetNamespace="ns1">
      ...
    </xs:schema>
    <xs:schema targetNamespace="ns2">
      ...
    </xs:schema>
  </atlas:AdditionalSchemas>
</atlas:SchemaSet>
```

PR #272, opened 2018-02-08
feat: Support multiple namespaces on a XML Document

igarashitm:
Fixes: #266


igarashitm:
Hi @zregvart , here is a `SchemaSet` inspection input used for unit test.
https://github.com/atlasmap/atlasmap/pull/272/files#diff-a573f917c8175dba539006c9b5b74ba2

None of `xs:import` or `xs:include` is used. The inspector just retrieves all the `xs:schema` element from the file, and take the `/atlas:SchemaSet/xs:schema` as a root schema. The namespace prefixes are added if it's different from the target namespace of the root one.

zregvart:
@igarashitm awesome!

Issue #269, opened 2018-02-02
WARN  i.a.c.DefaultAtlasConversionService - Converter ... exists.


igarashitm:
_From @chirino on February 2, 2018 14:21_

Lots of Atlasmap warning on start up:

```
14:19:17.341 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.String aleady exists.
14:19:17.341 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Boolean aleady exists.
14:19:17.342 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Long aleady exists.
14:19:17.342 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Byte aleady exists.
14:19:17.342 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Short aleady exists.
14:19:17.342 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Character aleady exists.
14:19:17.342 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Double aleady exists.
14:19:17.343 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Float aleady exists.
14:19:17.343 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Integer aleady exists.
14:19:17.344 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.String aleady exists.
14:19:17.344 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Boolean aleady exists.
14:19:17.344 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Long aleady exists.
14:19:17.344 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Byte aleady exists.
14:19:17.345 [main] WARN  i.a.c.DefaultAtlasConversionService - Converter between java.lang.Object and java.lang.Short aleady exists.
[...]

PR #280, opened 2018-02-09
fix: filter out redudant converter methods

rohanmars:
Fixes: #269

### DIFF
--- a/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasConversionService.java
+++ b/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasConversionService.java
@@ -174,7 +174,10 @@ public class DefaultAtlasConversionService implements AtlasConversionService {
             while (klass != Object.class) {
                 final List<Method> allMethods = new ArrayList<Method>(Arrays.asList(klass.getDeclaredMethods()));
                 for (final Method method : allMethods) {
-                    if (method.isAnnotationPresent(AtlasConversionInfo.class) && method.getParameters().length > 0) {
+                    // we filter out methods which aren't annotated @AtlasconversionInfo and have to
+                    // also filter out methods which are synthetic methods to avoid duplicates
+                    if (method.isAnnotationPresent(AtlasConversionInfo.class) && method.getParameters().length > 0
+                            && !method.isSynthetic()) {
                         String sourceClassName = method.getParameters()[0].getType().getCanonicalName();
                         ConverterKey coordinate = new ConverterKey(sourceClassName,
                                 method.getReturnType().getCanonicalName());


igarashitm:
And we do need to accept Object as a parameter. Like I said Object is not `FieldType.COMPLEX` but rather it's anything including boxed primitives.

rohanmars:
I debugged that code and found that the getDelaringMethods() was returning the same method (converter method) twice. One with the real typed class for the paramaters, say Boolean, and another with the parameter as Object but the same method name. I believe as a consequence of the primitive type converters implementing a method with a generic typed parameter. Thus it was trying to add all the object to class methods again and again and getting key collision. I couldn't find a method which did not return duplicate methods so I thought a simple filter might be more prudent.

I think the internals of the conversion service shouldn't accept java.lang.Object to some typed class as there is not enough information passed down to the converter method and even if we use the fieldType it will break for FieldType.COMPLEX scenarios.  Ideally the actual fields would be passed down to the converters as they might give some clues about the actual classes, especially in the case of a java mappings. However, I ruled that out as it required significantly more changes when I investigated that approach previously.

igarashitm:
`FieldType.COMPLEX` is not a mappable field, but it's a container of the fields.

rohanmars:
FieldType.COMPLEX was/is complex in my head. haha. I had asked @mattrpav previously what his intention with COMPLEX was which was what I based the broader non-primitive support on. I think we should do a call on Monday to discuss this in more detail.

igarashitm:
Even in the code there's still a confusion, I think this method should return `FieldType.ALL` (or introduce `FieldType.ANY`? not sure for what purpose ALL was introduced at first place) or something for the `java.lang.Object`. It might be even in `default`
https://github.com/atlasmap/atlasmap/blob/master/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasConversionService.java#L420

`FieldType.COMPLEX` means it's expandable in the UI. Therefore it never appears as a mappable field. We might even need to throw an error if `FieldType.COMPLEX` is requested to be mapped.

rohanmars:
As I understand the original intention of COMPLEX was a field which contains child fields and/or is a non-primitive type. 

@igarashitm why would you want an java.lang.Object to passed as a converter parameter?

igarashitm:
Just because that restriction is unnecessary

mattrpav:
Hi @igarashitm it sounds like there are a couple of convos going on here, all related to AutoConversion. Happy to have the convo (see below). At the same time, I'd put forth that this issue is addressed per requirement, and if there are further change or discussion requests to AutoConversion, we roll those into separate GitHub issues

// Begin general discussion on AutoConversion

Remember-- each mapping definition is expecting *known* types ahead of time. The intent of AtlasMap is to be very type-specific and type-aware. This is complimentary (but different) from Camel, where you could have a process flow auto-detect and auto-convert based on multiple datatypes coming through. Another way to look at it is that AtlasMap should be *deterministic*.

The main goal of AutoConversion is to prevent having to have really verbose mapping definition files. Imagine having to do:

> <InputField type="Integer"... path="myInt" ../>
> <convert method="integerToString" ../>
> <OutputField type="String" .. path="otherString" />

For every field mapping! Yikes. AutoConversion eliminates a lot of redundant declarations in the mapping definition. 

However, AutoConversion is *not* intended to support things like reading from a JMS Queue and auto-converting different source types to the same target type based on the payload. I'd put forth that in that use case, a Camel route should do CBR and pass the messages to separate AtlasMapping steps based on the supported content types.

1) FieldType.COMPLEX is intended to define a non-primitive type, or a collection of fields. Its primary use is inspection and design time. Other than being a marker for 'unsupported' it shouldn't be that useful at runtime.

2) To define a conversion of non-primitive Java classes, like 'com.company.Order to com.company.Invoice', the information needed to pull in the right AutoConversion *is* the classname. If FieldType.COMPLEX is used, the unique identifier is a composite key of "FieldType.COMPLEX + com.company.Order" to "FieldType.COMPLEX + com.company.Invoice", which can just be reduced to "com.company.Order to  com.company.Invoice", since that provides enough uniqueness to identify a proper conversion method. Therefore, declaring or annotating support for FieldType.COMPLEX really doesn't provide any meaning or value. 

3) I think we should resist having any java.lang.Object defined converters as long as possible (or at least until there is a very specific customer use case defined), b/c it breaks the "strongly-typed-we-know-exactly-what-we-are-converting" deterministic paradigm.

4) As far as the idea of a  FieldType.ANY or FieldType.ALL, I think I'd want to understand some specific use cases. In general, I think if we mix and match strongly-typed with loosely-typed the framework becomes difficult to define and extend.

5) IMHO AtlasMap should be opinionated in that if we get something unexpected, or unsupported that we fail early and provide a clear message-- "No converter found for com.company.Order to com.company.Invoice".  This is clean.  If we have an Order as a field in a mapping and want to convert it to an Invoice as field in a target type, the user should clearly declare that intent at design time, and reflected in the AtlasMapping definition file. If something isn't defined or a converter is not wired in correctly, we should error out.

I think there are a number of concepts you bring up here, that I'd be happy to talk through, but it is probably to big a convo for just a GitHub discussion. I'd be happy to have a call later today or tomorrow to talk through.

TLDR; Camel for loosely typed non-deterministic conversions, AtlasMap for strongly typed deterministic conversions.

igarashitm:
I would like to defer having discussion about java.lang.Object thing as we can't guarantee there's no scenario we want it (so called devil's proof) at least that's not something to talk in the context of bug fix.

igarashitm:
If you think this is the only way you can find, I would keep the issue open and wait for myself to find some spare time to look into. It's far better than accept that huge restriction right now and hide the issue without any guarantee that it won't cause any problem.

rohanmars:
@igarashitm the code as is will not support object parameters properly anyway. Due to key collision. Which goes to this very issue/bug of warning message appearing saying that the converter exists already. At the point there is a requirement to support Object as convertible entity then the converter service will need some significant rework anyway I think.

mattrpav:
@igarashitm I'm not following you, what is the huge restriction and what use case do you see it blocking?

igarashitm:
I think I got @rohanmars was saying now, so method name check can't detect if there are 
* method:intToString, Param:Integer, Return:String
* method:toString, Param:Object, Return:String

I think this should be avoided by the converter developer though, at some point we can make even better check, looking at the parameter/return inheritance and put a WARN that the method is hidden by the other.

igarashitm:
ah wait, not putting WARN, use the Integer->String method for the Integer parameter and fallback to Object->String method for other parameter instance.

mattrpav:
Notes from call:

1. @igarashitm agreed to merge the change after Rohan applies a minor adjustment the method inspection to look for duplicate method names. Psuedo-code:

(duplicate name AND (return type of java.lang.Object OR parameter type of java.lang.Object))

2. Matt created #287 #288 for discussions about follow-on concepts brought up by @igarashitm

igarashitm:
Oh no, I didn't know Java method resolver is smart enough to do this...(I thought compiler would reject this)
```java
import org.junit.Test;

public class MyTest {

    public String toStr(Integer integer) {
        return "int";
    }

    public String toStr(Number number) {
        return "number";
    }

    public String toStr(Object object) {
        return "obj";
    }

    @Test
    public void test() {
        System.out.println(toStr(1));
        System.out.println(toStr("2"));
        System.out.println(toStr(3.0));
    }
}
```
Output
```
int
obj
number
```

So still it's possible that it's a declared method in the source code...

igarashitm:
@rohanmars Does `isBridge()` flag help?
https://bugs.openjdk.java.net/browse/JDK-8060179
https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Method.html#isBridge--

igarashitm:
might also want to filter out if `isSynthetic()`

mattrpav:
@igarashitm agreed. Synthetic and Bridge should tell us what we need.

@rohanmars please test by adjusting the logic test to see if this gets us there w/o looking for java.lang.Object specifically

(method.isSynthetic || method.isBridge) .. skip adding to AutoConversion.

rohanmars:
@igarashitm @mattrpav both methods return true if for the duplicate methods. I'll switch to use isSynthetic() for filtering out these methods.

Issue #282, opened 2018-02-13
Support XML schemas without namespaces


zregvart:
With #266 we now have support for a schema set containing multiple target namespaces. How difficult would be to match the imported schema with the schema in the `additionalSchemas` using `id` instead of namespaces? Turns out specifying XML namespaces in Swagger documents is optional, and the specification could contain non-namespaced elements mixed with namespaced elements. 

Something like:
```xml
<atlas:SchemaSet xmlns:atlas="http://atlasmap.io/xml/v2/schemaset" 
  xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:schema>
    <xs:import namespace="ns1"/>
    <xs:import id="id2"/>
    ...
  </xs:schema>
  <atlas:AdditionalSchemas>
    <xs:schema targetNamespace="ns1">
      ...
    </xs:schema>
    <xs:schema id="id2">
      ...
    </xs:schema>
  </atlas:AdditionalSchemas>
</atlas:SchemaSet>
```

If it would be simpler/easier perhaps we could rely just on the `id`s, i.e. not support matching by `namespace` at all; something like:

```xml
<atlas:SchemaSet xmlns:atlas="http://atlasmap.io/xml/v2/schemaset" 
  xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:schema>
    <xs:import id="id1" namespace="ns1"/>
    <xs:import id="id2"/>
    ...
  </xs:schema>
  <atlas:AdditionalSchemas>
    <xs:schema id="id1" targetNamespace="ns1">
      ...
    </xs:schema>
    <xs:schema id="id2">
      ...
    </xs:schema>
  </atlas:AdditionalSchemas>
</atlas:SchemaSet>
```

zregvart:
See this https://gist.github.com/zregvart/4ab273a55641c06c654559512a6f4542 for example of this idea.

igarashitm:
Hi @zregvart - If I understand correctly your example is equal to having
```xml
...
<xs:schema id="id2" targetNamespace="">
...
```
I think I handled it wrong at first place, but not that hard to fix. I'll look into today.

`xs:import` (and `xs:include`) require EntityResolver to handle it, but currently we don't do that. This is required when we need to resolve the schema from external resource. As far as we can pass all the dependent `xs:schema` at once, apparently we don't need it.

zregvart:
[...]

PR #284, opened 2018-02-13
fix: Support XML schemas without namespaces

igarashitm:
Fixes: #282


igarashitm:
Fixed to get working with sub-schemas which doesn't have target namespace.

@zregvart FYI - I fixed `/SchemaSet/AdditionalSchemas` element to start with capital case `A`.

Issue #128, opened 2017-11-22
Fix circular references in AtlasMap UI


deeleman:
The following circular dependencies have been found. This might lead to scenarios where a particular dependency is imported as `undefined` from and `import { ... } from '...'` statement:

- [x] src/app/lib/atlasmap-data-mapper/common/data.mapper.util.ts -> src/app/lib/atlasmap-data-mapper/models/config.model.ts -> src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts -> src/app/lib/atlasmap-data-mapper/common/data.mapper.util.ts
- [ ] src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts -> src/app/lib/atlasmap-data-mapper/components/line.machine.component.ts -> src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts
- [ ] src/app/lib/atlasmap-data-mapper/components/line.machine.component.ts -> src/app/lib/atlasmap-data-mapper/components/document.definition.component.ts -> src/app/lib/atlasmap-data-mapper/components/line.machine.component.ts
- [x] src/app/lib/atlasmap-data-mapper/models/config.model.ts -> src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts -> src/app/lib/atlasmap-data-mapper/models/field.model.ts -> src/app/lib/atlasmap-data-mapper/models/config.model.ts
- [x] src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts -> src/app/lib/atlasmap-data-mapper/models/field.model.ts -> src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts
- [x] src/app/lib/atlasmap-data-mapper/models/field.model.ts -> src/app/lib/atlasmap-data-mapper/models/document.definition.model.ts -> src/app/lib/atlasmap-data-mapper/models/field.model.ts

pleacu:
@deeleman - Hey Pablo - Paul here (@pleacu) - I was looking at this trivial fcn:
 
https://github.com/atlasmap/atlasmap/blob/master/ui/src/app/lib/atlasmap-data-mapper/common/data.mapper.util.ts#L45

It's causing a circular dependency in several components - I was considering moving it to config.model.ts since it's generating a config.model specific error - wdyt???

deeleman:
@pleacu Hi Paul!

[...]

PR #306, opened 2018-02-26
fix: Remove document-definition.component dependence on line-machine …

pleacu:
…component using a custom event

Fixes: #128

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/components/data-mapper-app.component.html
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/data-mapper-app.component.html
@@ -14,13 +14,13 @@
     <div class="row" style="height:calc(100% - 32px); position:relative; padding-top:10px;" *ngIf="hasSourceDoc && hasTargetDoc">
       <div style='height:100%; padding:0;' [attr.class]="(getConfig().showMappingTable || getConfig().showNamespaceTable) ? 'col-md-9 hidden' : 'col-md-9'">
         <div style="float:left; width:40%; height:100%;">
-          <document-definition #docDefInput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="true" [lineMachine]="lineMachine"></document-definition>
+          <document-definition #docDefInput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="true"></document-definition>
         </div>
         <div style="float:left; width:20%; height:100%; margin-left:-5px; margin-right:-5px;">
           <line-machine #lineMachine [cfg]="getConfig()" [docDefInput]="docDefInput" [docDefOutput]="docDefOutput"></line-machine>
         </div>
         <div style="float:left; width:40%; height:100%;">
-          <document-definition #docDefOutput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="false" [lineMachine]="lineMachine"></document-definition>
+          <document-definition #docDefOutput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="false"></document-definition>
         </div>
         <div class="clear"></div>
       </div>
--- a/ui/src/app/lib/atlasmap-data-mapper/components/document-definition.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/document-definition.component.ts
@@ -26,7 +26,6 @@ import { PropertyFieldEditComponent } from './property-field-edit.component';
 import { ConstantFieldEditComponent } from './constant-field-edit.component';
 import { FieldEditComponent } from './field-edit.component';
 
-import { LineMachineComponent } from './line-machine.component';
 import { ModalWindowComponent } from './modal-window.component';
 
 @Component({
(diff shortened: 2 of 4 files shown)


ui/src/app/lib/atlasmap-data-mapper/components/document-definition.component.ts
```diff
@@ -132,9 +139,16 @@ export class DocumentDefinitionComponent implements OnInit {
...
+  /**
+   * Handle scrolling in this document definition instance.  Avoid a circular dependence with the
+   * LineMachineComponent by dispatching a custom mappings-line-redraw event.
+   * @param event
+   */
   handleScroll(event: any) {
     this.scrollTop = event.target.scrollTop;
-    this.lineMachine.redrawLinesForMappings();
+    this.redrawMappingLinesEvent = new CustomEvent('redrawMappingLines', {detail: this.lineMachine});
+    document.dispatchEvent(this.redrawMappingLinesEvent);
```
igarashitm: oh I just got to know this is a DOM document, built in var in javascript. I'm curious if  it's common way vs. using Angular `EventEmitter`.
https://github.com/syndesisio/syndesis/blob/248a2dbf077f324840b74830f69fe229752ceef1/app/ui/src/app/integration/edit-page/current-flow.service.ts#L352

ui/src/app/lib/atlasmap-data-mapper/components/data-mapper-app.component.html
```diff
@@ -14,13 +14,13 @@
...
         <div style="float:left; width:40%; height:100%;">
-          <document-definition #docDefInput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="true" [lineMachine]="lineMachine"></document-definition>
+          <document-definition #docDefInput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="true"></document-definition>
         </div>
         <div style="float:left; width:20%; height:100%; margin-left:-5px; margin-right:-5px;">
           <line-machine #lineMachine [cfg]="getConfig()" [docDefInput]="docDefInput" [docDefOutput]="docDefOutput"></line-machine>
         </div>
         <div style="float:left; width:40%; height:100%;">
-          <document-definition #docDefOutput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="false" [lineMachine]="lineMachine"></document-definition>
+          <document-definition #docDefOutput [cfg]="getConfig()" [modalWindow]="modalWindow" [isSource]="false"></document-definition>
```
deeleman: Id recommend to add a short custom prefix to all new atlasmapper custom elements, so we prevent naming collisions with any other 3rd party library we might implement in the future

ui/src/app/lib/atlasmap-data-mapper/components/document-definition.component.ts
```diff
@@ -61,6 +60,14 @@ export class DocumentDefinitionComponent implements OnInit {
     this.documents = this.cfg.getDocs(this.isSource);
   }
 
+  getLineMachine(): any {
+      return this.lineMachine;
+  }
+
+  setLineMachine(lm: any): void {
+      this.lineMachine = lm;
+  }
```
deeleman: I’d reformat this to two spaces indenting to stay consistent with the rest of the file

ui/src/app/lib/atlasmap-data-mapper/components/document-definition.component.ts
deeleman: Exactly, although it’s not commonly used outside the Web Workers arena unless one does not rely on any web components framework such as React, Vue or Angular, which is our case.

I rather stick to one single way to dispatch events in order not to mix up different paradigms in a single architecture and therefore I’d stick to Angulars built in output properties, just for consistency sake. :)

ui/src/app/lib/atlasmap-data-mapper/components/line-machine.component.ts
```diff
@@ -50,13 +50,21 @@ export class LineMachineComponent implements OnInit {
 
   @ViewChild('lineMachineElement') lineMachineElement: ElementRef;
 
-  private yOffset = 3;
+  redrawMappingLinesEventListener = document.addEventListener('redrawMappingLines', function(e: CustomEvent) {
+      const lmcInstance: LineMachineComponent = e.detail;
+      setTimeout(() => {
+          lmcInstance.redrawLinesForMappings();
+          }, 10);
+    }, false);
```
deeleman: We’re mixing native DOM API methods and globals (document, for instance), with callbacks as functions, assignment expressions with typing, parameters and constant names not matching its type, 4 spaces indenting, and defer execution using an arbitrary value (10).

This block might require some love and leverage Angular’s built in API for emitting events instead

ui/src/app/lib/atlasmap-data-mapper/components/document-definition.component.ts
igarashitm: And this is an example to emit the event, directly invoke `emit()` on a `CurrentFlowService.events`
https://github.com/syndesisio/syndesis/blob/248a2dbf077f324840b74830f69fe229752ceef1/app/ui/src/app/integration/edit-page/step-configure/data-mapper/data-mapper-host.component.ts#L134

PR #308, opened 2018-02-27
fix: Remove document-definition.component dependence on line-machine …

pleacu:
…component using a custom Angular event

Fixes: #128


pleacu:
Respin of https://github.com/atlasmap/atlasmap/pull/306 - used a different PR so the two can be contrasted .
This uses the Angular EventEmitter instead of the DOM custom event dispatch/handler mechanism.

Issue #312, opened 2018-02-27
Use java.util.Date as a DATE_TIME representative class


igarashitm:
Changed my mind, let's use `java.util.Date` as an official for now. See this comment
https://github.com/atlasmap/atlasmap/issues/312#issuecomment-369251121

And I had a wrong assumption that the `java.util.Date` recognize timezone, but it doesn't. So it should be `DATE_TIME`.

~~We still need to support `java.util.Date` as a kind of `DATE_TIME_TZ` though, let's use new&better java.time.* for the official representative like we're doing here~~
https://github.com/atlasmap/atlasmap/blob/9382fc9041c571378ab3770700f5f9f018c81d9e/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasConversionService.java#L507-L514
https://github.com/atlasmap/atlasmap/blob/9382fc9041c571378ab3770700f5f9f018c81d9e/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java#L224-L226

igarashitm:
Hi @jpav, could you take a look at this one?

igarashitm:
Ah wait. What happen if input object is `java.util.Date`. Should we make the parameter `java.lang.Object`? sounds ugly... identifying field action method only by the method parameter doesn't sound right here.

igarashitm:
As `java.time.*` are pretty new and we do see `java.util.Date` in existing usecase like `twitter4j.Status`, I'll change the default representative class of `FieldType.DATE_TIME_TZ` to `java.util.Date` for now. Let's wait until we really want to use `java.time.*` there.

PR #316, opened 2018-02-28
chore: Use java.util.Date as a DATE_TIME representative class

igarashitm:
Fixes: #312


igarashitm:
`java.util.Date` doesn't recognize timezone. need rework.

Issue #257, opened 2018-01-29
Better error message when Atlasmap Service is not running


rhuss:
When the atlasmap pod/service is not running because of some reason, the error message that happens when opening the datamapper is quite obscure:

![image](https://user-images.githubusercontent.com/99080/35480694-37a5b0d4-0414-11e8-9bbc-7846b06c7c05.png)

It would be much better if there was a dedicated error message that the mapping service was not reachable. Maybe this could be also communitcated back in an error dialog instead of an inline error ?

See also https://github.com/syndesisio/syndesis/issues/1273 where it has been discovered first.

igarashitm:
Ah given that we'll embed AtlasMap runtime into other pod in Syndesis, priority is a bit lower now I guess.

PR #327, opened 2018-03-06
fix: Better error message when Atlasmap Service is not running

igarashitm:
Fixes: #257


igarashitm:
Hi @pleacu, could you take a look to see if I don't brake anything you're working on?

igarashitm:
Ugh, thanks, fixed and added one more left over of network error handling.

Issue #268, opened 2018-02-02
Constant value must respect field type


igarashitm:
Constant value is populated on JAXB unmarshalling, i.e. it's always String. It must be converted into specified field type if exists.
https://github.com/atlasmap/atlasmap/blob/ecc3770b0dcfac2215d35a58924e5f21f2bbcc98/runtime/core/src/main/java/io/atlasmap/core/ConstantModule.java#L62-L72

PR #341, opened 2018-03-09
fix: Add support for runtime properties as a source

igarashitm:
Fixes: #76


igarashitm:
Ugh I forgot to add conversion as ENV/sysProp is always String. will update.

igarashitm:
I'll submit a new PR with #268 fix

PR #343, opened 2018-03-12
fix: Issue an error if the user attempts to create a property with th…

pleacu:
…e same name or a constant with the same value.

Fixes: #104 
@igarashitm - this is what the error diagnostic looks like when I attempt to add another property 'cat' and another constant '3.14159':

atlas1


igarashitm:
Looks good!
I wonder if it's possible to even disable the "OK" button in the dialog and show a warning if it's duplicate.

pleacu:
redone - see new PR

Issue #382, opened 2018-04-16
[regression] Index of last target field is lost after separator change 


mmelko:
I'm experiencing weird behaviour when using "Separate". Index of last target is lost right after separator change. Index input field is not visible in UI (see attached image) and in actual mapping is index set to '0'.
![screen shot 2018-04-16 at 16 59 50](https://user-images.githubusercontent.com/6814482/38817047-ccd83382-4197-11e8-9521-ab03247f5f3e.png)

PR #387, opened 2018-04-17
fix: When the user changes a mapping details action it is nether sour…

pleacu:
…ce or target component specific (isSource=false).

Fixes: #382


jpav:
Won't this change fix the issue with separate, but cause a similar problem with combine?

pleacu:
No - coming from selectionChanged() in transition-selection.component.ts (someone updating the separate or combine token), the isSource parameter should always be false.  It's used down the line in updateSeparateOrCombineIndex() to remove field actions when in separate mode.  In thi8s case there is no selected element - just a selected component action.

Issue #386, opened 2018-04-17
Consider using dropdown's for specifying units in ConvertUnits transformation 


mmelko:
Input fields are used for setting units in Convert [Area,Distance,Mass,Volume] Units without any tooltips or help. 
![screen shot 2018-04-17 at 13 17 33](https://user-images.githubusercontent.com/6814482/38866654-d81f2002-4241-11e8-86a7-fd9856fbceb7.png)

igarashitm:
Ugh this is ugly. As those units are enum at runtime side, I would say this is a bug in UI.

igarashitm:
But it might not small enough to touch a few days before code freeze.

igarashitm:
Just noticed this is not trivial. We should avoid mixing different unit type, like mass unit for from and distance unit for to. Some advanced form control is required.

mmelko:
I see. I'm not sure how important is to have this transformation in 7.0 GA. Personally I think it would be better to disable it for GA and rework for 7.1 or include tooltips for possible values at least.

pleacu:
Implemented pulldown menus for the conversion values defined by the runtime:
![dm1](https://user-images.githubusercontent.com/2981421/39195255-71eca45a-47ad-11e8-9636-ce632efc0f23.png)

Error if the user attempts to convert to the original type:

![dm2](https://user-images.githubusercontent.com/2981421/39195264-771ec21e-47ad-11e8-8cd1-dcb4ac0c835f.png)

Cleaned up user-visible unit values:

![dm3](https://user-images.githubusercontent.com/2981421/39195270-79de97f4-47ad-11e8-8c4a-c5e0d677b5d7.png)

resulting xml:

![dm4](https://user-images.githubusercontent.com/2981421/39195278-80f9b7b2-47ad-11e8-8127-416b2930b99f.png)

TovaCohen:
Just noting here that I updated the doc to say to use the dropdown menus to select the units for the ConvertTo... transformations. You can see the updated doc here:
https://access.qa.redhat.com/documentation/en-us/red_hat_fuse/7.1/html-single/integrating_applications_with_ignite/#available-transformations

PR #392, opened 2018-04-18
fix: Deliver possible values for enum field action parameters

igarashitm:
Partially addresses #386
Now FieldActionArgument.values can be used to show possible values

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
@@ -447,13 +447,14 @@ export class MappingManagementService {
               fieldActionConfig.method = svcConfig.method;
               fieldActionConfig.serviceObject = svcConfig;
 
-              if (svcConfig.parameters && svcConfig.parameters.property
-                && svcConfig.parameters.property.length) {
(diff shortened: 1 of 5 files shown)


ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
```diff
@@ -447,13 +447,14 @@ export class MappingManagementService {
...
+              if (svcConfig.parameters && svcConfig.parameters.parameter
+                && svcConfig.parameters.parameter.length) {
+                for (const svcParameter of svcConfig.parameters.parameter) {
                   const argumentConfig: FieldActionArgument = new FieldActionArgument();
-                  argumentConfig.name = svcProperty.name;
-                  argumentConfig.type = svcProperty.fieldType;
-                  argumentConfig.serviceObject = svcProperty;
+                  argumentConfig.name = svcParameter.name;
+                  argumentConfig.type = svcParameter.fieldType;
+                  argumentConfig.values = svcParameter.values;
```
igarashitm: Hi @pleacu - so this `FieldActionArgument.values` is supposed to be an array of possible values. Can we make a pulldown menu when this has any actual value?

PR #417, opened 2018-05-03
Add NUMBER check in JsonFieldWriter

### LINKED ISSUE #415: "null" value during mapping to JSON number type
JsonFieldWriter doesn't consider `NUMBER` type, so it's always `null`.

mmelko:
fix #415

### DIFF
--- a/runtime/modules/json/core/src/main/java/io/atlasmap/json/core/JsonFieldWriter.java
+++ b/runtime/modules/json/core/src/main/java/io/atlasmap/json/core/JsonFieldWriter.java
@@ -234,7 +234,7 @@ public class JsonFieldWriter implements AtlasFieldWriter {
             valueNode = rootNode.booleanNode((Boolean) value);
         } else if (FieldType.INTEGER.equals(type)) {
             valueNode = rootNode.numberNode((Integer) value);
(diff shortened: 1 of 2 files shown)


igarashitm:
Thanks for the PR, but right now 1.34.x branch is for blocker only and I don't think it's a blocker. We'd want to keep it only in master for now.

mmelko:
I see, but from my POV it could be blocker. You basically can't map to json if it's specified by JSON schema and uses number types. 
```json 
sourceSchema: `{  
          "type" : "object",
 "properties" : {
          "sourceJsonString" : {
            "type" : "string"
          },
          "sourceJsonInteger" : {
            "type" : "integer"
          },
          "sourceJsonLong" : {
            "type" : "number"
          },
}
```

For example sourceJsonLong will be always `null`

igarashitm:
Ugh right, I almost forgot `number` is a primitive type in terms of json schema. Can you rebase this against latest 1.34.x branch?

mmelko:
It seems that it's already up to date.

```
git rebase upstream/atlasmap-1.34.x
Current branch atlasmap-1.34.x is up to date
```

PR #428, opened 2018-05-11
 fix: Unsubscribe abandoned Subscription, or use Promise for one-time callback

igarashitm:
Fixes: #111 

chore: Remove redundant converter which cause warnings


jpav:
What was wrong with the toZonedDateTime methods?

igarashitm:
@jpav those are duplicates causing warnings.

jpav:
ah, okay

Issue #432, opened 2018-05-15
Upgrade to Angular 6


igarashitm:
Just filing as a placeholder, let's see when we can tackle this
@seanforyou23 feel free to take over and try whenever you get a chance!

igarashitm:
I made a quick try and hit this issue, got `Unknown option: '--reporters'` from `ng test`
angular/angular-cli#10834

igarashitm:
I created a separate branch to test Angular 6 here
https://github.com/atlasmap/atlasmap/tree/angular6

Let's play around there and merge it back to upstream once it gets working.

igarashitm:
just removed `--reporters`. The next wall - dherges/ng-packagr#767

deeleman:
Just got an alert regarding this. My two (well, 3 actually) cents are:

- Check https://update.angular.io and follow the steps pointed out by the wizard.
- Upgrade the CLI and let it carry out all the heavylifting for you. Some peer dependencies such as RxJs will have to be upgraded in regards of implementation. I already introduced lettable operators in my latests implementations, butall the legacy ones still observe the (now deprecated) legacy chainable interface. Loiane Groner wrote a very useful walkthrough about how to handle the upgrade, available at https://loiane.com/2018/05/upgrading-to-angular-v6/
- With the new Angular CLI in place, now you can build component libraries as JS artifacts by executing `ng generate library atlasmap`. This would allow you to phase out ng-packagr should you guys do not want to keep taking advantage of it. Please refer to the official docs and the CLI changelog (hint: The "Hamilton" release):https://github.com/angular/angular-cli/releases

igarashitm:
Hmm, it seems `ng generate library @atlasmap/atlasmap.data.mapper` is not allowed,
```
Project name "@atlasmap/atlasmap.data.mapper" is invalid.
```
So it doesn't match with current NPM package name to be published.

deeleman:
Problem here is that NPM package name has been always wrong I'm afraid (package names should not contain dots, but dashes). Perhaps this would be a good time to fix that, now that AtlasMap is mostly used solely on Syndesis. Otherwise this might become a bigger issue in the future.

;)

igarashitm:
Ah OK, changing package name to dash-connected one is fine. The project name in the angular.json is already using dash, so we'd want to align sooner than later.

PR #447, opened 2018-05-31
fix: Upgrade to Angular 6

igarashitm:
Fixes: #432

now `yarn build:lib` runs `ng build @atlasmap/atlasmap-data-mapper-lib`. The output looks still weird, *.ts files are scattered over the dist/lib dir which looks wrong. It needs to be fixed.


igarashitm:
Verified it just works even within Syndesis (without upgrading syndesis-ui to angular6). I think it's time to merge into master and see what happen.

Issue #278, opened 2018-02-08
Enable source side transformation


igarashitm:
Today we only have target side transformation (field action) enabled in UI. But some of transformations (e.g. NumberFieldActions) expects a collection as an input and produce non-collection output. This is not possible with target transformation as it only applies with target field type.

Source:(Array[Number]) -> SumUp transformation(Array[Number]->Number) -> Target:(Number)

To utilize those transformations, we need to enable source side transformation. Apparently we already have it in runtime, hopefully it just works once it's enabled in UI.

jpav:
I would suggest that maybe we need **_only_** source-side transformations, with the last transformation being a map/split/combine transformation (probably involving the grouping transformation that would implemented by #244).

igarashitm:
Look&feel / wording aside, we do need both of pre-mapping and post-mapping transformation, e.g. String->Date mapping with `AddDays` transformation.

jpav:
I disagree.  That example can be accomplished just as easily by doing 1) String->Date conversion, 2) Add days to result, and finally 3) map result to target.

mattrpav:
I agree with that supporting FieldActions on both source and target fields is a useful feature

jpav:
Is there an example you can think of where it would be better with both? It seems to me that having both complicates the user experience, and most likely the code base.

mattrpav:
@jpav not sure how it would complicate the code base. Seems to me it should use same code path as target field action processing?

I believe it’s common in other mapping engines and if you take the transform, then split or map to multiple output it would be handy to perform the action once

jpav:
The main thing I was thinking of with the code base is having to constantly be aware of which 'field' context you're dealing with, source or target, and potentially having different rules for each. But maybe that's an over concern. Not sure if I'd say other mapping engines doing this is a good argument. Almost every other one I've seen sucks :-) For what it's worth, the one's I've worked on with Teiid and ModeShape didn't work that way. But considering UX, I'm mainly thinking it would be an awfully simple and straight forward approach to have the workflow be choose source(s), optionally perform one or more transformations on the data, and lastly map to the appropriate targets. I'm also considering possible complex scenarios where there may be multiple chained transformations involved, along with multiple transformation paths from different sources before the final values are set in the target fields. Allowing the mappings to appear anywhere in the middle of this process just seems to complicate things.

jpav:
[...]

PR #507, opened 2018-07-10
Fix for #278: Enable source side transformation

jpav:


syndesisqe:

PR #509, opened 2018-07-12
 fix: don't allocate memory exponentially in collection mappings

### LINKED ISSUE #508: Exponential memory usage with JSON collection mapping
I've noticed exponential memory usage with JSON collection mapping. Seems to be stemming from the expansion of collection mapping in DefaultAtlasContext.

zregvart:
`DefaultAtlasContext::extractCollectionMappings` has the a side effect
of modifying the mapping, thus for already expanded collection mappings
it expands again and replaces the mapping.

For example, given source document of:
```json
{ "array":
  [
    { "property": 1 },
    { "property": 2 },
    { "property": 3 }
  ]
}
```

And mapping containing a JSON field with path like `/array<>/property`.

First invocation modifies the mapping to contain three mappings:
 1. `/array<0>/property`
 2. `/array<2>/property`
 3. `/array<3>/property`

Since this is now replacing the previous mapping on the next invocation
this expands to:

 1. `/array<0>/property`
 2. `/array<1>/property`
 3. `/array<2>/property`
...
 9. `/array<8>/property`

This removes the mapping mutation side effect, though I'm not entirely
sure why it was there to begin with, could be for optimization?

Fixes #508

### DIFF
--- a/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
+++ b/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
@@ -488,7 +488,7 @@ public class DefaultAtlasContext implements AtlasContext, AtlasContextMXBean {
         }
     }
 
-    private List<Mapping> extractCollectionMappings(DefaultAtlasSession session, BaseMapping baseMapping)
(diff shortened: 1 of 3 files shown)


syndesisqe:

Issue #504, opened 2018-07-04
Drag and drop doesn't work in Firefox 


mmelko:
I'm unable to use drag n drop when creating mapping or changing indexes in combine/separate. I have Firefox 61

igarashitm:
Verified it doesn't work on firefox, I'm on Fedora28.

pleacu:
https://support.mozilla.org/en-US/questions/1166501#answer-983577

igarashitm:
escalated to p1 - a bit more love to firefox users ;-) but anyway if it's firefox bug we don't have much we can do.

mmelko:
Hi it still doesn't work (1.35.7 in syndesis)

igarashitm:
Verified it now works fine on both of master and atlasmap-1.35.7 tag of AtlasMap, with Fedora28 firefox (61.0.2).

pleacu:
thkxTomo!

igarashitm:
@mmelko do you still see this?

mmelko:
I'm still not able to create mapping by drag'n'drop. I can switch indexes in combine/separate though.

igarashitm:
OK so you mean drag source field and drop into target to create a mapping. @pleacu do we have same problem there with what you already fixed for combine/separate index?

pleacu:
I've got a fix once my previous push has cleared.

igarashitm:
great! will merge it after release work completes.

pleacu:
This is now fixed - please verify.

PR #517, opened 2018-07-20
fix: Correct drag&drop on firefox.

pleacu:
Fixes: #504


syndesisqe:

Issue #499, opened 2018-07-03
Reset the "Mapping Detail" panel, when a user delete a mapping in look up mode


nding-anges:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
In look up mode, deleting the mapping first and then create a new mapping, the mapping panel will not automatically be reset.

## Expected behavior
The mapping panel could automatically be reset when a user delete a mapping in look up mode.

## Screenshot

<img width="613" alt="2018-07-03 5 55 13" src="https://user-images.githubusercontent.com/34368385/42213068-67c8237c-7eea-11e8-9979-f21f0263ddf2.png">

PR #518, opened 2018-07-20
fix: Remove LookupTable when corresponding mapping is removed

igarashitm:
Fixes: #499


syndesisqe:

Issue #505, opened 2018-07-04
change 'Use ctrl-M1 to select multiple elements for 'Combine' or 'Separate' actions..' when running on MacOS


mmelko:
Currently mentioned warning might be confusing for MacOS users. Ctrl-m1 is represented as right click. It should really be 'Use cmd+M1 ...' when atlasmap is running on MacOs.

PR #519, opened 2018-07-23
Fix for #505: change 'Use ctrl-M1 to select multiple elements for 'Co…

jpav:
…mbine' or 'Separate' actions..' when running on MacOS

Fixes issues #505


syndesisqe:

Issue #521, opened 2018-07-25
Drag & drop of padding field can confuse index


pleacu:
![am1](https://user-images.githubusercontent.com/2981421/43222372-f4126b6e-901d-11e8-93de-128c38d2d6c5.png)

PR #522, opened 2018-07-26
fix: Correct artificial padding field document definition so source-s…

pleacu:
…ide d&d will function.

Fixes: #521


syndesisqe:

Issue #485, opened 2018-07-02
Apply UXD outcome for data preview - single entry


igarashitm:
Apply #455 outcome published here - https://redhat.invisionapp.com/share/MAMPCU2CNFR.

I placed `TODO`s to where we need to work on
* Preview input/output in card view: https://github.com/atlasmap/atlasmap/pull/475/files#diff-2e8307644c73900458654f91b680a90e
* Preview input/output in table view: https://github.com/atlasmap/atlasmap/pull/477/files#diff-d00fbe27f3af7bb3670cc9340af5ad1e

PR #523, opened 2018-07-26
fix: Apply UXD outcome for data preview - single entry

igarashitm:
Fixes: #485


syndesisqe:

PR #525, opened 2018-07-27
Fixes #515: Prepend FieldAction doesn't show up for STRING target field

### LINKED ISSUE #515: Prepend FieldAction doesn't show up for STRING target field
As targetType is already `STRING`, I guess it's a UI bug when filtering what to show
https://github.com/atlasmap/atlasmap/blob/e37825f49afc3e461420d2b4cd15c356dca6e6c6/runtime/core/src/main/java/io/atlasmap/actions/StringComplexFieldActions.java#L205

jpav:
Fixes #515


syndesisqe:

Issue #375, opened 2018-04-11
Improve error diagnostic if a user attempts multiple mappings (map mode) to the same target element


pleacu:
![image](https://user-images.githubusercontent.com/2981421/38618182-dda1ee0e-3d66-11e8-9a07-99af0392abf2.png)

from @apupier 

@pleacu message error improvement suggestion: currently the message is mentioning that there is "another" mapping. it would be nice to be help user to find which one. maybe mention the source? or have a link jumping to this mapping? (this can be done in a second iteration for sure)

PR #526, opened 2018-07-27
fix: Correct target mapping error message to indicate existing source…

pleacu:
… element.

Fixes: #375


syndesisqe:

PR #527, opened 2018-07-28
461

jpav:
- Renamed some classes and merged enums with associated interfaces
- Moved file attribute from context to MappingDocument
- Eliminated Session since all execution can be done within context of MappingDocument
- Moved messages to MappingDocument and added some initial code for validation of document and contents
- Added error handling for no default constructor in SPI classes
- Added unsaved indicator and API to check
- Merged some UI changes from master


syndesisqe:

Issue #495, opened 2018-07-03
Add Inactive Status to ‘Editor setting’ button


nding-anges:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Currently on Data Mapper, the ‘Editor Setting’ button is always highlighted no matter activated or not.

## Expected behavior
- The ‘Editor Setting’ button should be displayed by dark gray color while inactive Status, which is similar color with "Add a mapping" for example.

## Screenshot

Active Status of "Editor Setting" button
<img width="288" alt="2018-06-15 3 50 55" src="https://user-images.githubusercontent.com/34368385/42211956-59894712-7ee7-11e8-97f3-183b0e109274.png">

Inactive Status of "Editor Setting" button

<img width="1307" alt="2018-06-19 12 28 53" src="https://user-images.githubusercontent.com/34368385/42211982-6de68d5a-7ee7-11e8-96f8-f4cdeb545fc6.png">

igarashitm:
I got what is happening a bit more, so the gear is blue from beginning because `Editor Settings` -> `Show Lines` is enabled by default. Once you disable `Show Lines`, then the gear turns into black. So it shows blue if at least one of its sub-option is enabled.
@nding-anges do you think we still want to make it blue only when a dropdown list of sub-options are shown?

igarashitm:
We might want to make the sub-option icon blue instead.

nding-anges:
Hi @igarashitm  Since the sub-menu is by default hidden and for 90% of time, there will at lease  one sub menu is enable, as a new user , they won’t understand why the gear looks always active.

I'd like suggest that the active status of the gear as well as other icons on the top right will indicate they open a new panel.For example: icon of Open/Close Mapping detail panel.  In this case , I think that the gear will be blue when-sub menu panel is open. That will be consistent action as other icons . 

At the same time,  I’d like suggest to improve the design of the sub-menu in oder to better visualize their enable and disable status. This is related  to another Issue #77

igarashitm:
@nding-anges sounds cool! feel free to file separate issue to improve sub-menu design.

PR #532, opened 2018-07-30
fix: Make "Editor Settings" gear icon blue only when it's open

igarashitm:
Fixes: #495


syndesisqe:

PR #535, opened 2018-08-01
Update for auto-conversions and validations

jpav:
Plus some miscellaneous cleanup


syndesisqe:

Issue #537, opened 2018-08-02
Disable CustomAction parameter


igarashitm:
Makes no sense to introspect `CustomAction` class...
https://github.com/atlasmap/atlasmap/commit/ebd4aaa28e2fe692a51b8cc7addbf7aea154d94e#diff-dde7432b29519d7cac653e23213c06faR88

igarashitm:
Supporting parameters on CustomAction requires a bit more work, will address in a separate issue.

Issue #483, opened 2018-06-29
Support receiving FieldActions metadata from outside


igarashitm:
Accept FieldActions metadata from outside in UI instead of retrieving from `/fieldActions` REST API. syndesisio/syndesis#204 requires AtlasMap UI to be fed FieldActions metadata from outside, while today it always retrieves from `/fieldActions` REST API.

PR #539, opened 2018-08-02
feat: Support receiving FieldActions metadata from outside (#483)

igarashitm:
fix: Disable CustomAction parameter
Fixes: #537

feat: Support receiving FieldActions metadata from outside
Fixes: #483


syndesisqe:

Issue #544, opened 2018-08-28
Constants value is not saved into mapping definition


igarashitm:
A regression caused by #475, constants value is ignored when data preview is disabled. Constants values should be always serialized.

PR #545, opened 2018-08-28
fix: Constants value is not saved into mapping definition

igarashitm:
Fixes: #544


syndesisqe:

Issue #489, opened 2018-07-03
Modal Overlay consistency with Patternfly  


nding-anges:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Modal Overlay UI is inconsistency with Patternfly. 
- When the modal overlay appears, the background did not darken to bring attention to the primary content.
- Font weight of title is not using Blod style.
- Close  button is " PatterFlyIcons-webfont" with 18px size
- Cancel button style is inconsistency with Patternfly. 

## Expected behavior

Keep UI consistency with Patternfly for Modal Overlay.
http://www.patternfly.org/pattern-library/forms-and-controls/modal-overlay/#design
- The background should darken when the modal overlay appears.
- Font weight of title should use Blod.
- Cancel button should use Secondary Button style. See an example here:https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/buttons.html

## Screenshot

<img width="1373" alt="2018-07-02 2 35 02" src="https://user-images.githubusercontent.com/34368385/42207114-b0b95f42-7edb-11e8-9e95-b4f5c30bb323.png">

nding-anges:
Uploaded UI Design 

![delete](https://user-images.githubusercontent.com/34368385/42309175-6b7e204e-806a-11e8-8312-b411eb1770be.png)

 
![select mapping](https://user-images.githubusercontent.com/34368385/42309155-6147f8ca-806a-11e8-8b29-646ac413df8e.png)

nding-anges:
@igarashitm Hi Tomo, please help to review the above UI design. Thank you~!

PR #546, opened 2018-08-30
fix(modal): update title font, close icon, cancel button, modal fade in/out

michael-coker:
fixes https://github.com/atlasmap/atlasmap/issues/489


syndesisqe:

Issue #573, opened 2018-09-20
StackOverflowError when processing FHIR xsd


rkorytkowski:
If you try to load the FHIR xsd file from https://www.hl7.org/fhir/fhir-single.xsd (~2 MBs), you will hit StackOverflowError. Note I needed to replace:
```
<xs:import namespace="http://www.w3.org/XML/1998/namespace" schemaLocation="xml.xsd"/>
<xs:import namespace="http://www.w3.org/1999/xhtml" schemaLocation="fhir-xhtml.xsd"/>
```
with:

`<xs:import namespace="http://www.w3.org/1999/xhtml" schemaLocation="http://www.hl7.org/implement/standards/fhir/fhir-xhtml.xsd"/>`

to go around an issue with a not defined base uri.

The stacktrace is:

```
16:04:57.475 [XNIO-3 task-16] ERROR io.undertow.request - UT005023: Exception handling request to /v2/atlas/xml/inspect
org.jboss.resteasy.spi.UnhandledException: java.lang.StackOverflowError
	at org.jboss.resteasy.core.ExceptionHandler.handleApplicationException(ExceptionHandler.java:78)
	at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:222)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:171)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:414)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:205)
	at org.jboss.resteasy.plugins.server.servlet.ServletContainerDispatcher.service(ServletContainerDispatcher.java:221)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletDispatcher.service(HttpServletDispatcher.java:56)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletDispatcher.service(HttpServletDispatcher.java:51)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at io.undertow.servlet.handlers.ServletHandler.handleRequest(ServletHandler.java:85)
	at io.undertow.servlet.handlers.FilterHandler$FilterChainImpl.doFilter(FilterHandler.java:129)
	at org.springframework.boot.web.filter.ApplicationContextHeaderFilter.doFilterInternal(ApplicationContextHeaderFilter.java:55)
[...]

Issue #577, opened 2018-09-21
OOM during FHIR XML schema inspection


igarashitm:
I fixed #573 and then hit OOM...
It loaded `9,075,113` fields and died. A bit surprising that 9 millions of fields could be loaded as a AtlasMap Document model. 

So definitely loading whole FHIR schema is unrealistic. Browser would crash even with thousands of fields. Nobody like to scroll hundreds of fields in AtlasMap UI and map one by one.

How can we slim it down to realistic number of fields? Is it even possible? We would need some "filtering" on inspection, as well as deeper understanding of FHIR object model and real usecases to figure out how to screw this gigantic object in to AtlasMap UI.

rkorytkowski:
Fortunately, one will be interested in mapping only some resources and not the entire FHIR model, e.g. patient and observation. We could import https://hl7.org/fhir/2018May/fhir-base.xsd, which defines data types used across resources and then load on demand one or more resource definitions from https://hl7.org/fhir/2018May/fhir-all.xsd. We would have to have a way to load interesting parts of the model, based on user demand, e.g. using approach suggested in #579 and only load and process imports when related resources are expanded.

rkorytkowski:
To give you a specific example a typical use case would be to create an integration invoking e.g. an encounter rest endpoint (http://hapi.fhir.org/baseDstu3/Encounter?date=2018-01-01) and mapping an Encounter to some internal model. It is also possible to request a bundle, which may contain a number of resources of different types created in a transaction (e.g. http://hapi.fhir.org/baseDstu3/Bundle/6162125) thus we need to support mapping resources of different types at once eventually. One may also want to create different FHIR resources at once when mapping from some other model to FHIR.

igarashitm:
So I would say #579 is the right direction unless FHIR usage has some incompatible requirements we can't implement in generic xml module. But it's a lot of work... so if it's easier, we could implement fhir module for the initial support, then migrate into xml module once #579 is addressed at some point. Of course #579 is much more efficient in total cost.

By the way, if you can slim XSD down at configuration time in terms of Syndesis action configuration, then you can provide a dynamic DataShape with tailored XSD to AtlasMap, then it might just work without #579 if that tailored XSD is slim enough.

igarashitm:
@rkorytkowski Here is the sql connector example which builds dynamic DataShape specification from sql statement parameters, it's JSON schema though
[...]

Issue #255, opened 2018-01-26
Recursive class doesn't even expand


igarashitm:
The fix for #81 finally revealed the "right" issue, recursive class can't expand in UI. UI doesn't support expanding cached class. Now `/quotedStats` doesn't expand as it's recursive twitter4j.Status class.
![screenshot from 2018-01-26 16-43-33](https://user-images.githubusercontent.com/265462/35462179-e3fbaecc-02b8-11e8-8aa3-4ddd9f533066.png)

We should
* Gray out the recursive(cached) field until expand
* Un-gray out and show fields once it's expanded

rohanmars:
It might be a different issue, but I think it's related, but referencing the same type form multiple times in an object wont populate (or automatically display) the second instance sub fields of that type. The UI should probably pre populate the cache based on the cached values, so that search works across repeating fields. It would have a potential impact on performance, but the priming could be capped at some magic field count.

Greying out recursive fields is a good idea!

igarashitm:
Yep, sounds like a same root cause. Updated the title and description.

pleacu:
![i255](https://user-images.githubusercontent.com/2981421/36104937-5a80a70c-0fe1-11e8-9a93-eb2918d316af.png)
Kept the expansion arrow so users can expand and have content filled in when un-caches.

jpav:
Seems like we should also have a tooltip that explains why the field is grayed out, i.e., that it's recursive.

PR #578, opened 2018-09-21
fix: XML schema inspection should cache complex type to handle recurs…

igarashitm:
…ive item

Fixes: #573

Still it can't inspect whole FHIR schema due to its incredible size (#577)


runtime/modules/xml/core/src/main/java/io/atlasmap/xml/inspect/SchemaInspector.java
```diff
@@ -317,44 +321,48 @@ private void printSchemaSet(XSSchemaSet schemaSet) throws Exception {
...
         }
     }
 
-    private void printGroup(XSModelGroup modelGroup, String rootName, XmlComplexType xmlComplexType)
+    private void printGroup(XSModelGroup modelGroup, String rootName, XmlComplexType xmlComplexType, Set<String> cachedComplexType)
             throws Exception {
         // this is the parent of the group
         for (XSParticle particle : modelGroup.getChildren()) {
-            printParticle(particle, rootName, xmlComplexType);
+            Set<String> cachedTypeCopy = new HashSet<>(cachedComplexType);
```
rkorytkowski: It's not obvious, why you need to make a copy of cachedComplexType here.

igarashitm: Ah I should have left a comment - this is to avoid caching siblings, as we don't support expanding cached field in UI yet (#255). This is another reason we should address #579.

rkorytkowski: Thanks! Makes sense. Indeed a comment would be great.

Issue #461, opened 2018-06-08
Handle simple transformations currently unaddressable


jpav:
We need to discuss how to handle a host of very simple use cases that we can't currently address. Consider, for example, if the user wants to create a mapping that involves taking two numeric fields from a source and from those create a target equation, as text, that represents the sum of those two numbers. For instance, if the source instance document contained 2 and 3 in these two numeric fields, the desired result in the mapped target field would be '2+3=5'. Even with the provision of additional actions, enabling source transformations, etc., I don't believe we have any good way of handling this.

nding-anges:
Hi Tomo @igarashitm , Is this also one type of transformations, when user select some numeric data on source and target field and it could do math operations? So it could automatically appear in the transformation list where it's applicable. There's no specific UI design required,I suppose :-)

igarashitm:
Right, no new design is required for this one.

jpav:
After looking into alternative approaches to address my concerns with this issue, I still have been unable to find a suitable approach that doesn't involve rewriting much of the runtime. So, I want to first iterate in detail what I see as our current deficiencies from a feature perspective, what we'd need to provide the abilities, and finally why our current architecture doesn't seem to allow for a good solution. After all of that, I will discuss the alternative solution I have been working on that addresses all of these concerns. Fair warning, this will certainly be viewed by many as TL;TR.

But I want to first point out that _all_ of these concerns are based off of abilities and features known to exist in our competitors' products and even in previous attempts at our own solutions that never fully matured.

**Missing Features**: 

1. The example described in the initial description of this issue, mapping the addition of two source fields into a target field, currently could only be accomplished if we introduce additional transformation 'strategies' rather than field actions. There are many additional operations of this type that could easily be envisioned and that indeed exist in our competitors' solutions. But adding 'strategies' along side types would be very confusing to users, with the only real difference is based on technical design. We've discussed a grouping transformation to make certain transformations possible, with the hope of one day eliminating the need for strategies, but the process as a whole would remain very unintuitive to the user, with the real result being they'd probably have no idea we even had the capabilities. For instance, I feel confident in saying a user would never to think they'd first need to group one or more source elements into some type of collection before they could actually add them together (or whatever similar class transformation they might want to perform) and map the results to a target field.

[...]

Issue #338, opened 2018-03-08
Allow field actions to receive whole collection


igarashitm:
ATM collection mapping is extracted to a set of mappings for each entries anyway, which prevents field actions to consume whole collection. This needs to be refactored.
https://github.com/atlasmap/atlasmap/blob/d8c2b186c26aaf7a6d703118114c6a4ef17098fe/runtime/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java#L295-L362

igarashitm:
Sorry I think uxd label here was not accurate, this task doesn't involve design change. Once runtime and UI code handle them correctly, right thing should be shown.

This however increases the importance of supporting indexed item (#30), which involves design change.

PR #602, opened 2018-10-03
feat: Handle FieldGroup contributed via mapping definition

igarashitm:
Fixes: #550

This allows transformations to change multiplicity, which is supposed to replace `COMBINE`/`SEPARATE` mode eventually. UI needs to add `Concatenate`/`Split` transformation instead of using `COMBINE`/`SEPARATE` mode, and for `Concatenate` case the source fields needs to be wrapped with `FieldGroup`, so that the `Concatenate` transformation is placed on a `FieldGroup` level.
[An example mapping is here](https://github.com/atlasmap/atlasmap/pull/602/files#diff-11b53e171519d064b287692ae917660e)

This also allows transformations like `Add` mentioned in #461, which consumes multiple number fields and sum up. In combination with #338 we can also apply it for collection field.

At this point we don't support to have a collection field as a part of grouping, i.e. put collection field and non-collection field into a same group, so UI should keep prohibiting it. We can revisit it at some point.

### DIFF
--- a/docs/src/main/asciidoc/inc/swagger/core/definitions.adoc
+++ b/docs/src/main/asciidoc/inc/swagger/core/definitions.adoc
@@ -228,6 +228,41 @@ __optional__|object
 |===
 
 
+[[_atlas-service-core-fieldgroup]]
+=== FieldGroup
+
+[options="header", cols=".^3a,.^4a"]
+|===
+|Name|Schema
+|**actions** +
+__optional__|<<_atlas-service-core-actions,Actions>>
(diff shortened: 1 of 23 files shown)


igarashitm:
One thing is that this is built on top of existing model without breaking, so `COMBINE`/`SEPARATE` mode still works like today. This allows us to gradually migrate to that direction from UI side.
I had to make some refactoring around internal SPI and Java field writer though. We'll need to enrich unit tests around that from next week.

Issue #385, opened 2018-04-17
DayOf{Month|Week|Year} transformation doesn't show up for Date type source field


mmelko:
DayOfWeek,DayOfYear transformations are missing in UI when mapping from Date to Integer. I'm sure I've seen those transformations before. 

Also it would make sense to add DayOfMonth transformation. (or did I miss it?)

igarashitm:
Date/Time field actions appear only when target field is Date/Time field. There were some bugs field action is applied to source field in very limited cases, but disabled it as it's confusing.

@jpav can you respond about `DayOfMonth`?

mmelko:
Yeah I see, but DayOfYear was number and DayOfWeek too(but could be also string such as 'Monday'..). I agree that it's confusing.

igarashitm:
Those field actions consume Date/Time and return a number. In order to apply field actions on target field, both of input and output must be Date/Time.
Those field actions only make sense for source field, which requires #278 to be implemented ahead.

jpav:
Now that I'm getting more knee-deep in the UI code, I really see what a mess this is. It's definitely very confusing the way things are now. I'm now thinking we're doing this completely backwards in the UI. The "flow" of a transformation should start with the source field(s), followed by one or more transformations, then end with the mapping(s) to the target field(s). With this in mind, it seems to me that we really need the UI to do the opposite of what it's currently doing and provide **_only_** source field transformations (i.e., field actions), with the end transformation being either a mapping, combine, or separate. Having said all that, since things work the way they do now, these DayOfXXX functions basically aren't currently supported by the UI, so although I agree it would otherwise make sense to have a DayOfMonth transformation too, it's a moot point at the moment.

jpav:
Blocked by #278

nding-anges:
Hi @igarashitm Since we have already Source side transformation, I suppose that we could move forward to design the DayofMonth transformation now? And this kind of transformation appear only  when user select a Date and time value on source and  target field? Currently, we could support Date, time, month and year as a type of value? Thanks~

nding-anges:
Great~Thanks!Tomo @igarashitm

PR #628, opened 2018-10-12
fix: Correct DayOf transformation inclusion

pleacu:
Fixes: #385

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/models/transition.model.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/models/transition.model.ts
@@ -124,7 +124,9 @@ export class FieldActionConfig {
 
       // Check for matching types - date.
       if (this.matchesDate(this.sourceType, selectedSourceField.type)) {
-        if ((this.multipleTransformations(fieldPair)) || (this.matchesDate(this.targetType, selectedTargetField.type))) {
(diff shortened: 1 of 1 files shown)


ui/src/app/lib/atlasmap-data-mapper/models/transition.model.ts
```diff
@@ -124,7 +124,9 @@ export class FieldActionConfig {
 
       // Check for matching types - date.
       if (this.matchesDate(this.sourceType, selectedSourceField.type)) {
-        if ((this.multipleTransformations(fieldPair)) || (this.matchesDate(this.targetType, selectedTargetField.type))) {
+        if ((this.multipleTransformations(fieldPair)) || (this.matchesDate(this.targetType, selectedTargetField.type)) ||
+            ((this.targetType === 'INTEGER') && ((this.name === 'DayOfWeek') || (this.name === 'DayOfMonth')) ||
```
igarashitm: weren't they fit with the rule? Since these transformations have different type between source and target, it only needs to appear as source transformation where source field type is Date/Time.

pleacu: yes - they don't need to call out the functions any more - will fix

Issue #624, opened 2018-10-12
mapping table: consider field order in separate/combine by index rather than alphabetically 


mmelko:
I believe order by index is more user friendly especially when previews are used.
![screenshot_20181012_171709](https://user-images.githubusercontent.com/6814482/46878104-b2ef7100-ce42-11e8-95a3-cdcca0c62061.png)

pleacu:
Rather than overthink this...  :) the method I changed just deals with alphabetizing the content aof a field pair.  Note that the total list of mapped elements has always been in non-alphabetical order.

![ss1](https://user-images.githubusercontent.com/2981421/47184473-c42ff600-d2f8-11e8-92ca-be474ad73d8b.png)

PR #637, opened 2018-10-18
fix: Leave combine/separate elements in index order - not alphabetica…

pleacu:
…l order.

Fixes: #624


igarashitm:
one lint error
`ERROR: /workspace/ui/src/app/lib/atlasmap-data-mapper/components/mapping/mapping-list.component.ts[73, 9]: Identifier 'fields' is never reassigned; use 'const' instead of 'let'.
`

pleacu:
ugh - sorry - will fix
  --P

Issue #670, opened 2018-11-01
source-side transformations depends on target-type


mmelko:
I believe source-side transformations should take into account only source type. Currently we have different  source transformations in Number->String mappings than Number -> Number mappings. I don't see any reason why I cannot use i.e unit conversions on source side during Number -> String mapping.

pleacu:
I guess the logic was that if you select a string source-side object and you select a target side object to complete the mapping that it makes sense to restrict the available transformations to those that match the target.  If you're adding a transformation and a string -> string mapping it doesn't make sense to add an absolute-value transformation.  If however you have in mind a multi-stage source-side transformation then you could simply not choose a target mapping and have the full range of transformations available.

mmelko:
in that case .. isn't it really just another target-side transformation? It looks like you are transforming source as target type. Maybe I'm looking at it wrong way, but from my POV you can do everything just on target then and source side transformation is not needed as you are always get only subset of target transformations.

pleacu:
The other side of that coin is the user should perform all transformations on the source side and then do a single mapping to the target.  The way it is provides two knobs to turn that hopefully makes some logical sense to use.

mmelko:
Totally agree. In current design however it looks like first step is convert source to target and then do some transformations related to target type, which seems to be not ideal approach.(at least for me).

igarashitm:
So to summarize:
Now the source transformations are:
 * if target field is selected
   * show transformations which has source field type as an input type **AND** target-field-type as an output
 * else (target field is not selected)
   * show transformations which has source field type as an input type

The other option to consider:
  * show transformations which has source field type as an input type, no matter what the target field type is

pleacu:
@mmelko - please try this out now - ref PR https://github.com/atlasmap/atlasmap/pull/676

PR #671, opened 2018-11-01
fix: Rework the transformations acceptance to remove target type depe…

pleacu:
…ndency

Fixes: #670

### DIFF
--- a/runtime/core/src/main/java/io/atlasmap/actions/DateFieldActions.java
+++ b/runtime/core/src/main/java/io/atlasmap/actions/DateFieldActions.java
@@ -62,17 +62,17 @@ public class DateFieldActions implements AtlasFieldAction {
         return input.plusSeconds(addSeconds.getSeconds() == null ? 0L : addSeconds.getSeconds());
     }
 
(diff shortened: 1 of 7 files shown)


runtime/core/src/main/java/io/atlasmap/actions/DateFieldActions.java
```diff
@@ -62,17 +62,17 @@ public static ZonedDateTime addSeconds(Action action, ZonedDateTime input) {
         return input.plusSeconds(addSeconds.getSeconds() == null ? 0L : addSeconds.getSeconds());
     }
 
-    @AtlasFieldActionInfo(name = "CurrentDate", sourceType = FieldType.ANY, targetType = FieldType.ANY_DATE, sourceCollectionType = CollectionType.NONE, targetCollectionType = CollectionType.NONE)
+    @AtlasFieldActionInfo(name = "CurrentDate", sourceType = FieldType.ANY_DATE, targetType = FieldType.ANY_DATE, sourceCollectionType = CollectionType.NONE, targetCollectionType = CollectionType.NONE)
```
igarashitm: This seems like a `VOID` input kind of transformations

pleacu:
retry with pull 676

Issue #663, opened 2018-10-30
indexOf and lastIndexOf should have INTEGER as a targetType


igarashitm:
... as `String#indexOf()` returns int

Also need to check other transformations which return number types

PR #680, opened 2018-11-06
fix: Change indexOf and lastIndexOf target type to INTEGER in runtime…

pleacu:
… and add a typeahead wait of 900 ms to panel search

Fixes: #663
Fixes: #670 

I didn't change the target type for the NUMBER transformations - these two were the only non-NUMBER transformation functions that were returning NUMBER.


igarashitm:
wonder if it makes sense to also skip search if the word is 1 character?

pleacu:
Sure - I can look into that.

pleacu:
Actually upon further review... :) - Just talking with John G and he pointed out that it is likely someone could legitimately want to search for a '$' or '%' etc.  I think we're ok with a 900ms delay and then a search even for one char.

igarashitm:
Ah, OK fair enough :)

Issue #736, opened 2019-01-29
Add runtime support for variables in transformations


jpav:
Add support so that the results of transformations are placed in temporary variables that can be subsequently used by later transformations via references in parameters.

jpav:
I've checked in the initial support for this to branch Issue-#736, but, since this is still in a POC stage and to make it easier to see what I'm going for,  I should make the following notes:

- This initial support has no breaking changes, and therefore I haven't introduced any API or behavioral changes to existing code or mappings. Having said that, the consequence is that there is some ugliness involved. There is not enough context passed within downstream methods to provide transformations with the ability to access existing variables, so I created a static accessor to the active mapping within the DefaultAtlasContext class. That should be removed in the future as API changes are made to include the proper context.

- The current architecture doesn't provide for any way to name or store the results of transformations, so I'm currently auto-creating variable names.

- To again avoid too many changes to existing code, for now I've used the global source properties normally managed by the properties module to store the variables. These really need to be stored within each mapping, but that would require a number of additional attributes and supporting code, so I've saved that for later implementation.

PR #738, opened 2019-01-30
[DO NOT MERGE] Issue 736: Add runtime support for variables in transformations

paoloantinori:


pleacu:
Hey @paoloantinori - this is a {DO NOT MERGE} POC private branch commit. Are you asking for a master merge??

jpav:
Here is a screenshot of an example of what's currently supported via the code changes in this PR:

image

As stated in the issue I created for this PR, issue #736, the current design has no way to store the output of transformations in anything that's persisted or that can be viewed by the UI. Thus, the code is currently auto-generating variable names using simple indexing starting at 1 and I'm referencing the 'known' name of that variable in the screenshot. The idea would be to eventually have a icon on each transformation that shows the variable(s) in which it's output/results are stored, and possibly a separate list, similar to properties and constants, that would show the variables and their values. Also, variables would eventually be scoped to each mapping, not globally alongside properties as I'm doing now.

jpav:
I've illustrated several of the use-cases I'm trying to address in issue #461, but I'll specify a couple here:

1. Subtract a source field containing an order quantity from another source field representing an inventory count, then map the maximum of the result and zero to a string target field.

1. Add a substring of a source field to a numeric target field where the start and/or length of the substring is determined by other source fields.

chirino:
I like where this PR is going.  Being able to use variables/expressions as arguments to transformations is a powerful thing that will be an escape hatch for non-simple transformations.

Things that should be cleaned up before this PR is merged:
1: Lets just follow the excel expression cell approach/syntax to enable this feature.  Specifically, the use of the = and ' field prefix to enable or disable literal text vs expression mode for the cell.  Defaulting to literal text if it's not explicitly prefixed.
2: Implement some way to reference other existing document fields.  Perhaps $DOC/path/to/field. not sure what the best syntax would be.
3: Remove the auto creation of temp variables.  Once we have 2, we will not need this to demo the expression feature.

Some random thoughts for future PRs:
1) Eventually I'd like to make these expressions be very similar to excel expressions.  The type of user that would like to use this type of a datamapper is more similar to an excel power user than a java developer.  Could we get a future PR which adds a gramer to implement a simple expression parser that can reuse existing transformation functions?
2) Would like to have a future PR which adds a 'Set Variable' transformation which takes any input and returns the same input as output, but which gets configured with a variable name and perhaps a variable scope (global, document, current object) to control how long the variable is lives for.  Once we have this using the current variable references in this PR will make more sense.  
3) We should also consider if we can do variable dataflow analysis to make sure transformations which set variables are run before the transformations that reference variables.  Should cycles be forbidden so that the variables are effectively treated as finals.  I think this constraint would make the transformations using variables much safer and easier to reason about.

igarashitm:
Hiram's comment changed the world. That is nice to have. In addition, I would prefer introducing drag&drop from beginning, so drag source field into transformation parameter text box. So that typing math formula is left for complex usecases.

PR #741, opened 2019-01-31
Fix locale specific test failure

christophd:
Unit tests were failing on my machine as I have another default Locale setting.

Failing test:
https://github.com/atlasmap/atlasmap/blob/master/lib/core/src/test/java/io/atlasmap/actions/StringComplexFieldActionsTest.java#L141

Different local setting returns `"1.234,00"` and assertion fails.

I explicitly added root locale to String format in https://github.com/atlasmap/atlasmap/compare/master...christophd:fix/locale-specific-test-failure?expand=1#diff-e8deec4e7a8840412fdb2406fb2a66b5R123 so the Locale specific adjustments are disabled. 

Also fixed some typos ;-)


igarashitm:
Merged... 0 possibility to introduce new issue in codacy

Issue #740, opened 2019-01-31
Be compatible with VS Code Webview API


apupier:
VS Code WebView API is not supporting cookie/localstorage (and they don't want to [support it](https://github.com/Microsoft/vscode/issues/66207))

AtlasMap is using cookie/localstorage . Some details of investigation in the conversation of this PR: https://github.com/jboss-fuse/vscode-atlasmap/pull/9

to help setup environment for investigation see https://github.com/jboss-fuse/tooling_material/blob/master/atlasmap.md (ask to @paoloantinori for right access) or this mail https://post-office.corp.redhat.com/mailman/private/atlasmap-dev/2019-January/msg00013.html (a bit less detailed)

pleacu:
![vs1](https://user-images.githubusercontent.com/2981421/52236257-53efb500-2894-11e9-9547-bfdee0887f20.png)

There's another issue - see https://github.com/atlasmap/atlasmap/issues/744

PR #745, opened 2019-02-04
fix: Don't read cookies from local storage in the ApiXsrfInterceptor …

pleacu:
…- use default token value

Fixes: #740


ui/src/app/lib/atlasmap-data-mapper/services/api-xsrf-interceptor.service.ts
```diff
@@ -17,15 +17,19 @@ export class ApiXsrfInterceptor implements HttpInterceptor {
   constructor(private tokenExtractor: HttpXsrfTokenExtractor) { }
 
   intercept(httpRequest: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    /*
+      ADM doesn't need/use cookies so we'll use the default token value initialized from the environment when
+      adding the XSRF header to HTTP requests.  This will avoid the need to read cookies from local storage.
+    */
     if (this.cfg.initCfg.xsrfHeaderName) {
-      const token = this.tokenExtractor.getToken() || this.cfg.initCfg.xsrfDefaultTokenValue;
+      // const token = this.tokenExtractor.getToken() || this.cfg.initCfg.xsrfDefaultTokenValue;
```
apupier: why do you keep this commented line? You plan to add it back soon?

Issue #748, opened 2019-02-06
VSCode - import of an ADM does not show mappings (will show them on restart)


pleacu:
1.  Open AtlasMap from within VSCode
2.  Import the ADM file from the mock documents area
     All  files are imported and fields look good but the mappings do not appear
3.  Kill and reopen AtlasMap in VSCode
    All mapping lines are visible
4.  Import a file from AtlasMap running in the browser - all is well.

pleacu:
This appears to be the issue:
https://stackoverflow.com/questions/54032106/encouter-prevented-webview-navigation-when-using-webview-api

As of VS Code 1.31, webviews may only display a page of html content and may not navigate to other resources (such as making a POST request using a form).
You can still make requests using fetch or similar APIs, but standard POST submit form will not work.

pleacu:
This bug also appears relevant:
https://github.com/Microsoft/vscode/issues/48926

Basically the first time I reload the app from the UI it will fail with:  (see console)
*prevented webview navigation*
This occurs when the AtlasMap UI does:
{code}window.location.reload(true);{code}

* After the first time the error appears it will not show up again and all will work well.
* If I run under the VSCode debugger all will work well all of the time.

May be related to this line in VSCode:
https://github.com/Microsoft/vscode/blob/82d97b4c3c9c0c42f3005d4fc9114885ca652094/src/vs/workbench/parts/webview/electron-browser/webview-pre.js#L365

200 ms is way too tight.

apupier:
> May be related to this line in VSCode:
> https://github.com/Microsoft/vscode/blob/82d97b4c3c9c0c42f3005d4fc9114885ca652094/src/vs/workbench/parts/webview/electron-browser/webview-pre.js#L365
> 
> 200 ms is way too tight.

it is way too tight for current AtlasMap time load? is it possible to have a first load with "loading state" and then update when everything is ready? Will it work?
Or it is not possible with a reasonable amount of time and then we should ask a feature request to have the possibility to configure this load timeout? In this case, which timeout would be reasonable? It will depend on which parameters? size of the data transformation? depends on the type of data transformation?

apupier:
> This appears to be the issue:
> https://stackoverflow.com/questions/54032106/encouter-prevented-webview-navigation-when-using-webview-api
> 
> As of VS Code 1.31, webviews may only display a page of html content and may not navigate to other resources (such as making a POST request using a form).
[...]

PR #760, opened 2019-02-13
fix: VSCode support - correct missing mapping lines - restructure set…

pleacu:
…Timeout w/ async/await

Fixes: #748


pleacu:
@igarashitm - not too bad - basically changed window.location.reload() to initialize() and changed the setTimeout() to async/await when importing an ADM file.

Issue #536, opened 2018-08-02
Introduce @AtlasFieldActionParameter


igarashitm:
Allow custom field action to define more restriction on its parameters via an annotation, like length or precision. That should eventually get into `ActionDetail` so UI can consume.

igarashitm:
Addressed by #877

PR #764, opened 2019-02-15
fix: Introduce @AtlasFieldActionParameter

jpav:
Fixes: #536


pleacu:
This is more in @igarashitm 's area than it is mine.

igarashitm:
right, reviewer should be runtime team, not UI

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java
```diff
@@ -156,35 +146,72 @@ public void setMethodInstance(Method m) {
...
-                if (params == null) {
-                    params = new ActionParameters();
+        for (java.lang.reflect.Field field : actionClazz.getDeclaredFields()) {
+            AtlasFieldActionParameter anno = field.getAnnotation(AtlasFieldActionParameter.class);
+            if (anno == null) {
+                continue;
+            }
+
+            if (field.getType() != String.class) {
+                LOG.debug("Field action parameters must be declared of type String;"
```
igarashitm: We should support other atlas primitive types but not only String. Even when `type` is not defined in annotation, we should autodetect from `java.lang.Field.getType()`.

lib/core/src/main/java/io/atlasmap/v2/StoreOutput.java
```diff
@@ -0,0 +1,44 @@
...
+public class StoreOutput extends Action implements AtlasFieldAction {
+
+    private static final long serialVersionUID = 1L;
+
+    @AtlasFieldActionParameter()
+    private String variable;
+
+    @AtlasFieldActionInfo(name = "StoreOutput", sourceType = FieldType.ANY, targetType = FieldType.ANY, sourceCollectionType = CollectionType.ALL, targetCollectionType = CollectionType.ALL)
+    public static Object execute(StoreOutput action, Object input) {
+        return input;
```
igarashitm: This looks incomplete, can you remove this and file a separate issue for adding `StoreOutput`?

igarashitm:
Added wip label as we won't merge features until it stabilizes and decide to branch off for the Fuse 7.3 line.

jpav:
Right, I wasn't thinking correctly - should have set runtime as reviewer. I will also follow the guidelines you set for commit messages.

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java
jpav: This is in preparation for variable support. Since the intention is for all parameters to support variables, they will all need to define any parameters as type String in the class to allow for references to the variables (e.g., "$<variableName>$"), while setting the 'type' in the annotation to the actual type to be stored in the output (In the case of my StoreOutput transformation, it just happens to be the default String type anyways). Since this is a new annotation and therefore is currently unused outside of my example, it won't affect any of the current code or custom actions, so I included the check here.

lib/core/src/main/java/io/atlasmap/v2/StoreOutput.java
jpav: I can, but I was using it here because this leads towards the variable support I'm ultimately providing. If you like, I can remove it though and just provide unit tests alone with a test transformation for this.

jpav:
> Added wip label as we won't merge features until it stabilizes and decide to branch off for the Fuse 7.3 line.

These changes are meant to be mergeable with master since they should be introducing no breaking changes.

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java
igarashitm: Do not request to merge your personal preparation to the upstream master. This is really disrespectful to the project.

It's ugly if user have to declare all the custom action parameters as a String just to accept variables. Variable resolution just have to happen before it's set onto action class.
Also it makes more sense to replace current xsd based action parameters with this annotation completely, so OOTB ones also have to be addressed. It doesn't make sense to be specialized to custom action.

lib/core/src/main/java/io/atlasmap/v2/StoreOutput.java
igarashitm: Again, keep your personal preparation in your private branch

igarashitm:
As you should already know, we already entered bug fix sprint to focus on stabilizing  for productization. We have no spare resource to maintain 2 branches earlier.

chirino:
How’s this shaping up?  Is it being cleaned up so that it can be merged?

jpav:
@chirino I've been reworking this to deal with Tomo's concerns, but I've ended up in a rabbit hole dealing with JAXB issues with recognizing a non-generated field action referenced in an XML mapping. I'm working with Tomo to resolve that now.

jpav:
Thanks, @igarashitm, for removing all of the field action generation code. That made all of this much easier to implement, and the end result is much simpler and elegant. This update makes use of the @AtlasFieldActionParameter annotation to tie action parameters with action execution. Custom actions now support parameters, are simpler to handle, and are serialized in a simpler manner. The CustomAction model class was no longer needed. All built-in actions also have been refactored to use the new parameter annotation. Static class for action execution methods are also gone; these method reside in the action itself. Finally, to avoid circular dependencies, I had to move the io.atlasmap.v2 model classes to the atlas-api module since many of that module's interfaces and SPI classes depend on them.

igarashitm:
> Finally, to avoid circular dependencies, I had to move the io.atlasmap.v2 model classes to the atlas-api module since many of that module's interfaces and SPI classes depend on them.

Can you explain the details why that big refactoring decision needs to be made? It's expected that API refers to model objects, but the other way never should happen.

jpav:
Yes, I needed the actions to implement AtlasFieldAction so that they could contain the execution method along with their parameters. This also made it consistent with how custom actions would be created by users. It also didn't seem to make much sense that classes in the Model module were directly exposed by the API module (along with SPI classes), but weren't part of the API. Granted that now that all of the actual 'model' classes have been moved to the API module, the Model module could be more aptly named the 'Action' module. But having everything about the actions fully contained with the action, both parameters and execution methods, makes everything much simpler and allows for greater flexibility for future enhancements. It seemed like an odd disconnect to begin with, which I believe was mainly a consequence of the action classes being generated.

igarashitm:
It doesn't make sense for me. Models have to be isolated from API. Models are contract between UI and backend, which needs to be isolated and kept simple as much as possible.

jpav:
Not sure why you think the models are a contract between the UI and backend.  Our UI accesses our runtime via a REST API, so has no direct contact with any of our modules anyways other than the REST service module. Plus, I have not added one iota of complexity by refactoring these classes; on the contrary, their purpose is more obvious this way. The whole point of an API is to expose the public parts of the interface that outside users of the application are meant to access, and likewise the SPI the parts exposed for developer extensions to the application. Regardless of whether the code sat in the model module or the API module, all of this code has been and still is exposed as API. This refactoring simplifies that concept AND makes our implementation and use of field actions much more intuitive and flexible for further enhancements. For instance, with the action fully encapsulating everything about itself, it would now be possible to move the logic in the front-end UI, that's determining whether an action is applicable, to the runtime where it really should reside. The UI shouldn't be responsible for that type of business logic. What certainly didn't make sense was having actions that half acted as an interfaces, but also as implementation for parameters, with separate static code elsewhere that actually 'performed' the action. This wouldn't seem normal to any framework developer, but again, I believe this was a consequence of the auto-generated action code implemented by the original developers.

lib/api/src/main/java/io/atlasmap/v2/ActionParameter.java
```diff
@@ -19,124 +19,124 @@
...
 
     /**
      * Sets the value of the fieldType property.
-     * 
+     *
      * @param value
      *     allowed object is
      *     {@link FieldType }
-     *     
+     *
```
chirino: These white space changes make this already big PR even bigger.  I'd recommend reviewing your change set next time and dropping these types of changes.

lib/core/src/main/java/io/atlasmap/core/AtlasMappingService.java
```diff
@@ -38,7 +35,7 @@
 public class AtlasMappingService implements Serializable {
 
     private static final long serialVersionUID = 1668362984516180517L;
-    private static final Logger LOG = LoggerFactory.getLogger(AtlasMappingService.class);
```
chirino: Again these types of cleanup just make your change set larger.  Perhaps gather these in a subsequent patch.

lib/core/src/main/java/io/atlasmap/core/AtlasUtil.java
```diff
@@ -104,7 +105,7 @@ protected static void validateUri(String atlasUri) {
             pass1 = atlasUri.split("\\?", 2);
         }
 
-        List<String> parts = new ArrayList<String>();
```
chirino: here too.

lib/model/src/main/java/io/atlasmap/v2/Append.java
```diff
@@ -1,34 +1,50 @@
 package io.atlasmap.v2;
```
jpav: This is an example of what the Append action looked like before this PR. It contains only its parameter definition and accessor/mutator.

lib/core/src/main/java/io/atlasmap/actions/StringComplexFieldActions.java
```diff
@@ -1,362 +0,0 @@
...
-import io.atlasmap.v2.SubStringAfter;
-import io.atlasmap.v2.SubStringBefore;
-
-public class StringComplexFieldActions implements AtlasFieldAction {
-
-    public static final String STRING_SEPARATOR_REGEX = "^\\s+:_+=";
-    public static final Pattern STRING_SEPARATOR_PATTERN = Pattern.compile(STRING_SEPARATOR_REGEX);
-
-    @AtlasFieldActionInfo(name = "Append", sourceType = FieldType.STRING, targetType = FieldType.STRING, sourceCollectionType = CollectionType.NONE, targetCollectionType = CollectionType.NONE)
-    public static String append(Action action, Object input) {
```
jpav: This is where the Append action's execution method exists before this PR, amongst all the other action methods that are related to strings and that have parameters.

lib/core/src/main/java/io/atlasmap/core/AtlasUtil.java
jpav: I left this one because there are several other changes to this same class.

jpav:
@chirino I think I removed all of the extraneous changes. The whitespace changes in files that had other real changes are somewhat unavoidable since the editor settings are intentionally set to remove trailing whitespace. Most of the files affected by this I believe were generated code.

chirino:
I agree that the model classes should just be data transfer objects dedicated to encoding/decoding the interface documents.  If we want to modularize, and make the the actions more pluggable there are other ways of doing that.

Issue #755, opened 2019-02-11
Support specifying ADM file path via Java system property at backend startup


igarashitm:
Support specifying ADM file path via Java system property like `-Datlasmap.adm.path` or whatever at backend startup so that the client like VSCode can provide a ADM file and have it preloaded in AtlasMap UI.

Only for initial loading purpose. User need to export explicitly to get adm file updated.

lhein:
that sounds like a great idea

apupier:
it will preload all files provided in this folder? And how to open the UI with a specific file? There is a specific uri parameter to provide?

lhein:
"so that the client like VSCode can provide a ADM file and have it preloaded"

apupier:
started some work in this branch: https://github.com/apupier/atlasmap/tree/755-providepropertyToInitAtlasMapWithAdmFile

apupier:
> started some work in this branch: https://github.com/apupier/atlasmap/tree/755-providepropertyToInitAtlasMapWithAdmFile

I started a bit of work with some tests that might be reusable but it is far from working.
I was expecting to have the DefaultAtlasContextFactory.createContext(URI) supporting having an .adm file passed to it. In fact it seems not the case and the upload of .adm file from the UI has some splitted work between UI and server (or I have not found the correct entrypoint)
I'm wondering if I started to look at the right place then.
@igarashitm @pleacu Have i understood well? Do ou think that we (Fuse Tooling dev) can implement the functionality in a reasonable time? Can you provide some guidance please?

igarashitm:
It looks this is not trivial to achieve, need some refactoring around startup/working-directory/adm-file handling. Would be better to leave it to @pleacu

pleacu:
I'm tracking down an odd timing issue - I can look at this after that... :)
         --Paul

PR #793, opened 2019-02-25
fix: Add command line ADM target support:

pleacu:
java -Datlasmap.adm.path=/home/pleacu/Downloads/rat1.adm -jar atlasmap-standalone.jar

Fixes: #755


lib/core/src/main/java/io/atlasmap/core/AtlasUtil.java
```diff
@@ -461,4 +464,34 @@ else if (element.isDirectory())
         }
         return;
     }
+
+    public static void copyFile(Path sourcePath, Path destPath) throws IOException {
```
apupier: what about using Files.copy(Path, Path, CopyOption...)?
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#copy-java.nio.file.Path-java.nio.file.Path-java.nio.file.CopyOption...-
I think it is doing the same thing, isn't it?

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -113,6 +119,42 @@ public void onUpdate(AtlasLibraryLoader loader) {
         });
     }
 
+    /**
+     * The user has specified an ADM file on inception.  Clear current mappings and process
```
apupier: exra space between "inception." and "Clear"

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -113,6 +119,42 @@ public void onUpdate(AtlasLibraryLoader loader) {
...
+                Paths.get(mappingFolderPath.toAbsolutePath().toString() + File.separator + atlasmapCatalogName));
+
+            extractCompressedCatalog();
+
+        } catch (Exception e) {
+            LOG.error("Error processing the AtlasMap catalog file " + atlasmapAdmPath + "\n" +
+                e.getMessage(), e);
+            throw new WebApplicationException(e.getMessage(), e, Status.INTERNAL_SERVER_ERROR);
+        }
+        System.setProperty(ATLASMAP_ADM_PATH, "");
```
apupier: why are you setting the value to empty value? it means that after launch and first use the property provided by the user is overridden? Why not keeping it to know on which transformation it was initially launched?

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -214,13 +260,7 @@ public Response resetMappings() {
...
 
         try {
-            for (File mappingFile : mappings) {
-                if (mappingFile.exists()) {
-                    if (!mappingFile.delete()) {
-                        LOG.warn("Unable to delete mapping file " + mappingFile.toString());
-                    }
-                }
-            }
+            AtlasUtil.deleteDirectoryContents(mappingFolderPath.toFile());
```
apupier: what about using Files.deleteIfExists(Path) or Files.delete(Path)?

lib/core/src/main/java/io/atlasmap/core/AtlasUtil.java
pleacu: I was having trouble with StandardCopyOption.REPLACE_EXISTING - I can try it again.

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
pleacu: ok

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
pleacu: The user can import a totally different ADM file with the import icon.  Keeping ATLASMAP_ADM_PATH pointing to the original file causes it to be reread even of the user imported a different one from the UI.

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
pleacu: I'll look at Files.delete()

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
apupier: ok

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
pleacu: decided to keep my AtlasUtil.deleteDirectoryContents() as it keeps the parent but removes all subcontent

Issue #817, opened 2019-03-07
Correct AtlasMapping datasource URI for XSD document types


pleacu:
Also correct UI test resources ADM and Java file URI detection.

PR #818, opened 2019-03-07
fix: Correct AtlasMapping datasource URI for XSD document types and c…

pleacu:
…orrect JAVA detection

Fixes: #817 
This corrects the error:
2019-03-05 10:14:26.035 ERROR 28221 --- [  XNIO-2 task-1] io.atlasmap.core.DefaultAtlasContext     : Cannot find module info for the DataSource uri 'atlas:xsd:fhir-rafal-170'

It's odd that the DocumentType.JAVA enum was 'Java' instead of 'JAVA'.  Also added some defensive code in the RT service (pulled a NPE at one point).


ui/src/app/lib/atlasmap-data-mapper/models/config.model.ts
```diff
@@ -197,9 +197,12 @@ export class ConfigModel {
     docDef.inspectionSource = docInitModel.inspectionSource;
     docDef.inspectionResult = docInitModel.inspectionResult;
     docDef.selectedRoot = docInitModel.selectedRoot;
-    docDef.uri = 'atlas:' + docDef.type.toLowerCase() + ':' + docDef.id;
-    if (docDef.type === DocumentType.JAVA) {
+    if (docDef.type === DocumentType.XSD) {
+      docDef.uri = 'atlas:' + 'xml' + ':' + docDef.id;
+    } else if (docDef.type === DocumentType.JAVA) {
       docDef.uri += '?className=' + docDef.inspectionSource;
```
igarashitm: isn't it going to miss `atlas:java:id` prefix?

ui/src/app/lib/atlasmap-data-mapper/common/config.types.ts
```diff
@@ -15,7 +15,7 @@
 */
 
 export enum DocumentType {
-  JAVA = 'Java',
+  JAVA = 'JAVA',
   JAVA_ARCHIVE = 'JAR',
```
igarashitm: is it possible JAVA_ARCHIVE to be used for uri, causes `atlas:jar'?

pleacu: Is  atlas:jar:id valid?

igarashitm: nope, it should be `atlas:java:id` as it's still supposed to be handled with JavaModule.

pleacu: I'll push the fix for the first comment - if atlas.jar is required we can catch that in 7.4

igarashitm: ah no we don't need it, I just worry if in any case DocumentType.JAVA_ARCHIVE is passed into ConfigModel, and then it causes "atlas:jar:id" uri which is wrong.

ui/src/app/lib/atlasmap-data-mapper/models/config.model.ts
igarashitm: So actually wondering if we should add `atlas:java:id` when `docDef.type === DocumentType.JAVA` **OR** `docDef.type === DocumentType.JAVA_ARCHIVE`. I'm not sure if it's possible at all to get `DocumentType.JAVA_ARCHIVE` passed in here.

pleacu: so - this is what we've got now:
    if (docDef.type === DocumentType.XSD) {
      docDef.uri = 'atlas:' + 'xml' + ':' + docDef.id;
    } else {
      docDef.uri = 'atlas:' + docDef.type.toLowerCase() + ':' + docDef.id;
    }

    if (docDef.type === DocumentType.JAVA) {
      docDef.uri += '?className=' + docDef.inspectionSource;
    }

So - is this what you are suggesting?
    if (docDef.type === DocumentType.XSD) {
      docDef.uri = 'atlas:' + 'xml' + ':' + docDef.id;
    } else {
      docDef.uri = 'atlas:' + docDef.type.toLowerCase() + ':' + docDef.id;
    }

    if (docDef.type === DocumentType.JAVA ||docDef.type === DocumentType.JAVA_ARCHIVE ) {
      docDef.uri += '?className=' + docDef.inspectionSource;
    }

igarashitm: I would suggest this

```java
if (docDef.type === DocumentType.XSD) {
  docDef.uri = 'atlas:' + 'xml' + ':' + docDef.id;
} else if (docDef.type === DocumentType.JAVA_ARCHIVE) {
  docDef.uri = 'atlas:' + 'java' + ':' + docDef.id;
} else {
  docDef.uri = 'atlas:' + docDef.type.toLowerCase() + ':' + docDef.id;
}

if (docDef.type === DocumentType.JAVA ||docDef.type === DocumentType.JAVA_ARCHIVE ) {
  docDef.uri += '?className=' + docDef.inspectionSource;
}
```

might be even better splitting the enum into two, module type (java, json, xml) and others (schema, instance, etc) but maybe next time

Issue #823, opened 2019-03-11
Docs: Update remainder to 7.3 ADM functionality


pleacu:
Update export and custom field action

PR #824, opened 2019-03-11
fix: Update user guide for export and custom field action.

pleacu:
Fixes: #823 

@igarashitm @TovaCohen @jpav - if you have a sec plz take a quick look


igarashitm:
Oops it's already merged - filed a separate issue #825, thanks @TovaCohen

PR #840, opened 2019-03-26
ENTESB-10411 Merge prod changes back to dev branch

cunningt:
@igarashitm : I'm creating two profiles here - one for normal use (noproxy) and a productization profile that allows us to operate behind the PNC proxy.      

There's one more change here too - I've think I've changed bower-installer to use the binary rather than github - I had issues building atlasmap in prod with github location, but if I changed to the binary, it worked fine.   Can you check that that looks okay?   I don't know angular real well so I don't know if what I did was the right way to do it, but it did seem to work.

### DIFF
--- a/ui/pom.xml
+++ b/ui/pom.xml
@@ -66,64 +66,6 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>com.github.eirslett</groupId>
-        <artifactId>frontend-maven-plugin</artifactId>
-          <executions>
-            <execution>
-              <id>install yarn</id>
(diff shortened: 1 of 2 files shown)


cunningt:
If these look okay, it'd be awesome if we could cherry-pick them to master as well so that we get them for 7.4.

igarashitm:
I think this is fine, one thing I worry is that the yarn.lock is auto-generated from yarn command, so it'll just be overwritten in every build I guess. Even release might not pass due to that changed file.

Issue #860, opened 2019-04-04
Handle FieldGroup in Validation


igarashitm:
It's ignored ATM

Issue #844, opened 2019-03-29
Display a validation error when user combine multiple collection fields


claudio4j:
When user maps a collection to a single element, displays a validation error.

![datamapper](https://user-images.githubusercontent.com/836543/55201259-f47d9980-51a0-11e9-969f-62b2396c5d04.png)

christophd:
Also I noticed that the target document is a collection in that case, too. Although we have a single element specification as target. 

For instance if you have a Json array collection schema as source and want to map to a single element Java type you end up having a singleton list with the Java object in it.

I would expect AtlasMap to map the first element in the source collection to the Java object without any wrapping list type. This seems to be only the case for Java type target documents.

@igarashitm should I open a separate issue for that?

igarashitm:
@christophd Yes please, actual payload example would help - I don't completely understand, if it's `JSON collection field` to `Java non-collection field` mapping, default behavior should be to pick up last element in the collection. `itemAt` transformation could be used to pick up first or something else. There could be a pitfall with separate/conbine case though.

igarashitm:
This one specifically needs to get an validation error that we don't support combining collection fields. It would need to wait for #244 and #843 to be fixed beforehand.

PR #862, opened 2019-04-09
fix: Fix validation for FieldGroup and collection field selection

igarashitm:
Fixes: #844
Fixes: #860


igarashitm:
#864 supersedes this one

PR #877, opened 2019-04-18
Simplifying Action Serialization and Processing.

chirino:


lib/model/src/main/resources/META-INF/services/io.atlasmap.v2.Action
```diff
@@ -0,0 +1,61 @@
+io.atlasmap.v2.AbsoluteValue
+io.atlasmap.v2.Add
+io.atlasmap.v2.AddDays
+io.atlasmap.v2.AddSeconds
+io.atlasmap.v2.Append
```
chirino: Custom actions models can now just extend Action and list that class in their own 'META-INF/services/io.atlasmap.v2.Action' file.  Service discovery should add it and Jackson should be able to deserialize it.

lib/model/src/test/java/io/atlasmap/v2/JsonMarshallerTest.java
```diff
@@ -137,19 +138,40 @@ public void testFieldActions() throws Exception {
...
+        String content = "{\"io.atlasmap.v2.PropertyField\" : {\"jsonType\" : \"io.atlasmap.v2.PropertyField\",\"actions\" : [ {" +
+            "\"Split\" : { \"delimiter\": \"x\"}" +
+        "}]}}";
+        PropertyField f = mapper.readValue(content, PropertyField.class);
+//        System.out.println(mapper.writeValueAsString(f));
+
+        content = "{\"io.atlasmap.v2.PropertyField\" : {\"jsonType\" : \"io.atlasmap.v2.PropertyField\",\"actions\" : [ {" +
+            "\"Trim\" : null" +
+        "}]}}";
+        f = mapper.readValue(content, PropertyField.class);
```
chirino: We still support deserializing actions using the old  `{ "<actionName>": { ... config .... } }` syntax.

lib/model/src/test/java/io/atlasmap/v2/JsonMarshallerTest.java
```diff
@@ -137,19 +138,40 @@ public void testFieldActions() throws Exception {
...
+//        System.out.println(mapper.writeValueAsString(f));
+
+        content = "{\"io.atlasmap.v2.PropertyField\" : {\"jsonType\" : \"io.atlasmap.v2.PropertyField\",\"actions\" : [ {" +
+            "\"@type\" : \"Trim\"" +
+        "}]}}";
```
chirino: But by default we serialize and using the more compact and simpler to understand `{"@type":"<actionName>", ... config... }` form.  I hope one day we can drop the old style.

lib/core/src/main/java/io/atlasmap/core/BaseAtlasModule.java
```diff
@@ -134,15 +134,15 @@ protected void processLookupField(AtlasInternalSession session, LookupTable look
     protected Field applyTargetFieldActions(AtlasInternalSession session) throws AtlasException {
         Field field = session.head().getTargetField();
         if (isAutomaticallyProcessOutputFieldActions() && field.getActions() != null
-                && field.getActions().getActions() != null) {
+                && field.getActions() != null) {
```
chirino: Removing the Actions class simplified the field usage.

lib/core/src/main/java/io/atlasmap/actions/StringComplexFieldActions.java
```diff
@@ -198,19 +199,16 @@ public static String padStringLeft(Action action, String input) {
...
-    public static String prepend(Action action, Object input) {
-        if (!(action instanceof Prepend)) {
-            throw new IllegalArgumentException("Action must be a Prepend action");
-        }
-        String string = ((Prepend) action).getString();
-        if (input == null && string == null) {
-            return null;
+    @AtlasActionProcessor
+    public static String prepend(Prepend action, String input) {
+        String string = action.getString();
```
chirino: This is the one action handler that I've ported over to using the `@AtlasActionProcessor` style, instead of the @AtlasFieldActionInfo style.  Notice that it uses the message signature to discover what Action it should be applied to and the type of input and output fields that it operates on.

Issue #775, opened 2019-02-19
Allow to configure where "target" folder is created


apupier:
it will ensure that the folder is written in a safe place.

see https://github.com/jboss-fuse/vscode-atlasmap/issues/85

igarashitm:
e.g. `-Datlasmap.working.dir`

PR #914, opened 2019-05-03
fix: Add atlasmap.adm.workspace to specify a workspace for a standalo…

pleacu:
…ne atlasmap.

Fixes: #775

To test:
java -Datlasmap.adm.workspace=/home/pleacu/temp/adm-workspace -jar atlasmap-standalone.jar


pleacu:
The argument for specifying a cat file is atlasmap.adm.path so I went with atlasmap.adm.workspace to stay similar.

igarashitm:
Yep, but that's because it's for a path to the adm file, while workspace is not only for adm file.

igarashitm:
So I would say `.adm` is a special file extension for AtlasMap mapping definition archive, and `atlasmap.adm.path` is to specify a path to the specific `.adm` file. Just because working directory contains more than `.adm` file - including expanded temporary file, for me it looks better to avoid it in the name.

Issue #961, opened 2019-06-05
cond-expr: Selecting a new mapping with cond expr enabled show old


pleacu:
Selecting a new mapping when the cond expr is enabled for the old mapping still shows old cexpression.

igarashitm:
I don't see this issue, there might be some condition to reproduce

pleacu:
Procedure:
  1.  Establish 2 different mappings
  2.  Select mapping-A, select the conditional expression icon
  3.  Enter an arbitrary expression
  4.  Select mapping-B, select the conditional expression icon
  5.  Select mapping-A - the conditional expression retains the expression for mapping-B

Workaround:
  Close the expression box and re-open it.

PR #962, opened 2019-06-05
fix: Disable the conditional expression box when selecting a differen…

pleacu:
…t mapping.

Fixes: #961

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
@@ -558,7 +558,12 @@ export class MappingManagementService {
 
       if (mappingsForField && mappingsForField.length === 1) {
         mapping = mappingsForField[0];
-        this.cfg.mappings.activeMapping = mapping;
+
+        // Disable the conditional expression box when selecting a different mapping.
+        if (this.cfg.mappings.activeMapping && this.cfg.mappings.activeMapping !== mapping) {
+          this.cfg.mappings.activeMapping.getCurrentFieldMapping().transition.enableExpression = false;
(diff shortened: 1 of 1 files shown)


igarashitm:
Doesn't this disable expression if select other mapping and get back? Let's say mapping-a has expression and mapping-b without expression and mapping-a is activated. Once you select mapping-b and then select mapping-a, expression for mapping-a is going to be disabled?

pleacu:
The expression for mapping-A is retained, you just need to select the f(x) icon to see it.  The issue I was seeing is that the conditional expression for mapping-B was being shown when selecting mapping-A if both were enabled.  By disabling the cond expr box by default if the user establishes a new active mapping and then returns to an older one they have to reclick the f(x) icon

igarashitm:
But once enableExpression is set to `false`, expression is disabled and `Concatenate` is used:
https://github.com/atlasmap/atlasmap/blob/master/ui/src/app/lib/atlasmap-data-mapper/services/mapping-serializer.service.ts#L93

So I think instead we need to refresh the expression when mapping is selected.

igarashitm:
This is weird though, FieldMappingPair.transition.expression should have corresponding Expression object for each mapping. ExpressionComponent might need to subscribe mapping selection event.

pleacu:
True - I thought the simplest way to refresh was to trigger through the transition html by toggling off the enableExpression.  If a user enables the cond expression box are they enabling it for all mappings or for the selected mapping?

igarashitm:
I would say it's only for selected mapping, but once it's enabled for that mapping, it needs to stay enabled so that the expression automatically appear when mapping is selected.

pleacu:
Okay - that does appear more convenient for the user - I can't use my html trigger though - will withdraw this PR.

Issue #928, opened 2019-05-16
[UI] Insert/Delete text into/from where the caret is


igarashitm:
#927 only adds/deletes text/field to/from the end of existing expression, regardless of where the caret was. Handle caret position.

PR #964, opened 2019-06-05
fix: [UI] Insert/Delete text into/from where the caret is (#928)

igarashitm:
Added key input support at caret position. Another commit will follow for Backspace/Delete support.


igarashitm:
This might brake field auto completion - I need @pleacu 's help to align it :-D

Issue #1003, opened 2019-06-13
Allow multiple mappings for source elements


pleacu:

pleacu:
Clearing out everything seems to work but the 'Select mapping' dropdown appears so the user can distinguish which mapping they want:

![sc1](https://user-images.githubusercontent.com/2981421/59462434-bb44e500-8df1-11e9-8639-27c64015f20e.png)

![Screenshot from 2019-06-13 15-32-35](https://user-images.githubusercontent.com/2981421/59462471-ce57b500-8df1-11e9-8085-84e5ab0f0608.png)

igarashitm:
Yep, until we address #635 it's a bit annoying, but from functionality this is better I would say.

Issue #988, opened 2019-06-12
[Conditional mapping] Unable to specify one field twice


mmelko:
I find it limiting that I can't specify one field twice. i.e:
```if(ISEMPTY(a), b, a}``` or ```if(a < 20, b, a * 2 )```

If you think it's not a bug but rather enhancement feel free to change the label, but from my POV it's a bug.

igarashitm:
It's a bug for sure, what happen when you try to add it 2nd time?

mmelko:
nothing... it will leave there empty space. So IFEMPTY will throw an exception that it needs 3 arguments.

PR #1004, opened 2019-06-13
fix: Enable d&d of source elements already mapped to other targets in…

pleacu:
…to a condition expression.

Fixes: #1003

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/components/expression.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/expression.component.ts
@@ -176,23 +176,7 @@ export class ExpressionComponent implements OnInit, OnDestroy, OnChanges {
     if (droppedField === null || currentFieldMapping === null || !droppedField.isSource) {
       return;
     }
-
-    if (droppedField.partOfMapping) {
-
-      // The selected field is part of a different mapping.
-      if (!currentFieldMapping.isFieldMapped(droppedField)) {
-        return;
-      }
-      // TODO handle drop position - for now this appends a field ref to the end of expression
-      const mappedField = currentFieldMapping.getMappedFieldForField(droppedField, true);
-      this.addConditionalExpressionNode(mappedField, null, 0);
-
-    // Pulling an unmapped field into a transition expression evaluation implies a compound selection.
-    } else {
-      // TODO handle drop position - for now this appends a field ref to the
-      // end of expression in FieldMappingPair#updateTransition()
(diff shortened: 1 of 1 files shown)


ui/src/app/lib/atlasmap-data-mapper/components/expression.component.ts
```diff
@@ -176,23 +176,7 @@ export class ExpressionComponent implements OnInit, OnDestroy, OnChanges {
     if (droppedField === null || currentFieldMapping === null || !droppedField.isSource) {
       return;
     }
-
-    if (droppedField.partOfMapping) {
-
-      // The selected field is part of a different mapping.
-      if (!currentFieldMapping.isFieldMapped(droppedField)) {
```
igarashitm: I thought only removing this if clause would be enough

ui/src/app/lib/atlasmap-data-mapper/components/expression.component.ts
```diff
@@ -176,23 +176,7 @@ export class ExpressionComponent implements OnInit, OnDestroy, OnChanges {
...
-      const mappedField = currentFieldMapping.getMappedFieldForField(droppedField, true);
-      this.addConditionalExpressionNode(mappedField, null, 0);
-
-    // Pulling an unmapped field into a transition expression evaluation implies a compound selection.
-    } else {
-      // TODO handle drop position - for now this appends a field ref to the
-      // end of expression in FieldMappingPair#updateTransition()
-      this.configModel.mappingService.fieldSelected(droppedField, true);
-    }
+    this.configModel.mappingService.fieldSelected(droppedField, true);
```
igarashitm: isn't this removing a field from mapping if it's already partOfMapping?

pleacu: Debugging through fieldSelected() - it will only remove the field of the selected source field is part of the active mapping which ar the point that its dropped into the conditional expression its not.

igarashitm: but field reference could be added multiple times into same expression I think? like when writing an expression `IF(ISEMPTY(name), name, '')` and dropping a `name` field when filling 2nd argument?

igarashitm: kinda related to #988

pleacu: Yup - that was the use case I was missing - trying again...

pleacu:
resubmitting

Issue #998, opened 2019-06-13
[Regression][Separate/Split] Invoking type convertor failed when multiple types of targets


mmelko:
When we have mapping such as String -> (Integer, Long etc.. ) following exception is thrown 
```
Caused by: io.atlasmap.api.AtlasException: Errors: [Unexpected exception is thrown while populating target field: Invoking type convertor failed: docId='io.atlasmap.qe.test.TargetMappingTestClass', path='/targetDouble'], 
	at org.apache.camel.component.atlasmap.AtlasEndpoint.onExchange(AtlasEndpoint.java:262)
	at org.apache.camel.impl.ProcessorEndpoint$1.process(ProcessorEndpoint.java:71)
	at org.apache.camel.util.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:61)
	at org.apache.camel.processor.SendProcessor.process(SendProcessor.java:148)
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201)
	at org.apache.camel.component.direct.DirectBlockingProducer.process(DirectBlockingProducer.java:53)
	at org.apache.camel.processor.SharedCamelInternalProcessor.process(SharedCamelInternalProcessor.java:186)
	at org.apache.camel.processor.SharedCamelInternalProcessor.process(SharedCamelInternalProcessor.java:86)
	at org.apache.camel.impl.ProducerCache$1.doInProducer(ProducerCache.java:541)
	at org.apache.camel.impl.ProducerCache$1.doInProducer(ProducerCache.java:506)
	at org.apache.camel.impl.ProducerCache.doInProducer(ProducerCache.java:369)
	at org.apache.camel.impl.ProducerCache.sendExchange(ProducerCache.java:506)
	at org.apache.camel.impl.ProducerCache.send(ProducerCache.java:229)
	at org.apache.camel.impl.DefaultProducerTemplate.send(DefaultProducerTemplate.java:144)
[...]

PR #1010, opened 2019-06-14
chore: Add testcase for #998

igarashitm:


igarashitm:
adding one more testcase, string to [double,string]

Issue #1064, opened 2019-06-30
No label for enum transformation parameters


igarashitm:
![no-label](https://user-images.githubusercontent.com/265462/60403216-70a0c800-9b68-11e9-9b76-d1801490a702.png)

Issue #1063, opened 2019-06-30
[maven-plugin] Add fieldActionOutputDir to output field action metadata into separate dir


igarashitm:
Currently generate-field-actions uses same directory with generate-insoections. Allow it to use separate directory.

Issue #1061, opened 2019-06-28
Reemove {source,target}CollectionType and use oneToMany/manyToOne in field action metadata


igarashitm:
Actually it's only used to indicate whether it's multiplicity transformation or not. Align wording to avoid confusion. No need to be backward compatible as it's internal only metadata and doesn't appear in mapping definition. Custom field action would need to align with, but at this moment custom field action is not available in Syndesis.

Issue #880, opened 2019-04-23
Convert all built in actions to use the simpler @AtlasActionProcessor style introduced in #877


chirino:
Once #878 is done and we verified that the UI can use the action schema to configure actions, we should convert all built in actions to use the @AtlasActionProcessor style introduced in #877 and the  @JsonPropertyDescription and @AtlasActionProperty on the model fields so that we get nice action json schemas for them.

After that we can consider removing any special handling of the built in actions in the UI if the generic form builder based on the action schema is good enough.

PR #1065, opened 2019-06-30
Field action related fixes

igarashitm:
fix: Convert all built in actions to use the simpler @AtlasActionProcessor style introduced in #877
Fixes: #880

fix: Reemove {source,target}CollectionType and use oneToMany/manyToOne in field action metadata
Fixes: #1061

fix: Let UI use new action schema
Fixes: #878

feat: [maven-plugin] Add fieldActionOutputDir to output field action metadata into separate dir
Fixes: #1063

fix: No label for enum transformation parameters
Fixes: #1064


igarashitm:
Oops, forgot to remove CustomAction

igarashitm:
will follow up with #1066

PR #1074, opened 2019-07-02
fix: [UI] Don't show multiplicity transformations where not eligible

### LINKED ISSUE #905: [UI] Don't show multiplicity transformations where not eligible
many-to-one transformation like `Concatenate` and `Expression`, and one-to-many transformation like `Split` should not show up in the regular transformation pull down. They must be applied in specific order, i.e. many-to-one must come first and one-to-many must come last. We already kinda handle it in UI, just have to hide those transformations in the pull down.

~~We might have to consider transitioning to many-to-one if source field is in a collection and target field is not in a collection.~~ -> filed a separate issue #918

jpav:
fixes #905

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/models/field-action.model.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/models/field-action.model.ts
@@ -150,9 +150,7 @@ export class FieldActionDefinition {
       return false;
     } else if (this.multiplicity === Multiplicity.MANY_TO_ONE) {
       // MANY_TO_ONE field action only applies to collection field or FieldGroup
-      if (!selectedSourceField.isInCollection()) {
-        return false;
-      }
+      return false;
     }
 
     // Check for matching types - date.


ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
```diff
@@ -187,20 +187,27 @@ export class MappingManagementService {
    * @param field
    */
   transitionMode(mapping: MappingModel, field: Field): void {
+    this.transitionMode2(mapping, field.isSource());
     if (mapping.transition.mode === TransitionMode.ONE_TO_ONE) {
-      const mappedFields: MappedField[] = mapping.getMappedFields(field.isSource());
-      if (mappedFields.length > 2) {
-        if (field.isSource()) {
+      this.transitionMode2(mapping, !field.isSource());
```
pleacu: Why did you create a transitionMode2 out of transitionMode()??  These should remain as transitionMode().

ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
```diff
@@ -549,7 +556,7 @@ export class MappingManagementService {
   }
 
   private hasMultipleMappings(mapping: MappingModel): boolean {
-    return (mapping.sourceFields.length > 2 || mapping.targetFields.length > 2);
+    return (mapping.sourceFields.length > 1 || mapping.targetFields.length > 1);
```
pleacu: Checking for multiple mappings needs to account for the None field which occupies sourceFields[0] or targetFields[0] if in one-to-manmy or many-to-one.  I other words I think you were okay at '2'

jpav: The problem is when there are arrays or collections involved, the none field isn't always present, so I have to account for that.

ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
jpav: To deal with arrays and collections, I needed to analyze both the source and target fields in the mapping, so needed to call the method for both sides.

pleacu: transitionMode is really only trying to move from map-mode to either many to one or one to many.  Field.isSource() tells you which way - I don't see why calling it twice to run on all of the sources and targets is needed. Not clear how this was necessary for your issue.

ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
pleacu: Need to test the case where a user was in a multiple-mode and then removed their multiplicity to re-establish a onetoone mapping (ctrl-select on one of the previously establishes mappings).  This would trigger at 2 in the case

igarashitm: I also think this is unrelated to #905, please remove this change.

ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
igarashitm: I also think this is unrelated to #905, please remove this change.

PR #1138, opened 2019-07-24
chore: Added basics to user-doc

igarashitm:
Hi @pleacu, @TovaCohen,

I'm adding some description about mapping basics into user-doc, would you take a look?


igarashitm:
Thanks @pleacu! fixed them.

Issue #1180, opened 2019-08-15
Padding fields remain if you delete your way back to a one-to-one mapping


pleacu:
1.  Establish a compound mapping
2.  Modify the index to allow for padding fields
3  Delete either the source or target field mapping until it's 1-1
4.  padding fields remain...

PR #1181, opened 2019-08-15
fix: Clear extraneous padding fields on mapped field removal back to …

pleacu:
…1-1.

Fixes: #1180

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/models/mapping.model.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/models/mapping.model.ts
@@ -155,7 +155,7 @@ export class MappingModel {
   }
 
   /**
-   * Add a field to this field mapping pair.
+   * Add the specified field to this field mapping.
    *
    * @param field - field to add to the mapping
    * @param first - if true add the field to the beginning of the array, last otherwise.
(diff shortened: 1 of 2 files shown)


ui/src/app/lib/atlasmap-data-mapper/models/mapping.model.ts
```diff
@@ -343,4 +368,26 @@ export class MappingModel {
...
   }
 
+  /**
+   * Remove any trailing padding fields for the mapped field array.  This occurs when a user moves
+   * a mapped element above the last padding field.
+   *
+   * @param mappedFields
+   * @param trailing - Remove trailing padding fields only
+   */
+  clearExtraPaddingFields(mappedFields: MappedField[], trailing: boolean): void {
```
igarashitm: Is there a reason moving this from MappingManagementService? I prefer having more complex logic than basic CRUD in services rather than model itself.

pleacu: ya - I debated that myself.  The reason it's there is that it's needed by remoceField() in mapping.model.  If I put it into mapping-management.service then I'll introduce a circular dependency between model and mms.

igarashitm: So invoke it from here instead?
https://github.com/atlasmap/atlasmap/pull/1181/files#diff-33a65a8e6d54bd6716a9670a71fe3408L306
my vscode says this is the only place invoking MappingModel.removeField()
Then we can keep MappingModel.removeField() from handling other fields (i.e. padding)

pleacu: Makes sense - move the padding clear method back over to the management service.

Issue #1186, opened 2019-08-21
Prohibit collection field in expression


igarashitm:
We have to rule out collection field in expression until we address #1185 

It already causes an error if at least one source field is selected (collection field cannot be selected with other field). We also have to disable expression even when only one collection field is selected.

PR #1200, opened 2019-09-10
fix: Issue a warning if an expression is attempted on a field express…

pleacu:
…ion.

Fixes: #1186

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/models/mapping.model.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/models/mapping.model.ts
@@ -306,6 +306,16 @@ export class MappingModel {
     return paths;
   }
 
+  hasCollectionField(isSource: boolean): boolean {
+    const fields: Field[] = this.getFields(isSource);
+    for (const field of fields) {
+      if (field.isCollection) {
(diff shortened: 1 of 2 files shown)


ui/src/app/lib/atlasmap-data-mapper/models/mapping.model.ts
```diff
@@ -306,6 +306,16 @@ export class MappingModel {
     return paths;
   }
 
+  hasCollectionField(isSource: boolean): boolean {
+    const fields: Field[] = this.getFields(isSource);
+    for (const field of fields) {
+      if (field.isCollection) {
```
igarashitm: I think we have to check if any of ancesters is in collection as well - and I have seen it in some class already implemented. Let me check...

igarashitm: Yep, it's `Field.isInCollection()` method

pleacu: Ahh - missed that - will fix...

pleacu:
New PR coming.

Issue #1221, opened 2019-09-18
Add "Clear Mappings" in Syndesis


igarashitm:
As an alternative to "Reset All" which is supposed to be disabled with #1220, we want "Clear Mappings" in Syndesis to remove all mappings but preserve documents.

PR #1251, opened 2019-10-01
fix: Add 'Clear Mappings' to disable-navbar mode (non-stand-alone) to…

pleacu:
…olbar menu pull-down.

Fixes: #1221

### DIFF
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
@@ -279,6 +279,23 @@ public class AtlasService {
 
         try {
             AtlasUtil.deleteDirectoryContents(mappingFolderPath.toFile());
+        } catch (Exception e) {
+            throw new WebApplicationException(e.getMessage(), e, Status.INTERNAL_SERVER_ERROR);
+        }
+        return Response.ok().build();
+    }
+
+    @DELETE
+    @Path("/fieldActions/RESET_LIBS")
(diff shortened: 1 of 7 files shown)


lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -279,6 +279,23 @@ public Response resetMappings() {
...
         try {
             AtlasUtil.deleteDirectoryContents(mappingFolderPath.toFile());
+        } catch (Exception e) {
+            throw new WebApplicationException(e.getMessage(), e, Status.INTERNAL_SERVER_ERROR);
+        }
+        return Response.ok().build();
+    }
+
+    @DELETE
+    @Path("/fieldActions/RESET_LIBS")
```
igarashitm: Can we use like `resetLibs` format to follow others?

pleacu: Do you mean put it under the path /mappings/RESET_LIBS instead of /fieldActions/RESET_LIBS?  The lib jars are for custom field actions more than mappings.  I can use /mappings path of you want though

igarashitm: Ah wait that's also the one, because the jar could also contains Java Document class right?

I originally meant about the capital/small letters on `RESET_LIBS`, others are all using `capitalFirstCharWithStartingSmallLetter`, without underscore.

Issue #1275, opened 2019-10-10
Don't validate mappings too early.


pleacu:
We currently notifyMappingUpdated() in fetchMappings(0 in the mapping management service.  The JSON serializer won't add these mappings because the source and target fields aren't established yet - causing an empty mapping file to be written back down to the runtime.  This currently manifests itself when you start up with an existing ADM file or hit the browser reload button (import is okay).

PR #1278, opened 2019-10-10
fix: Move the mapping updated notification up a level to a point wher…

pleacu:
…e mapping processing is done.

Fixes: #1275


ui/src/app/lib/atlasmap-data-mapper/services/initialization.service.ts
```diff
@@ -489,9 +489,10 @@ isSource=${docdef.initModel.isSource}, inspection=${docdef.initModel.inspectionT
       }
 
       this.cfg.mappingService.fetchMappings(mappingFiles, this.cfg.mappings).toPromise()
-        .then((result: boolean) => {
+        .then( async(result: boolean) => {
         this.cfg.initCfg.mappingInitialized = true;
         this.updateStatus();
+        await this.cfg.mappingService.notifyMappingUpdated();
```
igarashitm: Can it be `...notifyMappingUpdated().then(() => resolve(true);)` instead? async-await tend to make more things async...

Issue #1398, opened 2019-10-22
pre-cleanup: reduce/condense common and merge w/ base


pleacu:
One more iteration - pull modal out into its own file, pull out all components from common and merge whats left with base.

PR #1406, opened 2019-10-23
fix: reduce/condense common and merge w/ base css

pleacu:
Fixes: #1398

This should be pretty close.  ANy other tweeks we can do when we create the subdirs.


ui/src/app/lib/atlasmap-data-mapper/components/data-mapper-mapping-list.component.css
```diff
@@ -91,6 +91,16 @@
   color: black;
 }
 
+/* lookup-table */
+.DataMapperUI .LookupTableComponent .lookupTableRow { margin:0px 10px 10px 5px; padding:5px 0px; }
```
igarashitm: IIRC lookup table modal pops up from `Mapping Details` pane? if so maybe better to be in mapping-detail CSS.

igarashitm: Yep, it's in the transition-selection component, which is a sub component of mapping-detail.

Issue #1408, opened 2019-10-23
NPE when XML field reader read namespace


igarashitm:
In some case this happens
```
java.lang.NullPointerException: null
	at io.atlasmap.xml.core.XmlFieldReader.getNamespace(XmlFieldReader.java:258) ~[atlas-xml-core-1.42.x-SNAPSHOT.jar:1.42.x-SNAPSHOT]
	at io.atlasmap.xml.core.XmlFieldReader.extractSegment(XmlFieldReader.java:128) ~[atlas-xml-core-1.42.x-SNAPSHOT.jar:1.42.x-SNAPSHOT]
	at io.atlasmap.xml.core.XmlFieldReader.read(XmlFieldReader.java:107) ~[atlas-xml-core-1.42.x-SNAPSHOT.jar:1.42.x-SNAPSHOT]
...
```

PR #1411, opened 2019-10-24
fix: NPE when XML field reader read namespace

igarashitm:
Fixes: #1408


igarashitm:
heh, pros and cons I think... issue ID doesn't explain the symptom by itself. So maybe only when the condition is a bit too complex to express with a few words.

Issue #1442, opened 2019-10-31
Lines don't update when new fields mapped via mapping-details and expression


pleacu:
Currently the line machine is controlled by the document-definition component.  Since additional mapped fields can be added by the search box in the mapped details section and by the '@' insert node dropdown we must support the ability to trigger the mapping redraw event from outside of document.

Good chance to clean up the line machine a little.  Caught this is the demo prep.

igarashitm:
Yeah that sounds bad. The mapping change event must be centralized and line machine should just subscribe it.

PR #1443, opened 2019-10-31
fix: Corrected line machine support to handle added field mappings fr…

pleacu:
…om no document/ panel sources.

Fixes: #1442

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/components/app/line-machine.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/app/line-machine.component.ts
@@ -60,8 +60,9 @@ export class LineMachineComponent implements OnInit, OnDestroy {
   ngOnInit(): void {
     this.mappingUpdatedSubscription = this.cfg.mappingService.mappingUpdated$.subscribe(() => {
       this.mappingChanged();
-      this.docDefInput.setLineMachine(this);
-      this.docDefOutput.setLineMachine(this);
+      this.docDefInput.setLineMachineEventObserver(this);
+      this.docDefOutput.setLineMachineEventObserver(this);
+      this.cfg.lmcInstance = this;
     });
   }
 
--- a/ui/src/app/lib/atlasmap-data-mapper/components/document/document-definition.component.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/components/document/document-definition.component.ts
@@ -14,7 +14,7 @@
     limitations under the License.
 */
 
-import { Component, Input, ViewChildren, ElementRef, EventEmitter, QueryList, ViewChild, OnInit } from '@angular/core';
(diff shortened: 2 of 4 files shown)


igarashitm:
Could line-machine subscribe `mappingUpdated$` instead?
https://github.com/atlasmap/atlasmap/pull/1443/files#diff-33a65a8e6d54bd6716a9670a71fe3408L46
In this way individual components no longer have to explicitly emit mapping updated event. I think `mappingUpdatedSource` notified via `notifyMappingUpdated()` is already a centralized mapping updated event source.

pleacu:
> Could line-machine subscribe `mappingUpdated$` instead?
> https://github.com/atlasmap/atlasmap/pull/1443/files#diff-33a65a8e6d54bd6716a9670a71fe3408L46
> In this way individual components no longer have to explicitly emit mapping updated event. I think `mappingUpdatedSource` notified via `notifyMappingUpdated()` is already a centralized mapping updated event source.

Actually yes - I think I can get rid of the AdmRedrawMappingLinesEvent entirely.  The real issue is that validateMappings() fails every once in a while with:
`
javax.ws.rs.WebApplicationException: HTTP 400 Bad Request
	at io.atlasmap.service.AtlasService.fromJson(AtlasService.java:790) ~[atlas-service-1.43.0-SNAPSHOT.jar:1.43.0-SNAPSHOT]
	at io.atlasmap.service.AtlasService.validateMappingRequest(AtlasService.java:439) ~[atlas-service-1.43.0-SNAPSHOT.jar:1.43.0-SNAPSHOT]
`

It passes the next time through.  The UI trace dump looks good.  The problem is that when it fails notifyMappingUpdated() ends up not invoking this.mappingUpdatedSource.next() as it should 
https://github.com/atlasmap/atlasmap/blob/master/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts#L661-L663

We trap the error as just a warning in the log.

pleacu:
Hey @igarashitm - there are two different line-machine based events that are being subscribed to.  One associates a line update with a mapping update (mappingUpdatedSource mappingUpdatedSubscription mappingUpdated$) the other associates a line update event with anything else (scroll, search box, etc).  It didn't seem correct to trigger a mapping update line event on a scroll so this latest PR preserves the two.  If we need to distinguish them later we can.

igarashitm:
Ah OK. Then I would create a separate line refresh event source in the mapping-management.service and add `notifyLineRefresh()`. Then `notifyMappingUpdated()` would invoke `notifyLineRefresh()` as well. line-machine.component just subscribe line refresh event source instead of mapping updated. In this way document-definition.component and line-machine.component no longer have to be tightly coupled, reducing bug source around the complicated line refresh events.

pleacu:
new commit coming...

PR #1498, opened 2019-11-20
WIP for 1481

pleacu:
DO NOT PUSH - WIP
Hey @riccardo-forina - pls take a look.
This gets as far as MapperContextToolbar - FilePicker.  I'm calling 

`onChange={(selectedFile: File) => importAtlasFile(selectedFile)}`

which generates the error:
`Cannot invoke an expression whose type lacks a call signature. Type 'PropsWithChildren<IMapperContextToolbarProps>' has no compatible
call signatures.`

MapperContextToolbar defines:

`export interface IMapperContextToolbarProps {
  importAtlasFile: (selectedFile: File) => void;
}`

### DIFF
--- a/ui-react/packages/atlasmap-standalone/src/App.tsx
+++ b/ui-react/packages/atlasmap-standalone/src/App.tsx
@@ -47,6 +47,7 @@ const App: React.FC = () => {
       targets={targetDocs.map(fromDocumentDefinitionToFieldGroup)}
       mappings={[]}
       addToMapping={() => void(0)}
+      importAtlasFile={() => void(0)}
     />
   );
 };
--- a/ui-react/packages/atlasmap-ui/src/mapper/MapperContextToolbar.tsx
+++ b/ui-react/packages/atlasmap-ui/src/mapper/MapperContextToolbar.tsx
@@ -15,13 +15,16 @@ import {
 } from '@patternfly/react-core';
 
 import { FilePicker } from 'react-file-picker';
-import { processImportedFile } from './MapperUtilsToolbar';
+// import { processImportedFile } from './MapperUtilsToolbar';
 
-export interface IMapperContextToolbarProps {}
+
+export interface IMapperContextToolbarProps {
+  importAtlasFile: (selectedFile: File) => void;
+}
 
 export const MapperContextToolbar: FunctionComponent<
   IMapperContextToolbarProps
-> = () => {
+> = (importAtlasFile) => {
   return (
     <Toolbar
       className="view-toolbar pf-u-px-md pf-u-py-md"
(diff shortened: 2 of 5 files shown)


ui-react/packages/atlasmap-standalone/src/App.tsx
```diff
@@ -47,6 +47,7 @@ const App: React.FC = () => {
       targets={targetDocs.map(fromDocumentDefinitionToFieldGroup)}
       mappings={[]}
       addToMapping={() => void(0)}
+      importAtlasFile={() => void(0)}
```
riccardo-forina: You need to get `importAtlasFile` from line 36 and pass it down here

ui-react/packages/atlasmap-ui/src/mapper/MapperContextToolbar.tsx
```diff
@@ -15,13 +15,16 @@ import {
...
-export interface IMapperContextToolbarProps {}
+
+export interface IMapperContextToolbarProps {
+  importAtlasFile: (selectedFile: File) => void;
+}
 
 export const MapperContextToolbar: FunctionComponent<
   IMapperContextToolbarProps
-> = () => {
+> = (importAtlasFile) => {
```
riccardo-forina: This should be `({ importAtlasFile }) => {`, which uses [ES6 object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) to save some keystroke.

The extended version would be like this:

```tsx
export const MapperContextToolbar: FunctionComponent<IMapperContextToolbarProps> = (props) => {
  const importAtlasFile = props.importAtlasFile;
```

The former is the preferred way to do things in react-land.

ui-react/packages/atlasmap-ui/src/mapper/MapperContextToolbar.tsx
```diff
@@ -41,7 +44,8 @@ export const MapperContextToolbar: FunctionComponent<
           >
           <FilePicker
             extensions={['adm', 'jar']}
-            onChange={(selectedFile: File) => processImportedFile( {selectedFile} )}
+            // onChange={(selectedFile: File) => processImportedFile( {selectedFile} )}
+            onChange={(selectedFile: File) => importAtlasFile(selectedFile)}
```
riccardo-forina: `onChange={importAtlasFile}` would be preferable since lambdas can create performance issues.

Check [tslint jsx-no-lambda
](https://github.com/palantir/tslint-react)

> Creating new anonymous functions (with either the function syntax or ES2015 arrow syntax) inside the render call stack works against pure component rendering. When doing an equality check between two lambdas, React will always consider them unequal values and force the component to re-render more often than necessary.

ui-react/packages/atlasmap-ui/src/mapper/MapperUtilsToolbar.ts
```diff
@@ -1,55 +1,12 @@
...
-            level: ErrorLevel.ERROR, scope: ErrorScope.APPLICATION, type: ErrorType.INTERNAL, object: error}));
-        }
-      });
-      */
-  } else if (userFileSuffix === 'JAR') {
-    // this.cfg.documentService.processDocument(selectedFile, InspectionType.JAVA_CLASS, false);
-  }
-}
+  importAtlasFile(selectedFile);
+}
```
riccardo-forina: This whole file isn't needed, there should be no business logic on the ui package.

pleacu:
creating new PR

Issue #1501, opened 2019-11-21
Allow to embed AtlasMap in Eclipse Theia


apupier:
through VS Code extensions, the goal is to embed AtlasMap.

It seems that it requires to use an i-frame inside webviews.

It implies to disable x-frame-options.
What are the implications on a security level? is it safe to do it at AtlasMap standalone level?

references:
- Theia issue https://github.com/eclipse-theia/theia/issues/5007
- VS Code AtlasMap PR https://github.com/jboss-fuse/vscode-atlasmap/pull/239

apupier:
disabling the x-frame-option did the trick on AtlasMap side.

PR #1502, opened 2019-11-21
Disable x-frame-option in standalone application #1501

apupier:
it allows to embed it in Eclipse Theia (requirement to embed in Eclipse
Che)

Is it safe to do it this way? Are there other ways to avoid x-frame-options issues when embedding in an i-frame?

I can also imagine to disable it only when starting the standalone app with specific parameters. Would it make sense?

### DIFF
--- a/standalone/src/main/java/io/atlasmap/standalone/SecurityConfiguration.java
+++ b/standalone/src/main/java/io/atlasmap/standalone/SecurityConfiguration.java
@@ -26,6 +26,8 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
 
     @Override
     protected void configure(HttpSecurity httpSecurity) throws Exception {
-        httpSecurity.csrf().csrfTokenRepository(new AtlasMapXsrfRepository());
+        httpSecurity
+          .headers().frameOptions().disable()
+          .and().csrf().csrfTokenRepository(new AtlasMapXsrfRepository());
     }
 }


standalone/src/main/java/io/atlasmap/standalone/SecurityConfiguration.java
```diff
@@ -26,6 +26,8 @@
 
     @Override
     protected void configure(HttpSecurity httpSecurity) throws Exception {
-        httpSecurity.csrf().csrfTokenRepository(new AtlasMapXsrfRepository());
+        httpSecurity
+          .headers().frameOptions().disable()
```
jonathanchristison: Any reason why ```headers().frameOptions().sameOrigin(); ``` couldn't be used?

akosyakov: Maybe it should be configurable cia cli option? So a case for VS Code extension would be following, for example:
- webview iframe is served from origin like `https://uuid.webview.hostname/`
- nestead iframe for atlasmap is served from another origin like `https://hostname:3000/`

In this case the same origin policy won't allow. Although one try to configure nested iframe with `sandbox="allow-scripts allo-same-origin` attribute. It could work.

apupier: there is a blank page and this error in console browser when using same origin:
`Refused to display 'http://localhost:8586/' in a frame because it set 'X-Frame-Options' to 'sameorigin'.`

I have it also when I try to set `sandbox="allow-scripts allow-same-origin"` on the iframe of the VS Code Webview content:
`<iframe src="${fullWebServerUri}" width="100%" height="900" sandbox="allow-scripts allow-same-origin">`

apupier: maybe there's another way
`Sending the proper Content Security Policy (CSP) frame-ancestors directive response headers that instruct the browser to not allow framing from other domains. (This replaces the older X-Frame-Options HTTP headers.)`

I just have no idea what is CSP and where to configure it https://www.owasp.org/index.php/Clickjacking

akosyakov: I am not sure that `CSP` helps. It seems to be used instead of `X-Frame-Options`. What can it do differently?

It would make sense to set `CSP` in your extension for `webview.scpSource` (webview access) and a result of `asExternalUri`(localhost access). As you actually already do in https://github.com/jboss-fuse/vscode-atlasmap/pull/239

But then you embed atlasmap iframe in your extension which is secured by CSP as above. So I don't see why you need `CSP` or `X-Frame-Options` for atlasmap content.

jonathanchristison: I have to admit I'm not that familiar with internals of VS Code and Che or how its "seeing" the iframe, `X-Frame-Options: deny` is obviously very heavily handed, Thanks @apupier for testing out `sameorigin` I thought this might of been an easy fix but is dependent on the browser (ultimately Chromium for native VSCode) or users browser interpretation of sameorigin.

@akosyakov X-Frame-Options looks to be a bit outdated or else `allow-from` might of been the solution here, at least that would of allowed specific URIs to show the iframe but this has [poor support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#Browser_compatibility)

OWASP have a cheat sheet [here](https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html) which recommends CSP for this functionality of explicitly setting URIs which can display the iframe - 

Would ```Content-Security-Policy: frame-ancestors 'self' *.somesite.com https://myfriend.site.com http://localhost;``` work? (this is where my lack of familiarity of VSCode and Che mean i could be off the mark) I get the feeling adding localhost is probably a bad idea but doesn't defeat the whole purpose of CSP so its better than disabling any protections altogether.

Also is this just a problem likely to occur when testing? Will users be running this on localhost or is it likely to be on a openshift cluster somewhere?

apupier: > Would something like `Content-Security-Policy: frame-ancestors 'self' *.somesite.com https://myfriend.site.com http://localhost;` work? (this is where my lack of familiarity of VSCode and Che mean i could be off the mark) I get the feeling adding localhost is probably a bad idea but doesn't defeat the whole purpose of CSP so its better than disabling any protections altogether.

I tried to configure the CSP witr he frame-ancestors. If I don't disable the x-frame-options, it doesn't work. If I deactivate the x-frame-options, it seems that all requests are working whatever I put in the frame-ancestors but maybe I have not configured the Springboot security context for that.
I will try to provide more details later on what I tried.

> Also is this just a problem likely to occur when testing? Will users be running this on localhost or is it likely to be on a openshift cluster somewhere?

it is for end-uers. We have several use cases. One is to have AtlasMap standlaone launched locally and accessed by the VS Code instance locally.
Another one is having the VS Code extension running in a container and accessed through Theia bridge to VS Code.

jonathanchristison: > I tried to configure the CSP witr he frame-ancestors. If I don't disable the x-frame-options, it doesn't work. If I deactivate the x-frame-options, it seems that all requests are working whatever I put in the frame-ancestors but maybe I have not configured the Springboot security context correctly for that.

X-Frame-Options is [superseded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) by CSP 

> Note: The Content-Security-Policy HTTP header has a frame-ancestors directive which obsoletes this header for supporting browsers.

So if CSP is working correctly disabling X-Frame-Options should be fine, I'd need to look in the web dev console to get a feel for what is going on because without a doubt this can be complicated :smile: 

> it is for end-uers. We have several use cases. One is to have AtlasMap standalone launched locally and accessed by the VS Code instance locally.
> Another one is having the VS Code extension running in a container and accessed through Theia bridge to VS Code.

Ah ok, so the localhost is a required URI, I was just thinking maybe its something we could alter with a launch parameter or something

akosyakov: > Ah ok, so the localhost is a required URI, I was just thinking maybe its something we could alter with a launch parameter or something

If it is possible to alter it would be an option as well. VS Code extension can get a proper external URL (it will be localhost only if it runs locally) and then pass this URL. It also can pass own origin (https://uuid.webview.hostname/) if it helps.

apupier:
provided another PR which is making it configurable to disable the x-frame-options https://github.com/atlasmap/atlasmap/pull/1571

apupier:
we are going with an option to disable the x-frame-option for now. it seems to be enough to move forward.
Thanks all!

PR #1513, opened 2019-11-21
This PR enables the import icon for ADM and JAR. 

pleacu:
@riccardo-forina @igarashitm 

 It doesn't use the latest `state` structure in `useAtlasmap.tsx` but I think we should all examine it to
see if it meets with what we are expecting.  The significant item is line 36 of `useAtlasmap.tsx`:

`const cfg = ConfigModel.getConfig();`

Materializing the static `cfg` object in this support layer allows the downstream methods in service and models to operate as is.  We can use this mechanism to get the UI functionality to work without decomposing the config model (hopefully). Anyway - we can push this after review - it is convenient to be able to import an ADM into the standalone AtlasMap.

ps.  There was an issue with master before this where the standalone sits in a loading loop.
You can import to get out of it.


riccardo-forina:
It look ok to me 👍

PR #1517, opened 2019-11-22
chore(react ui): fix the rendering performance while dragging

riccardo-forina:


ui-react/packages/atlasmap-ui/stories/Mapper.stories.tsx
```diff
@@ -28,6 +29,7 @@ export const sample = () => createElement(() => {
       targets={targets}
       mappings={mappings}
       addToMapping={addToMapping}
+      importAtlasFile={action('importAtlasFile')}
```
igarashitm: Can you encapsulate these function pointers into a class? This ends up with having all the function pointers required for all downstream components and everytime this file have to be edited. Please decouple function pointers from unrelated components.

riccardo-forina: This is a story file for Storybook, it’s specific to the ui package and its development workflow. It has no impacts whatsoever on upstream.

igarashitm: Ah sorry I put it on a wrong file, I meant real components that has this issue

riccardo-forina: Right, I’m not sure I’m following. What’s the problem?

ui-react/packages/atlasmap-provider/src/useAtlasmap.tsx
```diff
@@ -1,19 +1,27 @@
...
+} from './models/error.model';
 import { MappingDefinition } from './models/mapping-definition.model';
 import { DocumentManagementService } from './services/document-management.service';
 import { ErrorHandlerService } from './services/error-handler.service';
 import { FieldActionService } from './services/field-action.service';
 import { FileManagementService } from './services/file-management.service';
 import { InitializationService } from './services/initialization.service';
 import { MappingManagementService } from './services/mapping-management.service';
-import { fromDocumentDefinitionToFieldGroup, fromMappingDefinitionToIMappings } from './utils/to-ui-models-util';
+import {
```
pleacu: I think this is where I have a comment/ question.  Essentially all of the component support logic - i.e. all of the business logic  code from:

https://github.com/atlasmap/atlasmap/tree/master/ui/src/app/lib/atlasmap-data-mapper/components

Needs to end up here.  It's great that it's in atlasmap-provider - but how should this be designed?  For example - importAtlasFile() came from toolbar.component.ts.  Should be use the same naming convention as the old UI and import the resulting files?  For example - create MapperUtilsToolbar.ts here in atlasmap-provider to correspond with toolbar.component.ts?

wdyt?

riccardo-forina: All the logic that's presentation specific will have to be implemented in the UI package. All the code that's about providing data to the components will have to be exposed by the provider. Where to put things is up to us, if you think organizing helper code like `importAtlasFile` in something that reminds of the UI component that will use it go for it. I personally don't feel the need for it.

pleacu: Ok - I think that would help organize everything to establish a similar structure - i.e. atlasmap-provider and atlasmap-data-mapper/components. We can always reduce later if it seems over-complex.

riccardo-forina: Sure, works for me

ui-react/packages/atlasmap-ui/stories/Mapper.stories.tsx
igarashitm: e.g. this one, seems like every components has same problem, the number of parameters will eventually exponentially explode
https://github.com/atlasmap/atlasmap/blob/master/ui-react/packages/atlasmap-ui/src/mapper/Mapper.tsx#L22-L28

riccardo-forina: That's the `Mapper`'s interface, it's the data and callbacks that component needs to work in a self-contained way.

riccardo-forina: It's not a problem, it's what's needed to achieve a decoupling of the UI elements from the underlying backend. This will allow us to implement things as the ones @chirino was suggesting in the chat a while ago.

igarashitm: I don't get it why it's not a problem, why we have to carry all function pointers  without any wrapping all the way down to the terminal components? It's fatally worse than carrying ConfigModel.

riccardo-forina: Function pointers? It's a Typescript interface, it's the type definition of the properties required by the component to work. It's a compile time thing.

igarashitm: I mean individual functions are more internal implementation details than ConfigModel as an interface to the AtlasMap main body (i.e. ui/models and ui/services)

pleacu: There's really two parts to this - there's the complexity of the atlasmap-provider which I'll address by splitting things out in a way similar to what we have in atlasmap-data-mapper/components.  Then there's the complexity of the UI <-> atlasmap-provider interface which is a reflection of the added support components.  Can I suggest that for now we keep the interface simple and add to it all of the functions that are required.  It should be obvious after it's all there and we can consider how to establish new classes to organize the interface (likely along the lines established by atlasmap-provider).  We can keep cfg as an interface object from atlasmap-provider to the systems and models packages.

Issue #930, opened 2019-05-16
Support multiple mappings accessed through route


AbbassFaytaroony:
In real case scenarios, especially when using atlasmap standalone, it is more practical to have multiple "Atlasmapping.json" for working on different transformations simultaneously without re-running the app. The different AtlasMappings can be accessed through route, example :
`url:port/id/{{atlas_mapping_id}}` 

Afaik the application currently does not support this functionality right?

igarashitm:
At this moment multiple mapping support with single instance is not on the plan. As a workaround, you can run multiple AtlasMap standalone instance by specifying different port
`java -jar atlasmap-standalone.jar --server.port=10000`

AbbassFaytaroony:
Hello @igarashitm . Is it open for pull requests, or it is out of the scope of the application?

igarashitm:
Contribution is always welcome for sure! Are you thinking about supporting it from backend side and open multiple browsers? Having UI side support for multiple mappings might conflict with current user like Syndesis.

AbbassFaytaroony:
Good to hear that! yes, i am thinking about implementing it from the backend. The only part in the UI is consuming the id from the url ( **If set**, leaving default functionality exactly as it is now), then using the id to refer to AtlasMapping in CRUD apis. In the backend each mapping will have its set of files as if running multiple instances of the app ( .adm, .gz, etc ). Again, all with defaults if no explicit id is set in the route.

igarashitm:
Sounds great, if you could come up with PoC code, team and I can join you discussing about it :+1:

AbbassFaytaroony:
Hello again, i already made a progress in this before which i had to postpone it. Soon i will push the POC to its branch for your review.

AbbassFaytaroony:
@igarashitm Hello again, i pushed a working PoC on the branch issue#930-poc . You can find short notes about testing the feature [here](https://github.com/AbbassFaytaroony/atlasmap/blob/issue%23930-poc/Issue%23930%20-PoC-notes.md)

igarashitm:
Sounds good, right now busy with fixing bugs though, I'll try jumping on it soon.

igarashitm:
I'm sorry it took longer - I took a slight look, that idea of using Angular route is interesting. However I wonder if it helps for following features, like showing a list of available mappinngs in UI and choose one from them.
[...]

Issue #1476, opened 2019-11-15
Refactor .adm file structure


igarashitm:
1. Put catalog JSON file without gzip
1. Externalize all schemas into individual files. The catalog JSON file would then have a file path instead of contents
1. Take DFDL support into account. It requires additional properties to be fed on both of design time **and runtime**.
1. Investigate if we still need to craft gzipped catalog file. It was added mainly for reducing the number of REST invocation for performance. But I believe UI doesn't always require all the schemas when it reads/writes a catalog file. It would be even better to migrate schema catalog into mapping definition. Once schema contents are splitted into individual files, the catalog should be small enough to sit in mapping definition JSON file.

Proposed directory structure
```

- ${atlasmap.workspace}/0
| - atlasmapping-UI.0.json
| - data-source-catalog.json
| - data-sources
    | - json
       | - {$DocumentID}.json
    | - xml
       | - ${DocumentID}.xsd
| - inspected
    | - java
        | - ${DocumentID}.json
    | - json
        | - ${DocumentID}.json
    | - xml
        | - ${DocumentID}.json
```

igarashitm:
#579/#2494 would be also affected

igarashitm:
@rkorytkowski one question, do you need any change in .adm exported archive structure? I guess not, .adm only holds raw schemas and inspection results are generated when it's imported through UI. So even after on-demand inspection is implemented, chunked inspection results are only temporarily stored in backend local directory and does not affect exported .adm archive structure.

PR #1525, opened 2019-11-23
Issue #930 - Support Multiple Mappings Accessed through Route

AbbassFaytaroony:
- Application is accessed through route : `{hostname}:{port}/mapping-id/{mappingId}` where mappingId is a number, if not set: `{hostname}:{port}/`, default is 0
- All mapping manipulation actions modify the current mapping according to mappingId. For that, mappingId is added as a path param to all REST api modifying mappings.
- Mapping files are saved in {mappingFolder}/{mappingId}, ex: 
    - target/mappings/0/adm-catalog-files-0.gz
    - target/mappings/0/atlasmap-catalog-0.adm
    - target/mappings/0/atlasmapping-UI.0.default.json
- Reset mappings deletes mapping files in the subdirectory where mappingId indicates.
- Exporting mappings exports current mappings ( according to current mappingId )
- Importing ADM file overwrites the current mappings ( according to current mappingId ). Exported mappings from older version are compatible, imported and renamed according to current mappingId.

One concern to be mentioned:
- Lib directory is still common for all mappings, so if any mapping was reset, this will clear the /lib, affecting other mappings.


igarashitm:
Thanks, the description all looks good. I'll look into code.
@pleacu would you also look at this? Let's triple check that this is compatible with current usecase so we can merge it.

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -78,10 +82,15 @@
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 
+import static io.atlasmap.v2.MappingFileType.ADM;
```
pleacu: I'm a little confused with MappingFileType.ADM.  Not trying to be too pedantic but this stuff can be confusing :).
The "mapping file" is a JSON file that's kept in the mappings directory.  It is already decorated with an ID (i.e.
atlasmapping-UI.950612.json).  The ADM and GZ files are "catalog files" that contain the source documents that are 
being mapped as well as a pre-digested version of the mappings. The ADM also contains user-defined JARs for
custom field mappings and custom transformations.  The atlasmapping-UI.950612.json exists seperately as an efficiency so when we auto update the mappings it doesn't take too long.

Given this - I like the idea of storing the mapping and catalog data hierarchically - which does mean that the backend
server and UI must be paired - I just find the MappingFileType.ADM confusing.

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -261,30 +271,47 @@ public Response removeMappingRequest(@ApiParam("Mapping ID") @PathParam("mapping
     }
 
     @DELETE
-    @Path("/mapping/RESET")
+    @Path("/mapping/RESET/{mappingId}")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = "Remove All Mappings", notes = "Remove all mapping files saved on the server")
```
pleacu: Need to change the @ApiOperation to indicate you're removing the specified mappingId - not all mappings.  Which
reminds me - is there a mechanism for remving all mappings with on backend REST invocation?  Needed for 'Reset All'
in the UI.

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -261,30 +271,47 @@ public Response removeMappingRequest(@ApiParam("Mapping ID") @PathParam("mapping
...
 
+        List <String> filesToDelete = getMappingFileNames(mappingId);
+        mappings = Arrays.stream(mappings)
+            .filter(file -> filesToDelete.contains(file.getName()))
+            .toArray(File[]::new);
+
         try {
-            AtlasUtil.deleteDirectoryContents(mappingFolderPath.toFile());
+
+            for (File mappingFile: mappings) {
```
pleacu: Ah - there it is... :)

ui/src/app/lib/atlasmap-data-mapper/utils/mapping-serializer.spec.ts
```diff
@@ -34,7 +34,8 @@ describe('MappingSerializer', () => {
     });
     cfg = ConfigModel.getConfig();
     cfg.clearDocs();
-    cfg.mappings = new MappingDefinition();
+    cfg.mappings = new MappingDefinition(0);
+    cfg.mappingId = 0;
```
pleacu: How does mappingId (config.model) correspond to MappingDefinition.activeMapping??

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
igarashitm: Discussed with Paul and tried to find better naming for this enum, but couldn't. We might have to revisit this as a part of #1476. Let's just leave it as-is for now.

Issue#930 -PoC-notes.md
```diff
@@ -0,0 +1,27 @@
+# Issue#930 PoC, Support Multiple Mappings Accessed Through Routes
```
igarashitm: Can you migrate this Multiple Mapping Definition access usage into docs/src/main/asciidoc/user-guide/?

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -78,10 +82,15 @@
...
 
+import static io.atlasmap.v2.MappingFileType.ADM;
+import static io.atlasmap.v2.MappingFileType.GZ;
+
 @Api
 @Path("/")
 public class AtlasService {
 
+    static final String MAPPING_NAME_PREFIX = "UI.";
+    static final String MAPPING_NAME_SUFFIX = ".default";
```
igarashitm: Do we need this `.default` suffix?

ui/src/app/lib/atlasmap-data-mapper/utils/mapping-serializer.spec.ts
igarashitm: Maybe the name of `mappingId` is confusing, @AbbassFaytaroony @pleacu WDYT about using `mappingDefinitionId` instead? yeah it's long, but less confusing with individual mapping.

igarashitm:
Before the merge, I'd like to ask you also to clean up the commits/commit logs a bit, just to make it trackable later.
https://github.com/atlasmap/atlasmap/wiki/AtlasMap-Developer-Memo

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
AbbassFaytaroony: > Need to change the @ApiOperation to indicate you're removing the specified mappingId - not all mappings

Yes definitely. 

> is there a mechanism for remving all mappings with on backend REST invocation? Needed for 'Reset All'
in the UI.

When i was thinking about this i thought about when might the user require to reset all mappings and the current only, but yes it can be provided at least as back-end functionality even if not used in the front-end.

Issue#930 -PoC-notes.md
AbbassFaytaroony: Sure!

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
AbbassFaytaroony: Initially i used it to distinguish in mappings/?filter API between UI.1 and UI.11 (where all mapping files were in the same directory) without changing the API. It can be removed now

AbbassFaytaroony:
@igarashitm Yes, definitely!

ui/src/app/lib/atlasmap-data-mapper/utils/mapping-serializer.spec.ts
AbbassFaytaroony: Yes, we are actually Identifying mapping definition as in front-end terminology. Still afraid however it may cause confusion in the backend where the same entity is called `AtlasMapping`. But yes, `mappingDefinitionId` will be overall more accurate.

Issue#930 -PoC-notes.md
AbbassFaytaroony: @igarashitm This reminds me with a question. I intended not to commit changes in generated swagger API docs, but the thing making me unsure is that they are not under `.gitignore` . Shall i commit them?

igarashitm: Oh yes please, they're auto-generated but still to be committed. As you changed REST API it would change the swagger doc as well.

AbbassFaytaroony:
@igarashitm all done

Issue#930 -PoC-notes.md
igarashitm: So now it's safe to remove this file?

AbbassFaytaroony: Yes it is. Actually i thought i did. I will delete it

AbbassFaytaroony: done

igarashitm:
@AbbassFaytaroony looks good! let's merge it - can you squash intermediate commits other than the ones you want to explicitly make it separate?

AbbassFaytaroony:
@igarashitm already done, reduced commits number from ~30 to 20, but if the log is still long for the feature i can squash more. I am ready both ways :)

igarashitm:
Maybe couple a few more, to avoid changes in later commit negates the earlier commit? for example last commit for deleting .md file can be squashed into adding usage in .adoc, as well as the first commit which added the .md? I prefer commiting only for the conclusion, WDYT?

AbbassFaytaroony:
Absolutely. You're right.

igarashitm:
nah, I would say it's preference ;)

AbbassFaytaroony:
@igarashitm Now it is better

AbbassFaytaroony:
My pleasure !! Willing to go further with this project, so please assign whenever needed.

igarashitm:
Awesome! first let's celebrate this achievement together
https://groups.google.com/forum/#!topic/atlasmap/VhBQNXNxPmk

igarashitm:
Hmm not sure why this fails only after merge, PR build was successful. @AbbassFaytaroony do you see what's an issue?
https://circleci.com/gh/atlasmap/atlasmap/4414

AbbassFaytaroony:
That's weird... i should have rebased the branch on the very latest master before asking to merge the PR.

PR #1537, opened 2019-11-26
chore(ui-react): heavy refactoring of the canvas view

riccardo-forina:
This paves the way for future views


ui-react/packages/atlasmap-provider/src/services/initialization.service.ts
```diff
@@ -220,7 +220,7 @@ isSource=${docdef.initModel.isSource}, inspection=${docdef.initModel.inspectionT
 
       // Fetch adm-catalog-files.gz if it exists.
       this.cfg.fileService.getCurrentMappingCatalog().subscribe( async(catalog: Uint8Array) => {
-
+        console.log("ZOMG", catalog)
```
pleacu: should probably not include this file

riccardo-forina: Thanks

PR #1541, opened 2019-11-27
fix: Add glue layer support for export icon.

pleacu:
Partial fix for i1523

Needs react modal/ dialog support for export file name.


ui-react/packages/atlasmap-provider/src/useAtlasmap.tsx
```diff
@@ -138,6 +138,13 @@ export function useAtlasmap({
     };
   }, [initializationService]);
 
+  const handleExportAtlasFile = useCallback(
+    (file: File) => {
+      exportAtlasFile(file);
+    },
+    [dispatch]
```
riccardo-forina: We don't need a `useCallback` in this case since the `exportAtlasFile` callback doesn't depend on anything created in this component.

PR #1549, opened 2019-11-29
feat(react-ui): searchable fields

riccardo-forina:
This introduces even more refactoring to allow the consumer of the views to deal with the events fired by the UI components without "magic" wrapper components.

Fields are now filterable, although the result seems buggy? @pleacu I'll need your eyes here.

Demo:

Kapture 2019-11-29 at 16 08 13


riccardo-forina:
CC @dongniwang you might be interested in the gif

PR #1560, opened 2019-12-04
Porting Fixes #1486 Enable support for mapping from nested to non-nes…

### LINKED ISSUE #1486: Enable support for mapping from nested to non-nested collection
#780 implemented backend changes to support this case. See testAsymmetricSingleTarget() at https://github.com/atlasmap/atlasmap/blob/master/lib/itests/core/src/test/java/io/atlasmap/itests/core/NestedCollectionJsonTest.java#L70 

Similar tests need to be added to NestedCollectionXmlTest and NestedCollectionJavaTest and it needs to be enabled in the UI at https://github.com/rkorytkowski/atlasmap/blob/cf7d463fe6e0e8782affc686c690a508143d69f9/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts#L399-L405

rkorytkowski:
…ted collection ff6cb1d89ae

### DIFF
--- a/ui-react/packages/atlasmap-provider/test/mapping-management.service.spec.ts
+++ b/ui-react/packages/atlasmap-provider/test/mapping-management.service.spec.ts
@@ -5,6 +5,7 @@ import { LoggerModule, NGXLogger, NgxLoggerLevel } from 'ngx-logger';
 import { MappingManagementService } from '../src/services/mapping-management.service';
 import { ErrorHandlerService } from '../src/services/error-handler.service';
 import { Field } from '../src/models/field.model';
+import {DocumentDefinition, MappingDefinition, MappingModel} from "../src";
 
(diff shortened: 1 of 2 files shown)


rkorytkowski:
@riccardo-forina that's all what I've changed that wasn't there in React.

ui-react/packages/atlasmap-provider/test/mapping-management.service.spec.ts
```diff
@@ -29,4 +30,104 @@ describe('MappingManagementService', () => {
       expect(service.getFieldSelectionExclusionReason(null, f.parentField)).toContain('parent');
     }),
   );
+
+    it('should reject if source and target have different collection counts and target has more than one collection',
```
rkorytkowski: I've copied the tests, even though we are not yet able to run them...

PR #1561, opened 2019-12-04
feat(react-ui): confirmation modals

riccardo-forina:
This adds a `useConfirmationDialog` hook to deal with confirmations in the app. 

It's a hook in the application and not a UI component since modals need to be handled by the host application - when embedded in Syndesis for example there will be Syndesis's way to deal with modals. The same decoupling will have to be implemented when we'll add notifications support.

Kapture 2019-12-04 at 19 07 09


riccardo-forina:
Note that the delete document dialog isn't actually deleting anything, but the logic to deal with the delete is there (see the useRef and how it's used in the onConfirm callback).

PR #1563, opened 2019-12-05
Partial fix for i1555

pleacu:
@riccardo-forina - This implements document delete in source/target panels.  There is an issue that
I need your help with.  I'm dispatching the handleDeleteAtlasFile through the useCallback as 'reset'
(It seemed the closest to what I needed).  The result is the UI still shows the deleted file (if you refresh
the window it correctly disappears).   Is there a new context state that's needed (basically refresh the UI but not reinit)??


pleacu:
will rebase and resubmit - I've got it working now.

PR #1565, opened 2019-12-05
fix: wire glue layer for document delete

pleacu:
Fixes: #1555

Addressed previous pull issues.

### DIFF
--- a/ui-react/packages/atlasmap-provider/src/AtlasmapProvider.tsx
+++ b/ui-react/packages/atlasmap-provider/src/AtlasmapProvider.tsx
@@ -187,7 +187,7 @@ export function useAtlasmap({
   sources: IAtlasmapDocument[];
   targets: IAtlasmapDocument[];
   mappings: IMappings[];
-  deleteAtlasFile: (fileName: any, isSource: any) => void;
+  deleteAtlasFile: (fileName?: string, isSource?: boolean) => void;
   exportAtlasFile: () => void;
   importAtlasFile: (file: File, isSource: boolean) => void;
   resetAtlasmap: () => void;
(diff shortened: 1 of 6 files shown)


ui-react/packages/atlasmap-provider/src/AtlasmapProvider.tsx
```diff
@@ -186,6 +187,7 @@ export function useAtlasmap({
   sources: IAtlasmapDocument[];
   targets: IAtlasmapDocument[];
   mappings: IMappings[];
+  deleteAtlasFile: (fileName: any, isSource: any) => void;
```
riccardo-forina: String and bool?

pleacu: The UI side returns a Mutable string | undefined which doesn't match string.  I didn't want to proliferate that mutable type into the provider side.

riccardo-forina: I’d still explicitly set the types and either wrap the calling functions in an if to check for undefined or use the ! operator to suppress the warning

pleacu: hmm - how does the ! operator work exactly?? :)

riccardo-forina: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

pleacu: Added new PR

riccardo-forina: Erh... I was suggesting the other way round, checking that current is not null and making the type required in the delete function

pleacu: Doh!  I'll give it another shot...

PR #1573, opened 2019-12-06
WIP - Work in Progress - Do NOT PUSH

pleacu:
@riccardo-forina - this is odd - I get the TextInput widget but when I type into it the typing is not echoed.  The handleTextInputChange callback is being called once for each character.  There's not a lot on Stackoverflow or other sites involving patternfly 4 so I'm a little at a loss.  Any quick opinion would be cheerfully accepted... :)

### DIFF
--- a/ui-react/packages/atlasmap-standalone/src/useSingleInputDialog.tsx
+++ b/ui-react/packages/atlasmap-standalone/src/useSingleInputDialog.tsx
@@ -1,0 +1,63 @@
+import { Modal, Button, TextInput } from '@patternfly/react-core';
+import React, { ReactChild, ReactPortal, useCallback, useState } from 'react';
+import { createPortal } from 'react-dom';
+
+export type Callback = (closeDialog: () => void) => void;
+
+export interface IUseSingleInputDialogArgs {
+  title: string;
+  content: ReactChild;
+  placeholder: string;
+  state: {
+    inputValue: ''
+  };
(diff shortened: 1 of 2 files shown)


ui-react/packages/atlasmap-standalone/src/useSingleInputDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const openModal = () => setIsOpen(true);
+  const closeModal = () => setIsOpen(false);
+  const handleTextInputChange = (value: string) => {
+    console.log('>>>>>>>>>>>>>>>>>>>>>> handleTextInputChange - value is ' + value);
+    // setState({ value });
+  };
+
+  const handleConfirm = useCallback(() => onConfirm(closeModal), [onConfirm]);
+  const handleCancel = useCallback(() => onCancel(closeModal), [onCancel]);
+  const { inputValue } = state;
```
riccardo-forina: State in React isn't like in Angular, you have to explicitly manage it. So you want to tell React that you need some state:

`const [value, setValue] = useState('');`

and then you want your `TextInput` to use `value` as the value and to trigger `setValue` when handling changes.

Eventually, you'll want to forward `value` to the `onConfirm` callback, so that the caller can deal with it.

Make sure to go through this https://reactjs.org/docs/forms.html

pleacu:
Have a new PR in the queue

Issue #1523, opened 2019-11-22
UI_REACT: Enable export


pleacu:
Enable the export icon.

PR #1580, opened 2019-12-09
fix: complete dialog portion of export catalog (ADM)

pleacu:
Fixes: #1523

### DIFF
--- a/ui-react/packages/atlasmap-provider/src/components/toolbar/toolbar-util.ts
+++ b/ui-react/packages/atlasmap-provider/src/components/toolbar/toolbar-util.ts
@@ -22,29 +22,44 @@ export async function deleteAtlasFile(fileName: string, isSource: boolean) {
   cfg.initializationService.updateInitComplete();
 }
 
+/**
+ * Correct the AtlasMap mappings catalog file name by giving it the correct
+ * suffix if needed.
+ *
+ * @param fileName - user-specified catalog file name
+ */
+ function handleMappingsInstanceName(fileName: string): string | undefined {
+
+  if (fileName !== null && fileName.length > 0) {
+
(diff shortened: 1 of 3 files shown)


ui-react/packages/atlasmap-provider/src/components/toolbar/toolbar-util.ts
```diff
@@ -22,29 +22,44 @@ export async function deleteAtlasFile(fileName: string, isSource: boolean) {
...
 
+/**
+ * Correct the AtlasMap mappings catalog file name by giving it the correct
+ * suffix if needed.
+ *
+ * @param fileName - user-specified catalog file name
+ */
+ function handleMappingsInstanceName(fileName: string): string | undefined {
+
+  if (fileName !== null && fileName.length > 0) {
```
riccardo-forina: File name can’t be null, so neither the return of this function can. This should simplify the code somewhere - I saw an ! that can be removed

ui-react/packages/atlasmap-standalone/src/App.tsx
```diff
@@ -38,6 +39,24 @@ const App: React.FC = () => {
     [importAtlasFile]
   );
 
+  const [exportDialog, openExportDialog] = useSingleInputDialog({
+    title: 'Export Mappings and Documents.',
+    content: 'Please enter a name for your exported catalog file',
+    placeholder: 'atlasmap-mapping.adm',
+    onConfirm: (closeDialog, value) => {
+      closeDialog();
+      if (value === undefined) {
```
riccardo-forina: I think this callback should guarantee not to be fired in case of no value. This could be achieved by setting the text input as required and do some validation on the minimum length. Doing this should streamline the code.

pleacu:
new PR coming

Issue #1592, opened 2019-12-12
UI_REACT: add placeholder back into export dialog


pleacu:
The way it is now - you have to back over the atlasmap-mapping.adm in order to enter the file name you want.  The placeholder is ghosted and will disappear as soon as you start typing.
![sc1](https://user-images.githubusercontent.com/2981421/70739198-4db43b80-1ce4-11ea-8841-761f0a5a13f9.png)

PR #1596, opened 2019-12-12
fix: add placeholder support for export

pleacu:
Fixes: #1592

This matches the user experience on the angular side.


pleacu:
wdyt?

PR #1599, opened 2019-12-13
chore(react-ui): refactor code to improve DX

riccardo-forina:
~Also automatically formatted code with prettier for consistency (`yarn run format`).~ (I'll make another PR for this)


riccardo-forina:
Basically this code here

https://github.com/atlasmap/atlasmap/blob/3f9b80366ffb0310f3df2b0bd8bc741e3a6e7547/ui-react/packages/atlasmap-standalone/src/App.tsx#L117-L138

is a shortcut to this

https://github.com/atlasmap/atlasmap/blob/3f9b80366ffb0310f3df2b0bd8bc741e3a6e7547/ui-react/packages/atlasmap-ui/src/AtlasmapUI/AtlasmapUI.tsx#L32-L50

right now the differences are not many, but when the MappingDetail components will be fleshed out and the TableView component added, the advantage will be more evident.

pleacu:
Okay - so is this the final commit for review or is there another coming?

riccardo-forina:
Final one. The one I removed is the automatic formatting done with prettier, that doesn't need a review really but since it will touch _a lot_ of files it's better kept separated.

riccardo-forina:
(actually let me change a little thing I just noticed 😅)

riccardo-forina:
Done, now the comment I made before makes more sense.

Issue #1512, opened 2019-11-21
Support dynamic path expression resolution


igarashitm:
Prerequisites for #1507 

We would have to pull expression processing up to the `DefaultAtlasContext` so that it dynamically read fields with passed in field path expression.
https://github.com/atlasmap/atlasmap/blob/master/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java#L569
And then enhance it to process dynamic field resolution in expression.

PR #1610, opened 2019-12-17
Fixes #890 Support invoking arbitrary transformation in expression

### LINKED ISSUE #890: Support invoking arbitrary transformation in expression
Add field action function to invoke arbitrary field action in conditional mapping expression.

I would start from `DefaultAtlasFunctionResolver#resolve()`, if no OOTB function is found for given name, then delegate to transformation. Maybe create a `FieldActionFunction` as a wrapper which lookup field action. We also have to investigate if we need any trick for user developped custom transformation that would be uploaded via UI.

rkorytkowski:
https://github.com/atlasmap/atlasmap/issues/890

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java
+++ b/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java
@@ -1,6 +1,5 @@
 package io.atlasmap.core;
 
-import java.lang.reflect.Array;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Parameter;
@@ -9,7 +8,6 @@ import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -57,9 +55,22 @@ public class DefaultAtlasFieldActionService implements AtlasFieldActionService {
     private ReadWriteLock actionProcessorsLock = new ReentrantReadWriteLock();
     private AtlasConversionService conversionService = null;
     private ActionResolver actionResolver = null;
+    private static DefaultAtlasFieldActionService instance;
 
-    public DefaultAtlasFieldActionService(AtlasConversionService conversionService) {
+    private DefaultAtlasFieldActionService(AtlasConversionService conversionService) {
         this.conversionService = conversionService;
+    }
+
+    public static DefaultAtlasFieldActionService getInstance() {
+        if (instance == null) {
+            synchronized (DefaultAtlasFieldActionService.class) {
+                if (instance == null) {
+                    instance = new DefaultAtlasFieldActionService(DefaultAtlasConversionService.getInstance());
+                    instance.init();
+                }
+            }
+        }
+        return instance;
     }
 
     public void init() {
@@ -357,7 +368,7 @@ public class DefaultAtlasFieldActionService implements AtlasFieldActionService {
                     Type itemType = method.getGenericParameterTypes()[1];
                     Class<?> itemClass = paramType.isArray() ? paramType.getComponentType()
                         : (Class<?>)((ParameterizedType) itemType).getActualTypeArguments()[0];
-                    
+
                     if (sourceCollectionType != CollectionType.NONE) {
                         if (sourceCollectionType == CollectionType.ARRAY) {
                             sourceList = Arrays.asList(sourceObject);
@@ -372,7 +383,7 @@ public class DefaultAtlasFieldActionService implements AtlasFieldActionService {
                         sourceList = Arrays.asList(sourceObject);
                     }
                     for (int i=0; i<sourceList.size(); i++) {
-                        
+
                         Object item = sourceList.get(i);
                         if (item != null) {
                             item = conversionService.convertType(item, null, itemClass, null);
@@ -532,8 +543,8 @@ public class DefaultAtlasFieldActionService implements AtlasFieldActionService {
         try {
             readLock.lock();
             for (ActionProcessor processor : actionProcessors) {
-                ActionDetail detail = processor.getActionDetail();
                 if (customAction != null) {
+                    ActionDetail detail = processor.getActionDetail();
                     if (customAction.getClassName().equals(detail.getClassName())
                         && customAction.getMethodName().equals(detail.getMethod())) {
                         matches.add(processor);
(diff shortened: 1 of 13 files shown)


lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFunctionResolver.java
```diff
@@ -44,17 +49,51 @@ private void init() {
         for (FunctionFactory f : implementations) {
             functions.put(f.getName().toUpperCase(), f);
         }
+
+        fieldActionService = new DefaultAtlasFieldActionService(DefaultAtlasConversionService.getInstance());
+        fieldActionService.init();
```
rkorytkowski: Is there any reason why DefaultAtlasFieldActionService is not a singleton like e.g. DefaultAtlasFunctionResolver or DefaultAtlasConversionService...?

lib/core/src/test/java/io/atlasmap/actions/ExpressionFieldActionTest.java
```diff
@@ -189,4 +189,31 @@ public void testNumberField() throws Exception {
...
+    public void testConcatenateActionWithDelimiter() throws Exception {
+        Expression action = new Expression();
+        action.setExpression("CONCATENATE{delimiter=${0}}(${1}, ${2}, ${3})");
+        assertEquals("a,b,c", ExpressionFieldAction.process(action, Arrays.asList(",", "a", "b", "c")));
+    }
+
+    @Test
+    public void testConcatenateActionWithWrongProperty() throws Exception {
+        Expression action = new Expression();
+        action.setExpression("CONCATENATE{delimiter=${0},fail='true'}(${1}, ${2}, ${3})");
```
rkorytkowski: Please note the format I picked for providing action parameters...

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFunctionResolver.java
igarashitm: Nothing I'm aware of, seems better to be a singleton

lib/core/src/test/java/io/atlasmap/actions/ExpressionFieldActionTest.java
igarashitm: Wow, it's great job that you found a way to assemble this into current code.
I was thinking about squeezing those properties as function arguments. And I understood that is not possible because the action detail is not unique with just a name, but it has to also look at the value type (for overloading).
What do you think, instead, separating field action id from display name so that field action ID could be always unique? If you can find a unique action detail with an ID, then you can pass function arguments to `ActionParameters` as much as it's declared, then we can assume that rest are the values to process. Once we get syntax assist in the UI expression input, then we can provide `ActionParameters` metadata as a part of argument description. We might also have to take care of the order of `ActionParameters`.
I think we can defer [1] as I don't think there're only a few if there is which has conflicting name. I hope the order [2] is always identical on initialization, but have to investigate.
1. Make field action id unique and separate from name
1. Take care of `ActionParameters` order for function usage

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java
```diff
@@ -547,20 +545,55 @@ public ActionProcessor findActionProcessor(Action action, FieldType sourceType)
...
+                if (valueType.equals(processor.getActionDetail().getSourceType())) {
+                    return processor;
+                }
+            }
+        }
+        return processors.get(0);
+    }
+
+    public Object processAction(ActionProcessor actionProcessor, Map<String, Object> properties, Object value) {
+        FieldType valueType = (value != null ? getConversionService().fieldTypeFromClass(value.getClass()) : FieldType.NONE);
```
rkorytkowski: Here I determine the type of value as some actions have the same name, but accept different types.

lib/core/src/test/java/io/atlasmap/actions/ExpressionFieldActionTest.java
rkorytkowski: Regarding [1] I'm not sure, if I follow you. In the current implementation I do take care of identifying the value type at https://github.com/atlasmap/atlasmap/pull/1610/files#r359283060 for the purpose of selecting the right action if there are multiple actions with the same name. 

Regarding the format [2]. Do we want to make it CONCATENATE(delimiter=${0}, fail='true', ${1}, ${2}, ${3}) i.e. use named arguments as in e.g. Python? I wouldn't simply assume that first arguments are for ActionParameters as they may be optional e.g. delimiter can be omitted, because it's "" by default.

igarashitm: I'd like to also ask @chirino's opinion here

Ah named arguments sound better IMO. But still I wonder which of this named arguments and disallow optional property is better - if we can auto-fill default property values in the UI as a part of expression syntax assist, I think we can make all action properties mandatory, thus just the order of arguments can identify them.

rkorytkowski: Okay, let's wait for @chirino to chime in. I've fixed the code to use named arguments.

igarashitm: As a side note, thinking about the syntax assist as well as a tooltip we would show for each function in UI, relying on the order of arguments would make it simpler.

rkorytkowski: Just to clarify do you think that expecting action parameters as first arguments i.e. CONCATENATE(${0}, ${1}, ${2}, ${3}), where ${0} is a delimiter, is simpler than using the named arguments i.e. CONCATENATE(delimiter=${0}, ${1}, ${2}, ${3})?

Do we have the syntax assist in place already? I see us being able to support both ways at some point (just as Python does) and think that named arguments are more readable until the syntax assist is in place. Also relying on the order becomes less readable when more arguments are expected and some with defaults, but I agree that with the syntax assist it is equally usable.

igarashitm: Yes, actually some of OOTB action has more than one action parameters, but I assumed that is already identical, as `ActionDetail` holds `ActionParameters`. Is that true? So if order is fixed, runtime should be able to internally identify which is action parameter(s) and the rest are field values.
IMO named argument is still a bit tricky for citizen integrator=non-programmer, and that somehow breaks the sales copy "Spreadsheet like simple math formula". Then when we want to show a tooltip help for the functions, if argument order could be changed with named argument, that is a bit more confusing.
Unfortunately we don't have syntax assist yet, actually after introducing dynamic path expression (#1512, field ref is changed to like ${docId:/path/to/field}) we'd have to rebuild expression UI.
So as far as argument order is fixed, introducing named arguments for the temporary solution until we deliver syntax support, could be an option I think. We can keep named argument support, but just let UI to fill those without name.

igarashitm: Just noticed that, even before we support syntax assist, we can describe those functions in manual. And I think fixed order arguments makes it a bit easier to read, as user don't have to find the argument name in the manual and type it in UI.

rkorytkowski: Okay, let me do the fixed order arguments. I've noticed that argument names are not consistently displayed in the current UI under transformations so indeed one would have to look in a manual to figure them out.

igarashitm:
LGTM thanks!
Let's merge it - can you squash the commits to leave only the conclusions?

rkorytkowski:
Squashed!

Issue #1606, opened 2019-12-16
UI_REACT: Wire onAddToMapping to atlasmap provider


pleacu:
Provide the ability to map new fields.

PR #1614, opened 2019-12-18
fix: add support for adding additional mappings for selected dropped …

pleacu:
…fields

Fixes: #1606

### DIFF
--- a/ui-react/packages/atlasmap-provider/src/components/field/field-util.ts
+++ b/ui-react/packages/atlasmap-provider/src/components/field/field-util.ts
@@ -1,0 +1,44 @@
+import { ConfigModel } from '../../models/config.model';
+import { getDocDef } from '../document/document-util';
+import { Field } from '../../models/field.model';
+
+/**
+ * Add the specified field to the current mapping.
+ *
+ * @param fieldId - field identifier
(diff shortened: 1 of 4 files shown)


ui-react/packages/atlasmap-provider/src/components/field/field-util.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Add the specified field to the current mapping.
+ *
+ * @param fieldId - field identifier
+ * @param mappingId - needed for multiple mapping files
+ */
+export function addToMapping(fieldId: string, _mappingId: string): void {
+  const cfg = ConfigModel.getConfig();
+  const [uri, docType, docName, panel, uuid] = fieldId.split(':');
+  const isSource = (panel === 'source');
```
riccardo-forina: I don’t like inferring info like this, it’s brittle. Let’s keep this in mind when we’ll refactor how we deal with the tree

pleacu: agreed.  Also - the AtlasCanvasViewMappings doesn't establish a new mapping when we drop src to target with no existing mappings.  Not sure how it is supposed to create the new mapping at that point so it can call addToMapping.

PR #1617, opened 2019-12-19
feat(react-ui): add mapping details (partially)

riccardo-forina:
Also renames atlasmap-provider to atlasmap-core. 

The scope of the
package changed slightly; trying to decouple the rxjs store from the ui
elements that require accessing it through property drilling was just
too complicated for no real benefit. 

Now the core package will provide
an already connected to the store `Atlasmap` component that will render
Atlasmap UI as required by the standalone app, leaving the option to
create a custom Atlasmap through a manual wiring of the store to the
desired components.


riccardo-forina:
Added some fine tuning of the column layout calculator.

Issue #1618, opened 2019-12-19
UI_REACT: Add core support for wiring up mapping creation.


pleacu:
Support onCreateMapping(0
ref: https://github.com/atlasmap/atlasmap/blob/c30e736de2ac539c8a29e34f760cc0ea28ea7f23/ui-react/packages/atlasmap-core/src/Atlasmap.tsx#L143

PR #1620, opened 2019-12-19
I1618

pleacu:
fix: add new mapping creation support
Fixes: #1618

### DIFF
--- a/ui-react/packages/atlasmap-core/src/Atlasmap.tsx
+++ b/ui-react/packages/atlasmap-core/src/Atlasmap.tsx
@@ -137,10 +137,7 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
               sources={sources}
             />
 
-            <AtlasmapCanvasViewMappings
-              onAddToMapping={addToMapping}
-              onCreateMapping={createMapping}
-            />
+            <AtlasmapCanvasViewMappings onAddToMapping={addToMapping}/>
 
             <AtlasmapCanvasViewTarget
               onAddToMapping={addToMapping}
--- a/ui-react/packages/atlasmap-core/src/Atlasmap.tsx
+++ b/ui-react/packages/atlasmap-core/src/Atlasmap.tsx
@@ -43,6 +43,7 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
     enableMappingPreview,
     onFieldPreviewChange,
     addToMapping,
+    createMapping,
   } = useAtlasmap();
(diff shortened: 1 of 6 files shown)


ui-react/packages/atlasmap-ui/src/AtlasmapUI/AtlasmapCanvasView/AtlasmapCanvasViewMappings.tsx
```diff
@@ -5,6 +5,7 @@ import { useAtlasmapUI } from '../AtlasmapUIProvider';
 
 export interface IAtlasmapCanvasViewMappingsProps {
   onAddToMapping: (elementId: ElementId, mappingId: string) => void;
+  onCreateMapping: (sourceId: ElementId, targetId: ElementId) => void;
```
riccardo-forina: This isn’t used, is it?

pleacu: Actually no - seemed like reasonable functionality to make available.

riccardo-forina: To create a mapping using a “new mapping” target in the mapping column we’ll need to either allow for creating incomplete mappings or figure out how to delay the mapping creation until we have at least a source and a target.

If incomplete mappings are possible then it’s easy, but we’ll need a different callback since we’ll have a single field to deal with.

Does it make sense?

pleacu: true - new PR on its way

ui-react/packages/atlasmap-core/src/Atlasmap.tsx
```diff
@@ -136,11 +137,14 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
               sources={sources}
             />
 
-            <AtlasmapCanvasViewMappings onAddToMapping={addToMapping} />
+            <AtlasmapCanvasViewMappings
+              onAddToMapping={addToMapping}
+              onCreateMapping={createMapping}
```
riccardo-forina: You’ll have to remove this too I think

pleacu: *ugh* - thought I needed that one - still new at this...

PR #1646, opened 2020-01-07
Fixes #1185 Support collection field in expression

### LINKED ISSUE #1185: Support collection field in expression
With transformation support in expression (#890), collection field should be able to participate in expression.

We would need to relax field type check in UI so that the collection field could be selected with other fields, but only when expression is enabled.

A question, how do we handle when:
1. Enable expression
2. Add one collection source field
3. Add one non-collection source field
4. Disable expression

This falls into error state, "collection field cannot selected with other field" in normal flow (https://docs.atlasmap.io/developer-guide/#mapping-state).

One solution is to wipe out field selection when user disables expression. In the example above, when user push the button to disable expression, show a modal window that has a warning "If you disable expression, source fields you selected are cleared out, are you sure?"
We can also stop filling initial expression when user enables the expression. Not necessary though, I don't think that initial expression just concatenating with "+" makes much sense for users.

Then, we can even enable having an expression with collection target. IOW expression mapping should allow anything other than compound target selection.

rkorytkowski:
https://github.com/atlasmap/atlasmap/issues/1185

I'll include ui-react changes as as follow up once merged.

### DIFF
--- a/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
+++ b/ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
@@ -65,6 +65,8 @@ export class MappingManagementService {
   private mappingPreviewInputSubscription: Subscription;
   private mappingUpdatedSubscription: Subscription;
 
+
+
   constructor(private logger: NGXLogger, private http: HttpClient) {}
 
   get cfg() {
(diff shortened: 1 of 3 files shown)


ui/src/app/lib/atlasmap-data-mapper/services/mapping-management.service.ts
```diff
@@ -694,7 +700,7 @@ export class MappingManagementService {
    * @param position
    * @param offset
    */
-  private updateTransition(mapping: MappingModel, position?: string, offset?: number): void {
+  updateTransition(mapping: MappingModel, position?: string, offset?: number): void {
```
rkorytkowski: For stubbing...

igarashitm: Does adding `<any>` help to get around it?
https://stackoverflow.com/questions/8475564/using-jasmines-spyon-upon-a-private-method

rkorytkowski: It did the trick. Thanks!

PR #1672, opened 2020-01-15
Upgrade to OpenAPI 3.0

zregvart:
In Syndesis I wanted to switch the swagger-core to the newest version. The newest version switches from Swagger 2.0 to OpenAPI 3.0 specification. Having both versions in the same project, one brought over by Atlasmap and a different version used by Syndesis didn't make sense -- that's why I need this update in Atlasmap and then at the point when we upgrade to the version of Atlasmap that includes this we can switch to the same version in Syndesis as well.


igarashitm:
@zregvart thanks a lot for this! do we need this for syndesis 1.9.x? It affects if we branch atlasmap-1.43.x off before merge this.

zregvart:
@igarashitm we need this for 2.0 currently on `master`, I still have to do the same upgrade at Syndesis, hopefully today. The plain is to create a draft pull request and then when a version of Atlasmap is released containing this we can upgrade and merge on that draft PR at Syndesis.

It's a bit shame that we're a bit interlocked on this dependency, but it is what it is.

zregvart:
There's a bit of synchronization issue between this and the Syndesis 2.0 version. Let's not get this merged right this moment before we hash out the version details.

docs/pom.xml
```diff
@@ -93,180 +82,52 @@
...
-                  </info>
-                </apiSource>
-              </apiSources>
-            </configuration>
-          </execution>
-          <execution>
-            <id>api-swagger-java</id>
-            <phase>prepare-package</phase>
+            <id>default-cli</id>
+            <phase>generate-resources</phase>
```
igarashitm, line 89: Hi @zregvart, is this on purpose to merge all services into a single doc? wondering if there's an easy way to put sub-sections per service.

igarashitm: Newer plugin doesn't look much different - I'll split the swagger doc like before.

igarashitm:
Thanks again @zregvart! merged into master
https://github.com/atlasmap/atlasmap/commit/9cef28b32fb8d56045add719d7b465fff874dd57
https://github.com/atlasmap/atlasmap/commit/e2d8d29b8b7ba5fc68549b7dbd01e42dd3d0b745
https://github.com/atlasmap/atlasmap/commit/72f344dd350cf88a5dd74a601a8a0826fafea344
https://github.com/atlasmap/atlasmap/commit/0cfef84c1e66d7848356e003e336c8cb25aaa2bb

docs/pom.xml
zregvart: I think it was just simpler for me to do.

PR #1673, opened 2020-01-15
Fixes #1487 Enable support for mapping nested collection when target …

### LINKED ISSUE #1487: Enable support for mapping asymmetric nested collection where target is deeper
#1486 enables support the other way. It should be possible to map:

`/Person<a>/fullName to /Company<b>/Org<c>/Employee<d>/fullName`
given a=d, b=0, c=0

Consider enabling mapping from nested collection to nested collection with 0 padding in parent levels if source nest  depth < target nest depth. It needs to be clearly communicated to the user that 0 is used for parent levels.

rkorytkowski:
…is deeper

https://github.com/atlasmap/atlasmap/issues/1487


lib/core/src/main/java/io/atlasmap/core/validate/BaseModuleValidationService.java
```diff
@@ -219,20 +221,40 @@ protected void validateField(String mappingId, Field sourceField, Field targetFi
...
         if (direction == FieldDirection.TARGET) {
             AtlasPath sourcePath = null;
+            Integer sourceCount = null;
             if (sourceField != null) {
                 sourcePath = new AtlasPath(sourceField.getPath());
+                sourceCount = sourcePath.getCollectionSegmentCount();
+                List<Action> actions = sourceField.getActions();
+                if (actions != null) {
+                    for (Action action : actions) {
+                        if (action instanceof Split) {
```
rkorytkowski: Not very elegant, but Split action is actually producing a collection so we need to increment the count of collections for validation purposes.

igarashitm: Ah... this is something we have to refactor, take transformation into account for detecting collection nest level. We could have more ONE_TO_MANY transformation down the line, so we have to check with it. Isn't opposite way also a problem? Concatenate will make collection into non-collection.

rkorytkowski: One more issue here is that in the UI one first selects fields to map and only then selects a transformation... In such a case we will not allow selecting fields even though they could be later transformed to a valid case... 

However, since we are not yet fully committing to support all different cases here I'd say that we don't support that edge case right now. It would require extra work on validation logic and in the UI, which later would not be needed having all cases valid, which is our ultimate goal.

For the same reason I'd not try to address recognizing transformations down the line for the validation purposes.

I'll look into test cases for concatenate to make sure it's properly supported and see, if it can be better addressed than with instanceof.

igarashitm: Yes that's the concern... maybe we shouldn't ban field selection at UI side and just have to rely on validation errors, so user can fix it later by modifying transformation.
I think we can move it to a separate issue as a bug and continue on there, apart from this enhancement.

rkorytkowski: Agree on relying on validation errors. Currently, we are duplicating checks in validation and UI selection logic.

rkorytkowski: I'll address instanceof and check the concatenate behavior.

igarashitm: Cc: @pleacu

rkorytkowski: Done! Please see my updates at https://github.com/atlasmap/atlasmap/pull/1673/files#diff-f9a31dd08584f9efeb8c374b222b4268R226-R243

PR #1688, opened 2020-01-21
Fixes #1687 Concatenate in expression applies only to the first item …

### LINKED ISSUE #1687: Concatenate in expression applies only to the first item of a collection
https://issues.redhat.com/browse/ENTESB-12698

rkorytkowski:
…of a collection

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
+++ b/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
@@ -571,10 +571,6 @@ public class DefaultAtlasContext implements AtlasContext, AtlasContextMXBean {
 
     private void processSourceFieldGroup(DefaultAtlasSession session, FieldGroup sourceFieldGroup) throws AtlasException {
         processSourceFieldMappings(session, sourceFieldGroup.getField());
-        if (session.head().getSourceField() instanceof FieldGroup) {
-            sourceFieldGroup.getField().clear();
-            sourceFieldGroup.getField().add(session.head().getSourceField());
-        }
         session.head().setSourceField(sourceFieldGroup);
         Field processed = applyFieldActions(session, session.head().getSourceField());
         session.head().setSourceField(processed);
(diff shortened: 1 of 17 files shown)


rkorytkowski:
@igarashitm we'll need this bug fix for 7.6

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
```diff
@@ -604,25 +601,10 @@ private void processSourceFieldMappings(DefaultAtlasSession session, List<Field>
                 return;
             }
 
-            if (sourceField.getIndex() != null) {
-                AtlasPath path = new AtlasPath(sourceField.getPath());
-                if (path.hasCollection() && !path.isIndexedCollection()) {
-                    if (sourceField.getIndex() != 0) {
-                        AtlasUtil.addAudit(session, sourceField.getDocId(),
-                            String.format("Index on a field is not supported, if there is a collection without index on the path",
-                                sourceField.getPath()), sourceField.getPath(), AuditStatus.ERROR, null);
```
rkorytkowski: We actually have to support it now since expressions require indexed parameters and we allow to pass a collection without index to e.g. concatenate all items.

lib/core/src/main/java/io/atlasmap/core/validate/BaseModuleValidationService.java
```diff
@@ -151,19 +149,15 @@ protected void validateMapMapping(Mapping mapping, List<Validation> validations)
         if (getMode() == AtlasModuleMode.SOURCE) {
             FieldGroup sourceFieldGroup = mapping.getInputFieldGroup();
             if (sourceFieldGroup != null) {
-                multipleFieldSelectionValidator.validate(validations, mappingId,
```
igarashitm: Can we skip this only when it's expression mapping?

rkorytkowski: Fair point, but instead of enabling this validation back, I've made it behave so that when you select a collection + any other field then only the first element of a collection is used. You can of course apply concatenate transformation on a collection field and have all elements.

rkorytkowski: Created https://github.com/atlasmap/atlasmap/issues/1690

igarashitm: That sounds good! maybe also add a WARN validation result when it only picks up first one? we can file separate issue.

rkorytkowski: Cool, I'll add a warning shortly. No need for a separate issue.

rkorytkowski: Done!

igarashitm:
LGTM! @rkorytkowski can you do me one more favor, have another PR against atlasmap-1.43.x branch?

rkorytkowski:
Backport: https://github.com/atlasmap/atlasmap/pull/1691

Issue #1704, opened 2020-01-27
Correct atlasmap app template component hook for vscode


pleacu:
A partial implementation of angular routing was pushed that removed the main body template.  This caused vscode to display nothing when internal running the atlasmap app.

pleacu:
Aurelien was able to fix this is vscode-atlasmap

PR #1705, opened 2020-01-27
fix: correct app component template for selector atlasmap-dev-root

pleacu:
in support of vscode.  WIll need to get migrated to 1.43.x
Fixes: #1704

### DIFF
--- a/ui/src/app/app.component.ts
+++ b/ui/src/app/app.component.ts
@@ -18,7 +18,8 @@ import { Component } from '@angular/core';
 
 @Component({
   selector: 'atlasmap-dev-root',
-  template: '<router-outlet></router-outlet>',
+  template: '<data-mapper-example-host></data-mapper-example-host>',  // Required for vscode
+  // template: '<router-outlet></router-outlet>',
 })
 
 export class AppComponent {}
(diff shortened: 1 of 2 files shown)


ui/src/app/app.component.ts
```diff
@@ -18,7 +18,8 @@ import { Component } from '@angular/core';
 
 @Component({
   selector: 'atlasmap-dev-root',
-  template: '<router-outlet></router-outlet>',
```
igarashitm: This disables multiple mapping definition support, DataMapperExampleHostComponent is now delivered through angular router https://github.com/atlasmap/atlasmap/blob/master/ui/src/app/app-routing.module.ts#L12

As it has been working for browsers without any URL change, I suspect root cause is in vscode-atlasmap and/or vscode webview
https://github.com/jboss-fuse/vscode-atlasmap

@AbbassFaytaroony do you have chance to look into this problem? I'll also look into.

igarashitm: It's getting `Unhandled Navigation Error` in @angular/core when it tries to handle `<router-outlet>` via vscode WebView.

igarashitm: This is replacing `<base href=` to the absolute URL, but it still doesn't work
https://github.com/jboss-fuse/vscode-atlasmap/blob/master/src/atlasMapWebView.ts#L95

pleacu:
Corrected in vscode-atlasmap

PR #1724, opened 2020-01-30
chore: Re-add free view buttons to toolbar

jpav:


jpav:
Thanks, Paul. To your points:

- I agree about the closeness. I'm using an 'experimental' version of a toolbar components since it's the only way I know of to provide the separators the UX team asked for. I'm currently keeping the JIRA issue open until they work out what they really want with the Patternfly team and in case I need to back out to using the regular toolbar components.
- Unfortunately, I can't verify this until I get my browser fully working again after the O/S upgrade I did.
- I agree this would be worth looking into.

Issue #1747, opened 2020-02-05
UI_REACT: UI should indicate when containers are collections


jpav:
Seems like the simplest way would be to just add square brackets behind the name.

PR #1750, opened 2020-02-06
fix: UI_REACT: UI should indicate when containers are collections

jpav:
Fixes: #1747

### DIFF
--- a/ui-react/packages/atlasmap-core/src/utils/to-ui-models-util.ts
+++ b/ui-react/packages/atlasmap-core/src/utils/to-ui-models-util.ts
@@ -6,10 +6,12 @@ import { DocumentDefinition } from "../models/document-definition.model";
 import { ConfigModel } from "../models/config.model";
 
 export interface IAtlasmapGroupWithField extends IAtlasmapGroup {
+  isCollection: boolean;
   amField: Field;
 }
 
 export interface IAtlasmapFieldWithField extends IAtlasmapField {
+  isCollection: boolean;
   amField: Field;
 }
 
@@ -27,6 +29,7 @@ function fromFieldToIFieldsGroup(field: Field): IAtlasmapGroupWithField | null {
     id: `${field.docDef.uri}:${field.docDef.isSource ? 'source' : 'target'}:${field.uuid}`,
     name: field.name,
     type: field.type,
+    isCollection: field.isCollection,
     fields: fields,
     amField: field
   } : null;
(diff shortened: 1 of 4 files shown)


pleacu:
@jpav - the code looks fine but don't we have an icon for arrays/complex containers in the angular UI??  We should try to match that if we can

riccardo-forina:
FYI, this broke stories in storybook. Also, since "collection" is an Atlasmap specific concern, it would have been better not to change the models and components inside the CanvasView. I'll try to address this in a future PR.

jpav:
@riccardo-forina Is there some way we can get the stories pulled into the regular build process then? Nothing indicated there was a problem. Also, not sure what you mean by the collection concern. It's certainly important for the user to know whether fields are collections are not, so not sure how to accomplish showing them that distinction otherwise. I'm probably just missing your meaning though. Could you expound a bit?

riccardo-forina:
> @riccardo-forina Is there some way we can get the stories pulled into the regular build process then? Nothing indicated there was a problem. 

Yeah, we should improve that letting CI build also the storybooks and run the tests. I'll do that today.

> Also, not sure what you mean by the collection concern. It's certainly important for the user to know whether fields are collections are not, so not sure how to accomplish showing them that distinction otherwise. I'm probably just missing your meaning though. Could you expound a bit?

All the components that are part of the [`Canvas`](https://github.com/atlasmap/atlasmap/tree/master/ui-react/packages/atlasmap-ui/src/Canvas) and [`CanvasView`](https://github.com/atlasmap/atlasmap/tree/master/ui-react/packages/atlasmap-ui/src/CanvasView) folders are not strictly about Atlasmap; what they do is solve the rendering of html elements inside an svg with lines connecting them. Ideally those components can/will be used by other projects (eg. Syndesis) to do something similar to what we do on Atlasmap but for other domains.

For this reason these components should not address Atlasmap specific concerns - like the collection one - because it's not in scope with the components' goal.

I'll try moving those components in a new package, with their own stories to make this separation more clear.

riccardo-forina:
#1781 should address building storybook (and run tests and the linter too) on CI.

riccardo-forina:
And about the bit about `CanvasView`, it looks like we (I) cut some more corner that I'd like to admit and there are many Atlasmap specific bits there. A good way to quickly find them is to search for usages of `@patternfly/react-core`, we have many points where we have strings and UI elements (dropdowns, buttons) that are about Atlasmap's UI.

I guess for the moment this is ok, but I think we should revisit this in a later moment if we'll ever need to use these components in a different project (namely, Syndesis).

PR #1753, opened 2020-02-06
fix: ENTESB-12726: Remove source/target behavior

jpav:
Fixes: [ENTESB-12726](https://issues.redhat.com/browse/ENTESB-12726)


jpav:
It was indeed suggested by the UXD team (See [ENTESB-12726](https://issues.redhat.com/browse/ENTESB-12726). After testing a bit, I discovered that the functionality to remove the item was already there, just not any visual indication of that, so all I needed to do was add the icon.

Issue #1749, opened 2020-02-05
UI_REACT: Add ability to modify mapping index


pleacu:
Currently you cannot change the mapping index.

PR #1758, opened 2020-02-07
fix: support mapping index modifications

pleacu:
Fixes: #1749


pleacu:
wait for circleci to get fixed - will resubmit

Issue #1763, opened 2020-02-10
UI_REACT: Show multiplicity transformation options/args in a pulldown.


pleacu:
Current UI react does not allow changing many-to-one multiplicity transfomration options or arguments (or one-to-many split delimiters).

arguments.  ref angular UI:

![sc1](https://user-images.githubusercontent.com/2981421/74157278-0106b700-4be6-11ea-9b3a-d72e36771fa4.png)

PR #1765, opened 2020-02-11
fix: support multiplicity transformations

pleacu:
Fixes: #1763

@riccardo-forina - the mapping details needs to move to the 'sm' patternfly font family - need help setting the style sheet...


riccardo-forina:
Uhm, AFAIK there is no way to make a section of the UI smaller. I mean, you can go and change the font size on a per component basis but only for those components that allow it which is titles and plain text. I'm not sure you can do it for other components. @dlabaj might help here.

pleacu:
Cool - otherwise good to go?

Issue #1766, opened 2020-02-11
UI_REACT: Add a pull-down for multiplicity delimiter


pleacu:
Currently multiplicity delimiter in react is an input widget:
![sc1](https://user-images.githubusercontent.com/2981421/74252398-a5a6f880-4cbb-11ea-86b4-475f2fc3b424.png)
It should be a pulldown menu:

![sc2](https://user-images.githubusercontent.com/2981421/74252621-eef74800-4cbb-11ea-8476-8e32cbd313bc.png)

PR #1776, opened 2020-02-14
fix: support multiplicity transformation delimiter pulldown and updat…

pleacu:
…e css styling

Fixes: #1766

sc1


riccardo-forina:
Uhm, I feel like we could benefit from some pointers from @atlasmap/uxd

pleacu:
Here's a similar capture of the angular UI:
sc2

@dongniwang - wdyt?

pleacu:
Updated to remove the lightgray background color:
sc3

dongniwang:
Hey @riccardo-forina and @pleacu - thanks for tagging me. 

Are we talking about the delimiter dropdown? I'm a little bit confused about the relationship between the dropdown and the content beneath it. What does the interaction look like?

pleacu:
The dropdown is a standard dropdown menu.  I think the main area of
contention was  the use of a lightgray background to help distinguish the
sources mapping fields from the target mapping fields.

PR #1781, opened 2020-02-17
chore(react-ui): improve the build process

riccardo-forina:
Now code is automatically linted and storybooks are being built too.

It would be nice to get the generated artifacts somewhere, but I don't know my way around the maven build.

Changes on the `src` files are to address problems found by the linter and that have been automatically fixed running `yarn format` on the relative package.


riccardo-forina:
```
[INFO] AtlasMap Data Mapper UI ............................ SUCCESS [02:04 min]
[INFO] AtlasMap Standalone Application .................... SUCCESS [ 46.170 s]
[INFO] AtlasMap Data Mapper UI: React ..................... SUCCESS [07:02 min]
```

It looks like it's working fine, at the expense of a way longer build. The `UI: React` step used to be around 3m I think before this PR.

PR #1791, opened 2020-02-19
feat(react-ui): add stories to atlasmap-core

riccardo-forina:
There are two stories right now, aboout how to use Atlasmap in a standalone way (it's code is directly used by the atlasmap-standalone package) and in an embedded way (this code will be used by Syndesis as is).

In general this should allow working on the core without relying on the atlasmap-standalone package, easing the debugging.

I also removed the example apps inside the core and ui packages. We never used them in favour of Storybook.


riccardo-forina:
That's the port used by storybook's server. 6006 is the default and it was used by the ui package, so I bumped it by one to be able to run both storybooks at the same time without having to manually change the port or rely on the interactive dialog asking to do that.

Issue #1533, opened 2019-11-25
Add Camel3 support


igarashitm:
Maybe by adding `camel3/` sub-module next to `camel` and add build option with maven profile. So that we can get both.

Actually we might not even need profile, just build both and release camel-atlasmap and camel3-atlasmap, then user can choose one of them.

PR #1797, opened 2020-02-20
chore: Upgrade to Camel 3.1

zregvart:
This upgrades to Camel 3.1. Notable changes are that the in/out pattern
on Exchange is now deprecated and Exchange how has the `message`
property. With that the original message can be mutated with the new
body resulting from the transformation.

### DIFF
--- a/camel/pom.xml
+++ b/camel/pom.xml
@@ -45,6 +45,10 @@
       <artifactId>camel-core</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.camel</groupId>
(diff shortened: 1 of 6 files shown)


zregvart:
Since Camel 3.1 has not been released I've created a draft pull request. I'll update once Camel 3.1.0 is released.

igarashitm:
@zregvart thanks for the work! can we instead create a `camel3` sub module next to `camel` and having both for a while?

igarashitm:
This is for #1533

zregvart:
Superseded by #1798

PR #1798, opened 2020-02-20
feat: Camel Atlasmap component targeting Camel 3.1

### LINKED ISSUE #1533: Add Camel3 support

zregvart:
This creates a new Camel component utilizing Camel 3.1, basically
copying the existing component and setting the version to Camel 3,
migrating where needed. Notable changes are that the in/out pattern on
Exchange is now deprecated and Exchange how has the `message` property.
With that the original message can be mutated with the new body
resulting from the transformation.

One other thing of note is that the Spring and Spring Boot version
levels are now bumped to versions used in Camel 3.1.

Fixes #1533


zregvart:
This should have been a draft PR, sorry for that.

zregvart:
I probably missed a thing or two. I'm mostly concerned about OSGi stuff and having not changed v2 to v3 somewhere that matters.

rvnyk:
camel  3.1.0 is released.  camel3.version  3.1.0-SNAPSHOT  in parent pom should be replaced.would that fix the build?

zregvart:
I've updated the pull request with the 3.1.0 release of Camel and rebased on master. The new addition is that the version of Spring framework needs to be added as Camel 3.1 is using version 5.2 of Spring framework.

Issue #1804, opened 2020-02-22
UI_REACT: Move topology buttons to bottom toolbar


jpav:
Related to [ENTESB-12730](https://issues.redhat.com/browse/ENTESB-12730)

PR #1805, opened 2020-02-22
fix: UI_REACT: Move topology buttons to bottom toolbar

jpav:
Fixes: #1804

### DIFF
--- a/docs/src/main/asciidoc/developer-guide/swagger/core/overview.adoc
+++ b/docs/src/main/asciidoc/developer-guide/swagger/core/overview.adoc
@@ -6,7 +6,7 @@
 
 === Version information
 [%hardbreaks]
-__Version__ : 1.44.0-SNAPSHOT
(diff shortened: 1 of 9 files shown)


docs/src/main/asciidoc/developer-guide/swagger/core/overview.adoc
```diff
@@ -6,7 +6,7 @@
 
 === Version information
 [%hardbreaks]
-__Version__ : 1.44.0-SNAPSHOT
```
riccardo-forina: What’s this about?

jpav: Not entirely sure. Looks like the SNAPSHOT as a whole was upgraded to 2.0.0 somewhere along the line, but somehow these doc files were missed. mvnw install repeatedly causes this to happen after I reset the version to 1.44.0 in these files.

Issue #1806, opened 2020-02-24
UI_REACT: Deselect current mapping when background clicked


jpav:
Relates to [ENTESB-12724](https://issues.redhat.com/browse/ENTESB-12724)

PR #1807, opened 2020-02-24
fix: UI_REACT: Deselect current mapping when background clicked

jpav:
Fixes: #1806 

I'm guessing, however, that there might be a more appropriate way to fix this. Based on previous comments, I believe this violates the division between the 'generic' REACT components we've created and the AtlasMap-specific ones. Should I instead place this code in a new AtlasMap component that gets inserted after AtlasmapLayout and before CanvasTransform?

### DIFF
--- a/ui-react/packages/atlasmap-ui/src/Canvas/CanvasTransforms.tsx
+++ b/ui-react/packages/atlasmap-ui/src/Canvas/CanvasTransforms.tsx
@@ -1,5 +1,6 @@
-import React, { FunctionComponent } from 'react';
+import React, { FunctionComponent, useCallback } from 'react';
 import { useCanvas } from './CanvasContext';
+import { useAtlasmapUI } from '../AtlasmapUI/AtlasmapUIProvider';
 
 export const CanvasTransforms: FunctionComponent = ({ children }) => {
   const { panX, panY, zoom /* width, height, xDomain, yDomain*/ } = useCanvas();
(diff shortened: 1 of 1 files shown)


riccardo-forina:
Yeah, ideally we'd want to:

1. add an `onClick` event handler [here](https://github.com/atlasmap/atlasmap/blob/master/ui-react/packages/atlasmap-ui/src/AtlasmapUI/AtlasmapCanvasView/AtlasmapCanvasView.tsx#L93)
2. forward the onClick to the `CanvasViewCanvas` component [here](https://github.com/atlasmap/atlasmap/blob/master/ui-react/packages/atlasmap-ui/src/CanvasView/CanvasView.tsx#L21)
3. add the `onClick` to the underlying svg component [here](https://github.com/atlasmap/atlasmap/blob/master/ui-react/packages/atlasmap-ui/src/CanvasView/components/CanvasViewCanvas.tsx#L33)

I'm currently working on a better separation of concerns between the 3 actors we have - Canvas, CanvasView and AtlasmapUI - that will probably make the above slightly simpler, so if you want to go ahead and merge this I guess it's ok.

jpav:
Ah, that makes sense. Pretty much create the ability to register a listener on the more generic classes that gets called when appropriate, and setup the specific AtlasMap class to register such a listener. That's pretty straight-forward UI stuff; I'll look towards that in the future. For now, taking advantage of your work to address this more globally and checking it in.

PR #1816, opened 2020-02-25
feat(react-ui): ui rewrite

riccardo-forina:
This is a full rewrite of the UI layer. Storybook is now the cornerstone for all UI related work and can/should be used to document how Atlasmap's UI should be developed.


ui-react/packages/atlasmap/src/Atlasmap/utils/field.ts
```diff
@@ -155,6 +182,23 @@ export function addToCurrentMapping(field: Field): void {
   cfg.mappingService.fieldSelected(field);
 }
 
+/**
+ * Add the specified field to the current mapping.
```
pleacu, line 186: Remove the specified fields from the current mapping.

PR #1833, opened 2020-02-28
fix: Freeview button should be moved to bottom toolbar

jpav:
Fixes: #1829

### DIFF
--- a/ui-react/packages/atlasmap-ui/src/AtlasmapUI/AtlasmapLayout.tsx
+++ b/ui-react/packages/atlasmap-ui/src/AtlasmapUI/AtlasmapLayout.tsx
@@ -5,6 +5,7 @@ import { useAtlasmapUI } from './AtlasmapUIProvider';
 import { Loading } from '../common';
 import { AtlasmapSidebar } from './AtlasmapSidebar';
 import { CanvasViewToolbar } from '../CanvasView/components';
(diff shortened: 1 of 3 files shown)


ui-react/packages/atlasmap-ui/src/AtlasmapUI/AtlasmapLayout.tsx
```diff
@@ -30,6 +31,8 @@ export const AtlasmapLayout: FunctionComponent<IAtlasmapLayoutProps> = ({
 }) => {
   const { pending, isEditingMapping } = useAtlasmapUI();
 
+  const toggleFreeView = useCanvasViewOptionsContext().toggleFreeView;
```
riccardo-forina: `const { toggleFreeView } = useCanvasViewOptionsContext();` maybe?

jpav: That actually worked. Didn't realize it could figure out just by the name I wanted the method rather than the flag.

riccardo-forina: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring

Issue #1562, opened 2019-12-04
Support reading parent field


igarashitm:
Read parent complex object and build a tree of AtlasMap Field. This is required for #1507.

Issue #1507, opened 2019-11-21
Support filtering in expression


igarashitm:
Map only items of a collection having certain values e.g.

~~`Person<a>[gender="m"]/fullName` -> `Person\<b\>/fullName`~~

=> The target will only contain male

Talked with Hiram as an author of expression engine, we'll go with `filter` and `select` functions
`select(filter( ${DocId:/Person<>}, ${/gender}=="m"), ${/fullName})`

One thing that might not straight forward is, it needs to access parent node (`/Person<>`) while AtlasMap fields readers doesn't support reading parent complex object.

PR #1844, opened 2020-03-03
feat: Support reading parent field

igarashitm:
Fixes: #1562
Fixes: #1507
Enables FILTER() and SELECT() in expression for real usecase documents i.e. Java/Json/XML


igarashitm:
@rkorytkowski I believe I didn't break what you've done but can you double check?

rkorytkowski:
@igarashitm, it looks great. NestedCollection*Tests cover all use cases I could imagine and they pass so I'd just merge it.

PR #1854, opened 2020-03-06
fix: Line the edit and delete icons up

pleacu:
Fixes: #1847

Still not perfect but the icons line up better.  @riccardo-forina is actively modifying the look and feel - will rely on his changes for final appearance.

sc1


pleacu:
Yup - I basically made a separate span for the buttons so they would always
be kept as a pair (load mockdoc.adm and you'll see they get split).

Issue #1860, opened 2020-03-09
UI_REACT: Add initial conditional expression support


pleacu:
Examine the conditional expression f(x) input widget

PR #1873, opened 2020-03-13
fix: Initial conditional expression support

pleacu:
Fixes: #1860

Successfully parse existing mapping conditional expressions

sc1

No widget tokenization yet - that's next.

### DIFF
--- a/ui-react/packages/atlasmap-core/src/react/Atlasmap.tsx
+++ b/ui-react/packages/atlasmap-core/src/react/Atlasmap.tsx
@@ -245,7 +245,6 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
         onGetMappingExpressionStr={onGetMappingExpressionStr}
         onToggleExpressionMode={() => {
           onToggleExpressionMode();
-          expressionTokens = [''];
         }}
         onExportAtlasFile={handleExportAtlasFile}
         onImportAtlasFile={(file: File) => handleImportAtlasFile(file)}
--- a/ui-react/packages/atlasmap-core/src/react/Atlasmap.tsx
+++ b/ui-react/packages/atlasmap-core/src/react/Atlasmap.tsx
@@ -84,6 +84,9 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
     deleteAtlasFile,
     changeActiveMapping,
     documentExists,
+    onToggleExpressionMode,
+    onConditionalMappingExpressionEnabled,
+    onGetMappingExpressionStr,
     toggleMappingPreview,
     toggleShowMappedFields,
     toggleShowUnmappedFields,
@@ -100,6 +103,7 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
 
   const sources = useAtlasmapSources(sourceFilter);
   const targets = useAtlasmapTargets(targetFilter);
+  let expressionTokens = [''];
 
   const handleExportAtlasFile = () => {
     onExportAtlasFile(exportAtlasFile);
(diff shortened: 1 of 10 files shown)


pleacu:
@riccardo-forina - wdyt?  This gets the expression component  classes introduced.

ui-react/packages/atlasmap-core/src/react/Atlasmap.tsx
```diff
@@ -235,9 +239,18 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
         onShowMappingPreview={toggleMappingPreview}
         onShowMappedFields={toggleShowMappedFields}
         onShowUnmappedFields={toggleShowUnmappedFields}
+        onConditionalMappingExpressionEnabled={
+          onConditionalMappingExpressionEnabled
+        }
+        onGetMappingExpressionStr={onGetMappingExpressionStr}
+        onToggleExpressionMode={() => {
+          onToggleExpressionMode();
+          expressionTokens = [''];
```
riccardo-forina: Uhm, usually this in the context of a React component is an error. There is no need to "reset" the expressionTokens, if that's the goal. Once the rendering function returns any volatile variable will be lost.

ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
```diff
@@ -579,6 +579,15 @@ export class MappingManagementService {
     }
   }
 
+  conditionalMappingExpressionEnabled(): boolean {
+    if (!this.cfg || !this.cfg.mappings || !this.cfg.mappings.activeMapping) {
+      return false;
+    };
+    return (this.cfg.mappings && this.cfg.mappings.activeMapping &&
```
riccardo-forina: This return can be simplified to `return !!(this.cfg.mappings?.activeMapping?.transition?.enableExpression?)` with the new TS

ui-react/packages/atlasmap-ui/src/AtlasmapUI/AtlasmapCanvasView/AtlasmapCanvasView.tsx
```diff
@@ -47,12 +55,26 @@ export const AtlasmapCanvasView: FunctionComponent<IAtlasmapCanvasViewProps> = (
...
+    return onConditionalMappingExpressionEnabled();
+  }, [onConditionalMappingExpressionEnabled]);
+
+  const getMappingExpressionStr = useCallback(() => {
+    return onGetMappingExpressionStr();
+  }, [onGetMappingExpressionStr]);
+
+  const toggleExpressionMode = useCallback(() => {
+    onToggleExpressionMode();
+    expressionTokens[0] = 'ratsky';
```
riccardo-forina: Same comment as above about setting and resetting variables inside event handlers.

ui-react/packages/atlasmap-ui/src/CanvasView/components/CanvasViewToolbar.tsx
```diff
@@ -35,18 +45,23 @@ export const CanvasViewToolbar: FunctionComponent<ICanvasViewToolbarProps> = ({
           <Button
             variant={'plain'}
             aria-label="Enable/ Disable conditional mapping expression"
-            onClick={onToggleExpressionMode}
+            onClick={() => {
+              onToggleExpressionMode();
+              // setCondExprEnabled(onConditionalMappingExpressionEnabled());
+              condExprEnabled = onConditionalMappingExpressionEnabled();
```
riccardo-forina: Sorry @pleacu, but this is the same thing as the previous comment. You can't change a local variable in an event handler in React, it makes no sense.

pleacu: Yes - will address that today.  My issue was that I was calling onConditionalMappingExpressionEnabled() to set a useState boolean.  The core service function (conditionalMappingExpressionEnabled) returns true but the onConditionalMappingExpressionEnabled() returns false in the onClick():

setCondExprEnabled(onConditionalMappingExpressionEnabled());

Is there a special syntax to invoke onConditionalMappingExpressionEnabled()??

PR #1896, opened 2020-03-23
wip: Create 'New Mapping' button

jpav:
@riccardo-forina I've tried several things and I'll still stumped on this. When you get a chance, can you take a look at this code and let me know why you think the new mapping I'm creating after clicking the plus icon doesn't work. Seems like all the code is being called appropriately to create the mapping itself, but the GUI doesn't update. I must be missing something important.

### DIFF
--- a/ui-react/packages/atlasmap-core/src/components/field/field-util.ts
+++ b/ui-react/packages/atlasmap-core/src/components/field/field-util.ts
@@ -128,6 +128,17 @@ export function createMapping(source: Field, target: Field): void {
 }
 
 /**
+ * Create a new mapping using the specified source and target IDs.
+ *
+ * @param source
+ * @param target
+ */
+export function newMapping(): void {
+  const cfg = ConfigModel.getConfig();
+  cfg.mappingService.newMapping();
+}
+
+/**
  * Add the specified field to the current mapping.
  *
  * @param field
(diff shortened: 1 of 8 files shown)


riccardo-forina:
Is the [`onUpdates`](https://github.com/jpav/atlasmap/blob/g1862/ui-react/packages/atlasmap-core/src/react/AtlasmapProvider.tsx#L154-L174) callback being triggered after clicking on the add button? 

If not, we are either not registered to changes in the right observer, or the observer didn't get triggered for some reason.

If it does, what's the content of [`initializationService.cfg.mappings`](https://github.com/jpav/atlasmap/blob/g1862/ui-react/packages/atlasmap-core/src/react/AtlasmapProvider.tsx#L167)? Does it look right?

jpav:
onUpdates is not being called. Would it be correct to call that directly after making it available via useAtlasmap, or is there some other wiring missing?

riccardo-forina:
If it's not being called you have to track down which subscription is being called and we'll have to register for updates on that too.
But I have the feeling that an empty mapping isn't supported by the "backend" part of the UI -the RxJS code- at all...

jpav:
Okay, I'm clearly over my head on this. I'll talk with you more tomorrow (i.e., later today I guess) about it.

PR #1900, opened 2020-03-25
Fixes #1511 Enable support asymmetric nested collection where source …

### LINKED ISSUE #1511: Enable support asymmetric nested collection where source is deeper
Map values from a collection when source has more collections on the path than target e.g.

Patient\<a\>/personalDetails\<b\>/name\<c\>/fullName -> Person\<d\>/name\<e\>/fullName

=> By default match source and target indexes from left to right i.e. d=a, e=b*c

Add validation warning that collection indexes are mapped this way. For user defined collection index mapping, we need #781

rkorytkowski:
…is deeper

### DIFF
--- a/lib/core/src/test/java/io/atlasmap/core/DefaultAtlasExpressionProcessorTest.java
+++ b/lib/core/src/test/java/io/atlasmap/core/DefaultAtlasExpressionProcessorTest.java
@@ -103,7 +103,7 @@ public class DefaultAtlasExpressionProcessorTest extends BaseDefaultAtlasContext
         assertEquals(1, childFieldGroup.getField().size());
         assertEquals("foo0", childFieldGroup.getField().get(0).getValue());
         childFieldGroup = (FieldGroup) fieldGroup.getField().get(1);
-        assertEquals("/testPathfoo<2>", childFieldGroup.getPath());
+        assertEquals("/testPathfoo<1>", childFieldGroup.getPath());
         assertEquals(1, childFieldGroup.getField().size());
         assertEquals("foo2", childFieldGroup.getField().get(0).getValue());
     }
@@ -148,7 +148,7 @@ public class DefaultAtlasExpressionProcessorTest extends BaseDefaultAtlasContext
         assertEquals("/testPathfoo<0>/value", child.getPath());
         assertEquals("foo0", child.getValue());
         child = fieldGroup.getField().get(1);
-        assertEquals("/testPathfoo<2>/value", child.getPath());
+        assertEquals("/testPathfoo<1>/value", child.getPath());
         assertEquals("foo2", child.getValue());
     }
 }
--- a/lib/itests/core/src/test/java/io/atlasmap/itests/core/NestedCollectionJavaTest.java
+++ b/lib/itests/core/src/test/java/io/atlasmap/itests/core/NestedCollectionJavaTest.java
@@ -155,6 +155,37 @@ public class NestedCollectionJavaTest {
         assert2ndLevelCollection(nestedTarget);
     }
 
+    @Test
+    public void test3To2LevelNestedCollection() throws Exception {
+        TargetClass target = processNestedJavaCollection(Arrays.asList("3-2"));
+
+        assertEquals(2, target.getSomeArray().length);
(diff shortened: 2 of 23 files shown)


lib/core/src/test/java/io/atlasmap/core/DefaultAtlasExpressionProcessorTest.java
```diff
@@ -103,7 +103,7 @@ public void testFilter() throws Exception {
         assertEquals(1, childFieldGroup.getField().size());
         assertEquals("foo0", childFieldGroup.getField().get(0).getValue());
         childFieldGroup = (FieldGroup) fieldGroup.getField().get(1);
-        assertEquals("/testPathfoo<2>", childFieldGroup.getPath());
+        assertEquals("/testPathfoo<1>", childFieldGroup.getPath());
```
rkorytkowski: I've changed the behavior here. When some item is removed after applying the filter, adjust succeeding indexes.

lib/itests/core/src/test/java/io/atlasmap/itests/core/NestedCollectionJavaTest.java
```diff
@@ -155,6 +155,37 @@ public void testAsymmetricPaths2ndLevelNestedCollectionTo4LevelsNestedCollection
...
+    @Test
+    public void test3To2LevelNestedCollection() throws Exception {
+        TargetClass target = processNestedJavaCollection(Arrays.asList("3-2"));
+
+        assertEquals(2, target.getSomeArray().length);
+        assertEquals(4, target.getSomeArray()[0].getSomeArray().length);
+        assertEquals("array000", target.getSomeArray()[0].getSomeArray()[0].getSomeField());
+        assertEquals("array001", target.getSomeArray()[0].getSomeArray()[1].getSomeField());
+        assertEquals("array002", target.getSomeArray()[0].getSomeArray()[2].getSomeField());
+        assertNull(target.getSomeArray()[0].getSomeArray()[3].getSomeField());
```
rkorytkowski: Not sure about the behavior here. It's different from JSON and XML, but seems to have been added explicitly. It basically includes items as null values for paths when some child levels are missing in source and adds a warning for those paths. See https://github.com/atlasmap/atlasmap/blob/master/lib/modules/java/core/src/main/java/io/atlasmap/java/core/JavaFieldReader.java#L104-L109 For JSON and XML such paths are simply skipped.

rkorytkowski: These are the affected tests when I remove that functionality and make it the same as json and xml: https://github.com/atlasmap/atlasmap/blob/master/lib/modules/java/core/src/test/java/io/atlasmap/java/core/JavaFieldReaderTest.java#L93-L128

igarashitm: Hmm... say Company<>/Employee<>/Contact/Email, and some employee could have null contact. From that perspective Java reader behavior might be better. We might have to change json/xml behavior instead? It would be a separate bug.

rkorytkowski: I'm sorry I'll respond tomorrow as I'm too tired to reason about it today :)

rkorytkowski: You are right! The Java reader behavior is better. The logic would be otherwise broken if one mapped 2 fields e.g. Company<>/Employee<>/Contact/Email and Company<>/Employee<>/Contact/Name. If we skip one Employee due to null Contact or Email then we will adjust indexes after that employee and the next mapping will assign wrong names for employees for adjusted indexes. It's just a bit awkward to see null entries in a merged collection when source collection count > target collection count. Maybe in such a case (since indexes are not preserved anyway) we should skip null values. I'll create a separate issue to fix these.

rkorytkowski: Added https://github.com/atlasmap/atlasmap/issues/1906

Issue #1862, opened 2020-03-10
UI_REACT: Create 'New Mapping' button


jpav:
Related to [ENTESB-12722](https://issues.redhat.com/browse/ENTESB-12722)

PR #1902, opened 2020-03-26
fix: Create 'New Mapping' button

jpav:
Fixes: #1862 

Note, the empty mapping currently allows no way to assign fields to it, but I will address this in a separate issue.


ui-react/packages/atlasmap-core/src/react/Atlasmap.tsx
```diff
@@ -215,6 +216,11 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
     [createMapping]
   );
 
+  const handleNewMapping = useCallback(() => {
+    console.log('handleNewMapping');
```
riccardo-forina: Maybe remove this?

jpav: Ah, good catch

Issue #1604, opened 2019-12-16
Support new Conditional Mapping Expression


igarashitm:
Switch from `Expression` field action that belongs to a field to the one right under the `Mapping` which supports dynamic path resolution (#1512)
https://github.com/atlasmap/atlasmap/blob/master/lib/model/src/main/java/io/atlasmap/v2/Mapping.java#L28

New expression supports COMPLEX field as a parameter, which enables more advanced usecase like filtering #1507. 

xref: https://issues.redhat.com/browse/ENTESB-12402

igarashitm:
Here is an example for `filter` & `select` function that relies on dynamic path expression.
https://github.com/atlasmap/atlasmap/blob/master/lib/itests/core/src/test/resources/mappings/atlasmapping-expression-filter-select-java.json#L18

Note that first path expression is absolute (include documentId) and the others are relative.

pleacu:
This epic is now complete for initial release.  Any refactoring or new ideas should be addresses by a new issue.

igarashitm:
Reopening - primary focus for this issue is to support filter/select which requires COMPLEX field support in UI.

pleacu:
Hey @igarashitm 

Given this as an example (from atlas-itests-core-2.1.0-SNAPSHOT-tests.jar), the user has a simple mapping
from `/someArray<>/someField` to `/someArray<>/someField`.  They then establish this conditional expression
to the mapping:

`SELECT( FILTER( ${SourceClass:/someArray<>}, ${/someField} != 'v1' ), ${/someField} )`

![sc1](https://user-images.githubusercontent.com/2981421/86836964-42ec3a00-c06c-11ea-8aae-a82014199971.png)

In this scenario, someArray needs to be selectable from the conditional expression pull-down menu not as a mappable element but as a reference component in the expression with 

field.serviceObject.status = 'SUPPORTED'
field.serviceObject.fieldType = 'COMPLEX'

, correct?  The user would see something like:

![sc3](https://user-images.githubusercontent.com/2981421/86837232-9bbbd280-c06c-11ea-80e9-71ad7142d4c4.png)

Is this accurate?

igarashitm:
Right, I'm not sure if it should be selectable or not though. From this point, COMPLEX field does participate into mapping, but only for conditional mapping ATM.

pleacu:
[...]

PR #1915, opened 2020-03-31
fix: Support new expression location - mark old style as deprecated b…

pleacu:
…ut keep for legacy ADM files.

Fixes: #1604


pleacu:
Need to correct inputfield and outputfield for inspection generated meta data - will refile.

Issue #1918, opened 2020-04-02
Conditional Mapping Expression Error


SvenC56:
## Description
Hey I encountered this behaviour in Syndesis that conditional mapping are not working for me. I get this strange error after the deployment: `TypeError: Cannot read property 'value' of undefined`. If I deploy the app and go to the result I would excpect to get `true` or `false` values. Instead it looks like this:

![image](https://user-images.githubusercontent.com/14073259/78236081-d47e6580-74d9-11ea-8b12-9d7d9ac4b56e.png)

I if query the SQL statement in the Oracle SQL Developer it looks like this:
![image](https://user-images.githubusercontent.com/14073259/78236295-0e4f6c00-74da-11ea-832b-ee68475668c5.png)

I tried to setup conditional mapping like described here: https://access.redhat.com/documentation/en-us/red_hat_fuse/7.5/html/integrating_applications_with_fuse_online/mapping-data_ug

Mapping:
![image](https://user-images.githubusercontent.com/14073259/78235410-e7dd0100-74d8-11ea-9576-f32b9663d8c4.png)

Error after deployment:
![image](https://user-images.githubusercontent.com/14073259/78235541-0fcc6480-74d9-11ea-8045-89e79e4b4e6b.png)

igarashitm:
Hi @SvenC56 can you export your Syndesis integration and attach here?

igarashitm:
@SvenC56 And which version of Syndesis do you use?

SvenC56:
Hi @igarashitm I'm using Syndesis `1.9.0-20200402`. Attached you find the redacted export. I hope it helps.
https://svencieslok-my.sharepoint.com/:u:/g/personal/mail_svencieslok_de/EZd4wM8cPflGoT9xwTA34aMBZ-DimEKQneurAWz9DxrsJg?e=5bBKKH

igarashitm:
Thanks @SvenC56!

OK let's debug - here is what I got so far. The error message
`Unable to map expression index "0" to field node.`
means an array of source field in the MappingModel is empty in UI code.

Here is a mapping definition extracted from Syndesis integration
https://gist.github.com/igarashitm/1ea278b05e26cc3c9a46a18a8b819b1c

I found that the source field `/<>/ANZAHL_RUECKMUST` is placed under `inputField`, but not in `inputFieldGroup`
[...]

PR #1921, opened 2020-04-02
fix: Don't deserialize field actions before adding the mapped field t…

pleacu:
…hey reference.

Fixes: #1918

### DIFF
--- a/ui-react/packages/atlasmap-core/src/models/expression.model.ts
+++ b/ui-react/packages/atlasmap-core/src/models/expression.model.ts
@@ -73,7 +73,7 @@ export class FieldNode extends ExpressionNode {
         this.field = mapping.getMappedFieldByName(fieldParts[1], true)!;
       } else {
         // TODO: check this non null operator
-        this.field = mapping.getMappedFieldForIndex((index).toString(), true)!;
+        this.field = mapping.getMappedFieldForIndex((index+1).toString(), true)!;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


ui-react/packages/atlasmap-core/src/models/expression.model.ts
```diff
@@ -73,7 +73,7 @@ export class FieldNode extends ExpressionNode {
         this.field = mapping.getMappedFieldByName(fieldParts[1], true)!;
       } else {
         // TODO: check this non null operator
-        this.field = mapping.getMappedFieldForIndex((index + 1).toString(), true)!;
+        this.field = mapping.getMappedFieldForIndex((index).toString(), true)!;
```
igarashitm, line 76: Doesn't this cause a problem for the index editor in the mapping detail pane? They are 1-base.

pleacu: No - see mapping.model change below - it immediately subtracts the 1 from the index.  This was a real index arg - not a user index argf.

igarashitm: OK, so index editor only uses `getIndexForMappedField()` here
https://github.com/atlasmap/atlasmap/blob/master/ui/src/app/lib/atlasmap-data-mapper/components/mapping-detail/mapping-field-detail.component.html#L46

Then if we change the `getMappedFieldForIndex()` to be 0-base, we also have to change `getIndexForMappedField()` to be 0-base to avoid mixing 0-base and 1-base. In that case `MappingFieldDetail` component have to do +1 by itself to be 1-base.

pleacu: Went back to the original index formula

Issue #1903, opened 2020-03-26
UI_REACT: Create mapping when source field selected


jpav:
...and no mapping selected. This gets us back to the same behavior seen with the Angular app.

Related to: [ENTESB-12725](https://issues.redhat.com/browse/ENTESB-12725)

PR #1940, opened 2020-04-13
fix: UI_REACT: Create mapping when source field selected

jpav:
Fixes: #1903

### DIFF
--- a/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
+++ b/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
@@ -425,9 +425,11 @@ export class MappingManagementService {
     if (selectedField != null) {
       mapping.addField(selectedField, false);
       this.updateTransition(mapping);
+      this.selectMapping(mapping);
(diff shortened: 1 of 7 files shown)


ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
```diff
@@ -425,9 +425,11 @@ export class MappingManagementService {
     if (selectedField != null) {
       mapping.addField(selectedField, false);
       this.updateTransition(mapping);
+      this.selectMapping(mapping);
       this.notifyMappingUpdated();
+    } else {
+      this.selectMapping(mapping);
```
pleacu, line 431: Is this change really necessary?  Can't the selectMapping occur at the end?

jpav: Yes, AFAICT, because the selectMapping call is what sets the activeMapping, which the subsequent notifyMappingUpdated call relies upon to update the list of mappings.

PR #1942, opened 2020-04-15
fix: UI_REACT: Provide mapping table view

jpav:
This isn't a complete fix, but represents most of what needs to be done for this view. The main thing remaining is getting the entire path of sources and targets displayed.

### DIFF
--- a/ui-react/packages/atlasmap-ui/src/AtlasmapUI/MappingTable.tsx
+++ b/ui-react/packages/atlasmap-ui/src/AtlasmapUI/MappingTable.tsx
@@ -1,0 +1,113 @@
+import React, { ReactNode } from 'react';
+import { FunctionComponent } from 'react';
+import { TableIcon, SearchIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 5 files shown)


ui-react/packages/atlasmap-ui/src/AtlasmapUI/MappingTable.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export const MappingTable: FunctionComponent<IMappingTableProps> = ({
+  contextToolbar,
+  viewToolbar,
+}) => {
+  const { mappings } = useAtlasmapUI();
+
+  const mappingRows: Array<ReactNode> = [];
+  mappings.map((mapping, index) => {
```
riccardo-forina, line 53: Using `map` and discarding its return value is misleading. A `forEach` would be clearer in its intention here. I'd personally used a `reduce` to avoid having a `const mappingRows` that's then mutated in the `forEach`.

Issue #1914, opened 2020-03-30
UI_REACT: Add filtered search dropdown to expression box


pleacu:
If a user types '@' then that should trigger a source field search.

igarashitm:
@pleacu I think it's better to switch to new expression #1604 before implement field completion.

pleacu:
Hey @igarashitm - I was originally going to complete support using the old style expression location but I'll tackle 1604 first.  Thkx for the heads up.

pleacu:
@igarashitm - just to be clear - there was no Angular UI support for the new expression location correct?  (i.e. expression.component.ts)

igarashitm:
nope, angular UI only supports earlier expression (as a transformation). It's incompatible and the earlier version of expression support will be discontinued in React UI.

PR #1944, opened 2020-04-16
fix: Add expression box support for adding field nodes.

pleacu:
Fixes: #1914

Ran into an issue with the newest server.  Works with older server - will address server issue separately.

Still needs d&d support as well as keyboard nav (arrows, tab)

### DIFF
--- a/ui-react/packages/atlasmap-core/src/models/expression.model.ts
+++ b/ui-react/packages/atlasmap-core/src/models/expression.model.ts
@@ -76,6 +76,9 @@ export class FieldNode extends ExpressionNode {
         this.field = mapping.getMappedFieldForIndex((index+1).toString(), true)!;
       }
     }
+    if (field && !field.parsedData.parsedPath && field.field) {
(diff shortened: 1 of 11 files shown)


riccardo-forina:
It doesn't seem to be working for me? If I select a mapping the fx input is always empty, if I write @ and pick a field the input remains empty.

pleacu:
Hey @riccardo-forina  -  checked again on my end - seems to be okay.  Are you running into any of your changes with it?  Should just be master with my PR.

riccardo-forina:
No I simply clicked on the netlify preview link up above

pleacu:
@riccardo-forina - strange - it looks like netlify doesn't like something - the 'reset all' button also doesn't work for instance.

pleacu:
I think I found the netlify issue - strange it doesn't manifest itself outside of netlify...

pleacu:
@riccardo-forina - https://deploy-preview-1944--atlasmap.netlify.app/atlasmap-core/?path=/story/atlasmap--standalone works now.

Issue #1945, opened 2020-04-16
Save mapping as a part of validation for backward compatibility


igarashitm:
The validate mapping API used to save mapping implicitly, but now it doesn't after #1474. We have to add it back until UI explicitly invokes save mapping REST call (#1946).

PR #1947, opened 2020-04-16
fix: Save mapping as a part of validation for backward compatibility

igarashitm:
Fixes: #1945


lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
```diff
@@ -204,12 +197,17 @@ public AtlasMapping getMappingDefinition() throws AtlasException {
...
+        this.mappingDefinition = mapping;
+    }
+
+    public void setMappingDefinitionBytes(InputStream is) throws AtlasException {
         try {
             this.mappingDefinition = null;
             this.mappingDefinitionBytes = readIntoByteArray(is);
             if (LOG.isDebugEnabled()) {
-                LOG.debug(new String(this.mappingDefinitionBytes));
+                LOG.debug(Json.withClassLoader(classLoader).writeValueAsString(getMappingDefinition()));
```
pleacu, line 210: format? - space over

PR #1961, opened 2020-04-24
feat(react-ui): expose the API for embedding Atlasmap

riccardo-forina:
`AtlasmapProvider` can be fed an external document and a callback to get back the serialized mappings. This is needed by Syndesis to embed Atlasmap.

It's also now possible to quickly debug the json payload coming from Syndesis (or other sources) using Storybook, like this:

c4e5a27fa37bc75f385919c6f0d18cd6

This is part of ENTESB-12080

@igarashitm I'll need a release on npmjs of this to finalise the PR on Syndesis. We need `packages/atlasmap-core` and `packages/atlasmap` published, both with a minimum versiom of 2.0.0 so not to clash with the previously published modules.


igarashitm:
@riccardo-forina next milestone release will be `1.44.0-M3` and coming GA will be `1.44.0`, does it work for you?

PR #1962, opened 2020-04-24
fix(react-ui): make tree group closable on click

riccardo-forina:
Also decouples the tree component from the Group and Field visual
components.


riccardo-forina:
Please review this by checking the new UI>Tree stories on the preview storybook.

PR #1975, opened 2020-04-27
fix(react-ui): always show properties and constants

### LINKED ISSUE #1963: React UI: alway show the Property and Constants cards
Right now we don’t show them if there is no property or constant in use, making it impossible to add them. We should show the card document with a “no elements here” text as we did before.

riccardo-forina:
Fixes #1963 

This also removes some styling from the `Tree` family of components to delegate it to the specialised `DocumentXXX` components. It isn't strictly related but we needed some way to properly style the empty state and this change sneaked into the PR.


riccardo-forina:
Preview: 

image

PR #1979, opened 2020-04-28
StreamCache not being resetted after read

Delawen:
`org.apache.camel.StreamCache` can be read only once, so if we want to extract the payload of the same message on more than one step, we need to reset the stream.

This fix makes sure we will be able to re-read the body on future steps (if possible, as described on StreamCache): https://github.com/apache/camel/blob/master/core/camel-api/src/main/java/org/apache/camel/StreamCache.java#L25

The reset is done *after* the read in case the reset throws some error that makes the stream unusable (because it may really be a only-one-read stream) so it is prepared for future calls.

### DIFF
--- a/camel3/src/main/java/org/apache/camel/component/atlasmap/AtlasEndpoint.java
+++ b/camel3/src/main/java/org/apache/camel/component/atlasmap/AtlasEndpoint.java
@@ -33,6 +33,7 @@ import java.util.zip.ZipInputStream;
 import org.apache.camel.Exchange;
 import org.apache.camel.ExchangePattern;
 import org.apache.camel.Message;
+import org.apache.camel.StreamCache;
 import org.apache.camel.component.ResourceEndpoint;
 import org.apache.camel.spi.UriEndpoint;
 import org.apache.camel.spi.UriParam;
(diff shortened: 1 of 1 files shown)


Delawen:
@igarashitm is it fine to fix the issue here or should we look for some fix more internal to AtlasMap?

igarashitm:
@Delawen This looks very fishy, as we don't need to explicitly invoke `reset()` in camel2. Shouldn't it be camel3 that requires a fix or just a configuration issue?

At least doc doesn't say `reset()` is required
https://camel.apache.org/manual/latest/stream-caching.html

But I found the doc says
`Stream caching is from Camel 2.12 onwards intended to be configured using org.apache.camel.spi.StreamCachingStrategy. The old kind of configuration using properties on the CamelContext has been marked as deprecated.`
One guess is that the old configuration might have been removed in camel3? Does syndesis configure it via `StreamCachingStrategy`?

Delawen:
Syndesis does use the StreamCachingStrategy indeed. Why it did work before, I don't know. But this `payload` function is already on AtlasMap and, afaik, Syndesis does never get the reference to that message instances ever again. So it falls on AtlasMap and/or Camel to make sure we can read it more than once.

Delawen:
Checking the reset documentation, I understand you have to explicitly call it to be able to read more than once the same stream:

https://github.com/apache/camel/blob/master/core/camel-api/src/main/java/org/apache/camel/StreamCache.java#L42

Delawen:
Or maybe we can always copy the message before read:
https://github.com/apache/camel/blob/master/core/camel-api/src/main/java/org/apache/camel/StreamCache.java#L32-L33

igarashitm:
@oscerd do we need to explicitly invoke `reset()` to make camel StreamCache re-readable in camel3? I think it was completely transparent in camel2 so that StreamCache implementation detail doesn't leak into endpoint.

Delawen:
Worst case scenario if we don't want to drag this into AtlasMap: we wrap that StreamCache on Syndesis side to call reset after each read.

But this looked more transparent to me.

oscerd:
> @oscerd do we need to explicitly invoke `reset()` to make camel StreamCache re-readable in camel3? I think it was completely transparent in camel2 so that StreamCache implementation detail doesn't leak into endpoint.

There is a MessageHelper method for this purpose, you can use the resetStreamCache method.

igarashitm:
@oscerd thanks, OK so it was changed by design. Then I'm OK to have this change in AtlasMap. But @Delawen please keep in mind that in Syndesis, any step that leverages Data Buckets (or anything that access old message later) will have to do the same.

igarashitm:
Ugh forgot to ask @Delawen, can you take one more step and use `MessageHelper.resetStreamCache()` instead like @oscerd suggested?

Delawen:
Sure, will correct this :)

PR #1982, opened 2020-04-28
fix: Fix expression box focus issues - used useEffect to isolate expr…

pleacu:
…ession box.

Still issues with caret positioning - will address later.
Fixes: #1957

### DIFF
--- a/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
+++ b/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
@@ -31,6 +31,7 @@ import { MappingSerializer } from '../utils/mapping-serializer';
 import { DataMapperUtil } from '../common/data-mapper-util';
 import { PaddingField } from '../models/document-definition.model';
 import { LookupTableUtil } from '../utils/lookup-table-util';
+import { MappingUtil } from 'utils';
 
 /**
  * Handles mapping updates. It restores mapping status from backend and reflect in UI,
@@ -592,6 +593,9 @@ export class MappingManagementService {
     this.cfg.mappings.activeMapping.transition.enableExpression
       = !this.cfg.mappings.activeMapping.transition.enableExpression;
     this.updateTransition(this.cfg.mappings.activeMapping);
+    if (this.cfg.mappings.activeMapping.transition.expression) {
+      this.cfg.mappings.activeMapping.transition.expression.expressionUpdatedSource.next();
+    }
   }
 
   /**
--- a/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
+++ b/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
@@ -31,7 +31,6 @@ import { MappingSerializer } from '../utils/mapping-serializer';
 import { DataMapperUtil } from '../common/data-mapper-util';
 import { PaddingField } from '../models/document-definition.model';
 import { LookupTableUtil } from '../utils/lookup-table-util';
-import { MappingUtil } from 'utils';
 
 /**
  * Handles mapping updates. It restores mapping status from backend and reflect in UI,
@@ -761,18 +760,18 @@ export class MappingManagementService {
    *
    * @param filter
    */
-  executeFieldSearch(configModel: ConfigModel, filter: string, isSource: boolean): any[] {
+  executeFieldSearch(configModel: ConfigModel, filter: string, isSource: boolean): string[][] {
     const activeMapping = configModel.mappings!.activeMapping!; // TODO: check this non null operator
-    const formattedFields: any[] = [];
+    const formattedFields: string[][] = [];
     let fields: Field[] = [];
     for (const docDef of configModel.getDocs(isSource)) {
       fields = fields.concat(docDef.getTerminalFields());
     }
     let documentName = '';
     let fieldCount = -1;
-    let formattedField = null;
 
     for (const field of fields) {
+      const formattedField: string[] = [""];
       let displayName = (field == null) ? '' : field.getFieldLabel(configModel.showTypes, true);
 
       if (filter == null || filter === '' || displayName.toLowerCase().indexOf(filter.toLowerCase()) !== -1) {
@@ -784,14 +783,17 @@ export class MappingManagementService {
             formattedFields.pop();
             continue;
           } else {
+            const documentField = [""];
             documentName = field.docDef.name;
-            formattedField = { 'field': null, 'displayName': documentName };
+            documentField[0] = documentName;
+            documentField[1] = "";
             fieldCount = 0;
-            formattedFields.push(formattedField);
+            formattedFields.push(documentField);
           }
         }
         displayName = DataMapperUtil.extractDisplayPath(field.path, 100);
-        formattedField = { 'field': field, 'displayName': displayName };
+        formattedField[0] = displayName;
+        formattedField[1] = field.path;
         fieldCount++;
         formattedFields.push(formattedField);
       }
--- a/ui-react/packages/atlasmap/src/Atlasmap/utils/ui.ts
+++ b/ui-react/packages/atlasmap/src/Atlasmap/utils/ui.ts
@@ -160,8 +160,8 @@ export function fromMappingDefinitionToIMappings(
 export function executeFieldSearch(
   searchFilter: string,
   isSource: boolean,
-): void {
-  initializationService.cfg.mappingService.executeFieldSearch(
+): any[] {
+  return initializationService.cfg.mappingService.executeFieldSearch(
     initializationService.cfg,
     searchFilter,
     isSource,
@@ -224,6 +224,7 @@ export function mappingExpressionInit() {
     mapping.transition.expression.setConfigModel(initializationService.cfg);
   }
   mapping.transition.expression.updateFieldReference(mapping);
+  initializationService.cfg.mappingService.mappingUpdatedSource.next();
 }
 export function mappingExpressionClearText(
   nodeId?: string,
@@ -252,13 +253,12 @@ export function mappingExpressionInsertText(
 }
 export function mappingExpressionObservable(): Observable<any> | null {
   if (
-    !initializationService.cfg.mappings ||
-    !initializationService.cfg.mappings!.activeMapping
+    !initializationService.cfg.mappings?.activeMapping?.transition?.expression
   ) {
     return null;
   }
-  return initializationService.cfg.mappings!.activeMapping.transition
-    .expression!.expressionUpdated$;
+  return initializationService.cfg.mappings.activeMapping.transition.expression
+    .expressionUpdated$;
 }
 export function mappingExpressionRemoveField(
   tokenPosition?: string,
--- a/ui-react/packages/atlasmap/src/Atlasmap/utils/ui.ts
+++ b/ui-react/packages/atlasmap/src/Atlasmap/utils/ui.ts
@@ -160,7 +160,7 @@ export function fromMappingDefinitionToIMappings(
 export function executeFieldSearch(
   searchFilter: string,
   isSource: boolean,
-): any[] {
+): string[][] {
   return initializationService.cfg.mappingService.executeFieldSearch(
     initializationService.cfg,
     searchFilter,
@@ -169,7 +169,7 @@ export function executeFieldSearch(
 }
 
 export function mappingExpressionAddField(
-  selectedField: any,
+  selectedField: string,
   newTextNode: any,
   atIndex: number,
   isTrailer: boolean,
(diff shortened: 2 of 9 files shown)


ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
```diff
@@ -31,6 +31,7 @@ import { MappingSerializer } from '../utils/mapping-serializer';
 import { DataMapperUtil } from '../common/data-mapper-util';
 import { PaddingField } from '../models/document-definition.model';
 import { LookupTableUtil } from '../utils/lookup-table-util';
+import { MappingUtil } from 'utils';
```
riccardo-forina, line 34: Better use relative paths. Unfortunately TS paths are not respected when building the package as an npm module (I found this the hard way).

ui-react/packages/atlasmap/src/Atlasmap/utils/ui.ts
```diff
@@ -160,8 +160,8 @@ export function fromMappingDefinitionToIMappings(
 export function executeFieldSearch(
   searchFilter: string,
   isSource: boolean,
-): void {
-  initializationService.cfg.mappingService.executeFieldSearch(
+): any[] {
```
riccardo-forina, line 163: Please never use `any`, it removes the benefit of using Typescript in first place. If we don't need the return of the `executeFieldSearch` method then we shouldn't return it.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -1,30 +1,82 @@
...
+  mappingExpressionInit,
+  mappingExpressionInsertText,
+  mappingExpressionObservable,
+  mappingExpressionRemoveField,
+  mappingExpression,
+  onConditionalMappingExpressionEnabled,
+  onToggleExpressionMode,
+  trailerId,
+}) => {
+  const [, setCondExprEnabled] = useState<boolean>();
```
riccardo-forina, line 32: Sorry, this usage of `useState` is wrong. It doesn't make sense in React to have a state and then never using it. Is this some way to force the refresh of the child components? If that's the case, then the child components should be refactored to follow React's flow.

ui-react/packages/atlasmap/src/Atlasmap/utils/ui.ts
pleacu: We do need the return of executeFieldSearch - It sends back an array of { displayName and a field node}.   I'm looking to see if I can send back just an array of strings.

ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts
pleacu: actrually didn't need MappingUtil - good catch.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
pleacu: agree - it wasnt' needed

pleacu: ended up needing this - see updates

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -1,30 +1,112 @@
...
+  mappingExpressionInit,
+  mappingExpressionInsertText,
+  mappingExpressionObservable,
+  mappingExpressionRemoveField,
+  mappingExpression,
+  onConditionalMappingExpressionEnabled,
+  onToggleExpressionMode,
+  trailerId,
+}) => {
+  const [conditionExpressionEnabled, setCondExprEnabled] = useState<boolean>(
```
riccardo-forina, line 60: If I follow the trail I see that the callback will eventually [set this flag](https://github.com/atlasmap/atlasmap/blob/ea7f43e4abec4684cc83718102b6b17af16adada/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts#L592). With this useState we are creating a copy of that flag. Since we want to have a single source of truth and since the expression box engine is provided by the RxJS code, we should make that flag available through the `AtlasmapProvider` and feed its value to this component.

ui-react/packages/atlasmap/src/UI/ExpressionContent.tsx
```diff
@@ -1,51 +1,27 @@
-import { FunctionComponent } from "react";
+import { FunctionComponent, useEffect, useCallback } from "react";
 import React from "react";
 import { Tooltip, Form, FormGroup } from "@patternfly/react-core";
-import { Observable } from "rxjs";
+import { Subscription } from "rxjs";
```
riccardo-forina, line 4: Picking this line but commenting on the whole file. This file is heavily coupled with the RxJS store, which is bad because it makes testing this component from a pure UI perspective very difficult. 

Since this component is in the `UI` folder it shouldn't really know or care about how things work under the hood. It should care about how things are displayed to the user and how the user can interact with the component, while exposing the necessary callbacks for the consumer of this component to do glue it to the business logic.

An handy way to understand if a component is too coupled is writing a test and/or a story for it. If you have to bring into the story too many bits it's a signal that the component is too coupled.

pleacu: What's evident about this file is that it has little to do with the UI - it's just a DIV in a react form that calls a typeahead search select.  It is likely difficult to test - it was difficult to write - but it doesn't violate any separation of duties set out in the architecture and it is providing a very specific purpose.  I think it's isolated well and attempting to refactor it into multiple abstractions to satisfy testability is wrong.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
pleacu: is this issue significant enough to block merging or can this be addressed is a separate issue?  There are a few things left to be fixed (d&d, caret positioning, etc).  I'd like to get the bulk checked in before it goes stale.

pleacu: Ok - I've addressed this issue with the latest change.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -1,30 +1,106 @@
...
+  mappingExpressionInit,
+  mappingExpressionInsertText,
+  mappingExpressionObservable,
+  mappingExpressionRemoveField,
+  mappingExpression,
+  onConditionalMappingExpressionEnabled,
+  onToggleExpressionMode,
+  trailerId,
+}) => {
+  function onToggle() {
```
riccardo-forina, line 60: This function does nothing useful, it's wrapping a callback without adding any value to it. Also the callback is already called at line #99, so there are two ways to do the same thing.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -1,30 +1,106 @@
...
+          enableFlip={true}
+          entryDelay={1000}
+          position={"left"}
+        >
+          <Button
+            variant={"plain"}
+            aria-label="Enable/ Disable conditional mapping expression"
+            tabIndex={-1}
+            onClick={onToggle}
+            disabled={!onConditionalMappingExpressionEnabled()}
```
riccardo-forina, line 78: From this component standpoint, a `boolean` is enough to know if the component needs to be disabled or not. If calling a function to get the boolean is required, it should be done on the caller side.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -1,30 +1,106 @@
...
+    selectedField: string,
+    newTextNode: any,
+    atIndex: number,
+    isTrailer: boolean,
+  ) => void;
+  mappingExpressionClearText: (
+    nodeId?: string,
+    startOffset?: number,
+    endOffset?: number,
+  ) => any;
```
riccardo-forina, line 26: Never use `any`

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -1,30 +1,106 @@
...
+  isMappingExpressionEmpty: boolean;
+  mappingExpressionInit: () => void;
+  mappingExpressionInsertText: (
+    str: string,
+    nodeId?: string | undefined,
+    offset?: number | undefined,
+  ) => void;
+  mappingExpressionObservable: () => any;
```
riccardo-forina, line 34: Never use `any`

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -1,30 +1,106 @@
...
+  Button,
+  Tooltip,
+} from "@patternfly/react-core";
+import { css, StyleSheet } from "@patternfly/react-styles";
+import { ExpressionContent } from "./ExpressionContent";
 
 const styles = StyleSheet.create({
   toolbarItem: { flex: 1 },
 });
+export interface IConditionalExpressionInputProps {
```
riccardo-forina, line 14: This interface should extend `IExpressionContentProps` to avoid re-declaring the same stuff twice.

ui-react/packages/atlasmap/src/UI/ExpressionContent.tsx
riccardo-forina: Then the UI component should be thin and expose the callbacks needed to wire it to the code that knows about the logic. Think about how e.g. the `select` html component works, it's generic and exposes callbacks you can use to do your own specialized work.

If this UI component should only show some text and open a drawer when a special key (`@`) is keyed down, it should do only that.

All the rest should stay away, in the `Atlasmap` folder I'd say. And IMHO there should be no need for these function to know about the input field or the html that's been used to render the widget. If the interface between the logic and the ui uses basic types like `string`, then they are decoupled from each other allowing us to change the UI in the future without worrying about breaking the logic.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
pleacu: yep - I've removed onToggleExpressionMode from the EclipseContent props

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
pleacu: The boolean that controls whether its disabled is part of the mapping.  It needs a function call - this is the correct way to do this.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
pleacu: will address in a new issue

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
pleacu: will address in a new issue

pleacu:
@riccardo-forina - this addresses all issues except the mappingExpressionClearText use of any.  Would still like to work on that as a separate issue to get exposure for what's here.

ui-react/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
riccardo-forina: `onConditionalMappingExpressionEnabled` is being called twice at each render. One here, the second time [here](https://github.com/atlasmap/atlasmap/pull/1982/files/3b94234a515c59699149785e94a3e251e302f0c4#diff-e5c48eedbcf5dc5fab5b849bda1e04e4R88). 

Ultimately, [this code](https://github.com/atlasmap/atlasmap/blob/d7f50903a3535a6cb8dfea91c6e3a6dbe74e6219/ui-react/packages/atlasmap-core/src/services/mapping-management.service.ts#L569) is being called, which does nothing but returning the value of a boolean.

I stand my point, this doesn't have to be a function because. Even more so, since it returns a value that deeply nested into the state used by RxJS, it's completely opaque to React's lifecycle. A change on that boolean will not trigger a render on the React side and will lead to a stale UI and bugs. I think right now it works only because we are registered to a number of observables on the `AtlasmapProvider` so that we "catch all changes" so to speak.

pleacu:
Ok - will merge and generate 2 issues.  One for any 'any' referenced in ExpressionComntent and one for making onConditionalMappingExpressionEnabled() a reference to the actual boolean instead of a fcn.
thkx for your patience...

Issue #1892, opened 2020-03-20
UI_REACT: How to delete mapping block which doesn't have any related mapping lines?


emilywwj:
I think the isolated mapping block cannot be deleted. Because when I try to edit a mapping block and click on the trash can button in order to delete, it doesn't work.

igarashitm:
@atlasmap/ui could anybody evaluate if it's a bug or not and put labels?

pleacu:
![sc1](https://user-images.githubusercontent.com/2981421/77179378-caeb0b80-6a9e-11ea-888b-664233e49514.png)

Verified - bug.

jpav:
Fixed by other changes to code

PR #1984, opened 2020-04-29
fix: UI_REACT: How to delete mapping block which doesn't have any related mapping lines?

jpav:
Fixes: #1892 

Technically, the user could click the edit details button, which then has a delete mapping button, but it certainly wasn't intuitive how to do this.

### DIFF
--- a/ui-react/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
+++ b/ui-react/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
@@ -365,11 +365,21 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
     ],
   );
 
+  const handleRemoveMapping = useCallback(
+    () =>
+      handlers.onDeleteMapping(() => {
+        removeMapping(selectedMapping!.mapping);
+        deselectMapping();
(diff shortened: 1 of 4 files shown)


ui-react/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
```diff
@@ -365,19 +365,29 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
...
+      }),
+    [handlers, deselectMapping, removeMapping, selectedMapping],
+  );
+
   const mappingEvents: IMappingDocumentEvents = useMemo(
     () => ({
       onSelectMapping: selectMapping,
       onDeselectMapping: deselectMapping,
       onEditMapping: () => void 0,
+      onDeleteMapping: handleRemoveMapping,
```
riccardo-forina, line 382: This will remove the selected mapping, not the mapping the user asked to be deleted.
Either change the `handleRemoveMapping` to accept the mapping to be deleted as an argument and change the other place where it's used, or create a new handler.

jpav:
The need for this fix has been obviated by the latest changes.

Issue #1986, opened 2020-04-29
UI_REACT: remove any 'any' reference from ExpressionContent.


pleacu:
Generate an ITextNode at the Atlasmap level:

export interface ITextNode {
  uuid: string;
  str: string;
}

Use it as the node type for mappingExpressionClearText

PR #1988, opened 2020-04-30
fix: Remove opaque any refs in the expression box.

pleacu:
Fixes: #1986

### DIFF
--- a/ui-react/packages/atlasmap/src/UI/ExpressionContent.tsx
+++ b/ui-react/packages/atlasmap/src/UI/ExpressionContent.tsx
@@ -1,8 +1,9 @@
 import { FunctionComponent, useEffect, useCallback } from "react";
 import React from "react";
 import { Tooltip, Form, FormGroup } from "@patternfly/react-core";
-import { Subscription } from "rxjs";
+import { Subscription, Observable } from "rxjs";
 import { ExpressionFieldSearch } from "./ExpressionFieldSearch";
+import { ITextNode, IExpressionUpdatedEvent } from "../../src/Atlasmap";
 
 let atIndex = -1;
 let atContainer: Node | undefined;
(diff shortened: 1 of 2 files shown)


ui-react/packages/atlasmap/src/UI/ExpressionContent.tsx
```diff
@@ -1,8 +1,9 @@
 import { FunctionComponent, useEffect, useCallback } from "react";
 import React from "react";
 import { Tooltip, Form, FormGroup } from "@patternfly/react-core";
-import { Subscription } from "rxjs";
+import { Subscription, Observable } from "rxjs";
 import { ExpressionFieldSearch } from "./ExpressionFieldSearch";
+import { ITextNode, IExpressionUpdatedEvent } from "../../src/Atlasmap";
```
riccardo-forina, line 6: Can't import from `Atlasmap` from the `UI` folder. I mean, you can, but it's reversed. Components in the `UI` are meant to be decoupled from `Atlasmap` and reusable.

You can fix this by having the models live here instead than in the `AtlasmapProvider`, since this is the only place that uses them.

riccardo-forina: Also no need to go that deep with the import, `../Atlasmap` would be enough.

ui-react/packages/atlasmap/src/UI/ExpressionContent.tsx
```diff
@@ -245,7 +246,7 @@ export const ExpressionContent: FunctionComponent<IExpressionContentProps> = ({
    *
    * @param event - expression keyboard event
    */
-  function onKeyDown(event: any): void {
+  function onKeyDown(event: React.KeyboardEvent<HTMLDivElement>): void {
```
riccardo-forina, line 249: Better import the elements explicitly than referring to them from the default export. So in this case `import React, { KeyboardEvent } from "react"` and then use it directly `KeyboardEvent<HTMLDivElement>`.

This helps addressing a number of issues, some with Storybook and some with the tree shaking performed by the consumer of this library.

ui-react/packages/atlasmap/src/UI/ExpressionContent.tsx
pleacu: Essentially just move ITextNode and IExpressionUpdatedEvent interfaces to ExpressionContent since that's where they're needed?

Issue #2011, opened 2020-05-06
UI_REACT: Support custom classes for field actions and mapable fields


pleacu:
Implementation

PR #2014, opened 2020-05-06
fix: Support enable custom class button.

pleacu:
Fixes: #2011


ui-react/packages/atlasmap/src/Atlasmap/utils/document.ts
```diff
@@ -95,3 +100,155 @@ export async function importInstanceSchema(
...
+            scope: ErrorScope.APPLICATION,
+            type: ErrorType.INTERNAL,
+            object: error,
+          }),
+        );
+      }
+    });
+}
+
+interface ISelectValues1 {
```
riccardo-forina, line 222: I’m curious, why the 1 in the name?

ui-react/packages/atlasmap/src/UI/useDoubleSelectDialog.tsx
```diff
@@ -0,0 +1,179 @@
...
+  FormSelect,
+  FormSelectOption,
+  Label,
+} from "@patternfly/react-core";
+import React, { ReactPortal, useCallback, useRef, useState } from "react";
+import { createPortal } from "react-dom";
+import { css, StyleSheet } from "@patternfly/react-styles";
+import { CancelCallback } from "./useSingleInputDialog";
+
+interface ISelectValues1 {
```
riccardo-forina, line 13: Maybe export this and import it where you use it instead of redeclaring it?

ui-react/packages/atlasmap/src/UI/useDoubleSelectDialog.tsx
```diff
@@ -0,0 +1,179 @@
...
+}
+
+export type ConfirmDoubleSelectCallback = (
+  value1: string,
+  value2: string,
+) => void;
+
+export interface IUseDoubleSelectDialogArgs {
+  title: string;
+  selectLabel1: string;
```
riccardo-forina, line 25: For another day: I’m not a fan of this style of hooks where they are kinda generic but in reality they are specialized. We should simply have a confirm dialog that accepts a child element and provide as a child whatever logic we need - a form, a video, anything.

Also there is an issue with all the refs and states in use that make the performance of the dialog very bad. Try typing something eg. in the create constant dialog and you should see very heavy lag. That shouldn’t happen.

ui-react/packages/atlasmap/src/UI/useInputTextSelectDialog.tsx
```diff
@@ -149,61 +157,60 @@ export function useInputTextSelectDialog({
...
       {textLabel1 && (
-        <InputGroup className={css(styles.iGroup)}>
-          <InputGroupText className={css(styles.iGroupTextLabel)}>
-            {textLabel1}
-          </InputGroupText>
+        <>
           <TextInput
-            key={"text-input-value1"}
+            className={css(styles.iGroupTextLabel)}
+            key={"itsd-text-input-value1"}
```
riccardo-forina, line 163: Why the key? It should be used only when creating elements dynamically like with a map over an array

ui-react/packages/atlasmap/src/UI/useInputTextSelectDialog.tsx
```diff
@@ -149,61 +157,60 @@ export function useInputTextSelectDialog({
...
       {textLabel2.length > 0 && (
-        <InputGroup className={css(styles.iGroup)}>
-          <InputGroupText className={css(styles.iGroupTextLabel)}>
-            {textLabel2}
-          </InputGroupText>
+        <>
           <TextInput
-            key={"text-input-value2"}
+            className={css(styles.iGroupTextLabel)}
+            key={"itsd-text-input-value2"}
```
riccardo-forina, line 180: Same as above

ui-react/packages/atlasmap/src/Views/ColumnMapperView/Actions/EnableJavaClassAction.tsx
```diff
@@ -1,33 +1,38 @@
 import React, { FunctionComponent } from "react";
-
 import { Button, Tooltip } from "@patternfly/react-core";
 import { AddCircleOIcon } from "@patternfly/react-icons";
 
 export interface IEnableJavaClassActionProps {
-  id: string;
-  onClick: () => void;
+  isSource: boolean;
```
riccardo-forina, line 6: No need to pass this function this info since it’s used just to pass it back again to the onCustomClassSearch callback. Just create the appropriate closure where this is used

ui-react/packages/atlasmap/src/Views/ColumnMapperView/Actions/EnableJavaClassAction.tsx
```diff
@@ -1,33 +1,38 @@
...
+  onCustomClassSearch: (isSource: boolean) => void;
```
riccardo-forina, line 7: Linked to the above, this should be just `() =>. void;`

ui-react/packages/atlasmap/src/Views/ColumnMapperView/Columns/SourcesColumn.tsx
```diff
@@ -121,8 +121,8 @@ export const SourcesColumn: FunctionComponent<
             key={"import"}
           />,
           <EnableJavaClassAction
-            id="Source"
-            onClick={onEnableJavaClasses}
+            isSource={true}
+            onCustomClassSearch={onCustomClassSearch}
```
riccardo-forina, line 125: And here `() => onCustomClassSearch(true)`

ui-react/packages/atlasmap/src/Views/ColumnMapperView/Columns/TargetsColumn.tsx
```diff
@@ -87,8 +87,8 @@ export const TargetsColumn: FunctionComponent<
             key={"import"}
           />,
           <EnableJavaClassAction
-            id="Target"
-            onClick={onEnableJavaClasses}
+            isSource={false}
+            onCustomClassSearch={onCustomClassSearch}
```
riccardo-forina, line 91: And here `() => onCustomClassSearch(false)`

ui-react/packages/atlasmap/src/Atlasmap/utils/document.ts
pleacu: Sorry - I should have been clearer - what I really want is for a FormSelect to list all of the class names defined by the user's imported JARs,  That will take more work on the server side - for now we're doing what the angular Ui did and having the user type the full class.  This interface is used by getCustomClassCandidates() which is currently not called.  It will be when I implement the select pulldown later on.

ui-react/packages/atlasmap/src/UI/useDoubleSelectDialog.tsx
pleacu: will do when the class name select is implemented.

ui-react/packages/atlasmap/src/UI/useDoubleSelectDialog.tsx
pleacu: This isn't currently used either.  Wanted to keep this around for teh user class select implementation.

ui-react/packages/atlasmap/src/UI/useInputTextSelectDialog.tsx
pleacu: corrected.

ui-react/packages/atlasmap/src/UI/useInputTextSelectDialog.tsx
pleacu: corrected

ui-react/packages/atlasmap/src/Atlasmap/utils/document.ts
pleacu: corrected - I took out my partial fix - will keep it on a dev branch

ui-react/packages/atlasmap/src/UI/useDoubleSelectDialog.tsx
pleacu: corrected - file removed.

ui-react/packages/atlasmap/src/UI/useDoubleSelectDialog.tsx
pleacu: removed

PR #2019, opened 2020-05-07
Don't serialize expressions of they don't exist.

pleacu:
fixes: ENTESB-13705


pleacu:
@tplevko - can you test this out and approve if its okay?

riccardo-forina:
Let's merge this and in case reopen it again.

PR #2020, opened 2020-05-07
Fixes #2018 Add CSV module

### LINKED ISSUE #2018: Add CSV module
Support CSV files using Apache commons-csv library

xref: https://issues.redhat.com/browse/ENTESB-13746

rkorytkowski:


rkorytkowski:
@igarashitm please have a look. 

From what I see I still need to add a component test for Camel and Camel3. Any important tests I missed in terms of adding a module?

Does it make sense to enable it in standalone?

I will be also adding more business logic tests...

lib/modules/dfdl/module/src/main/java/io/atlasmap/dfdl/module/DfdlModule.java
```diff
@@ -1,132 +0,0 @@
-/**
```
igarashitm, line 1: Removing DFDL module? it's there not only for CSV - CSV is just an example.

igarashitm:
@rkorytkowski other than DFDL module LGTM! It looks to me camel test is not mandatory as it doesn't change something specific to camel component. The itests will do the job enough I guess?
Also can you tweak commit message, to have `feat:` header and `Fixes: #2018` at second line?

lib/modules/dfdl/module/pom.xml
```diff
@@ -66,6 +66,12 @@
       <artifactId>atlas-core</artifactId>
       <classifier>tests</classifier>
     </dependency>
+    <dependency>
```
igarashitm, line 69: remove this dependency?

rkorytkowski:
@igarashitm, thanks. Updated the commit message.

igarashitm:
Sorry to be scattering, we'd need to add csv module dependency to the camel/camel3/standalone/standalone-reactui POM.

rkorytkowski:
Good catch. Fixed camel/camel3/standalone/standalone-reactui POM.

PR #2025, opened 2020-05-08
feat: Add CSV Module, include basic standalone UI

rkorytkowski:
Fixes: #2018

It adds a very basic standalone UI support, which basically allows for uploading a csv instance in default formatting with a header in the first record and a comma delimiter. 

In the next iteration we will want to add a dialog, which allows the user to specify additional parameters such as delimiter, comment mark, etc. before uploading a file.


rkorytkowski:
@igarashitm, done! 

It's quite a bit of refactoring in UI to add a dialog, which lets the user enter additional parameters so I didn't even risk doing that as it would require extensive testing.

rkorytkowski:
Fixed the failing test... I'm having an eye on this until all is green.

rkorytkowski:
@igarashitm, all good now.

ui-react/packages/atlasmap-core/src/services/document-management.service.ts
```diff
@@ -287,6 +290,11 @@ export class DocumentManagementService {
           inspectionType, isSource);
         break;
 
+      case DocumentType.CSV:
+        this.cfg.initializationService.initializeUserDoc(fileText, userFile, DocumentType.CSV,
```
riccardo-forina, line 294: `await` here, I just discovered we need that for this to work

ui-react/packages/atlasmap/src/Atlasmap/AtlasmapProvider.tsx
```diff
@@ -104,6 +104,7 @@ export interface IAtlasmapProviderProps {
   baseJavaInspectionServiceUrl: string;
   baseXMLInspectionServiceUrl: string;
   baseJSONInspectionServiceUrl: string;
+  baseCSVInspectionServiceUrl: string;
```
riccardo-forina, line 107: So this is a breaking change? We'll have to update Syndesis where we consume this to provide the right url then.

riccardo-forina:
Ah this got merged already. Ok. https://github.com/atlasmap/atlasmap/pull/2025#discussion_r422132410 this needs to be fixed else the document will not be persisted between refreshes (see https://issues.redhat.com/browse/ENTESB-13743). @rkorytkowski do you want to do the change? Else I can do it.

ui-react/packages/atlasmap/src/Atlasmap/AtlasmapProvider.tsx
igarashitm: As there's no connector leverages CSV support ATM, I think either way works in Syndesis - skip adding or internally add `DataShape.Kind = CSV` in UI and add CSV REST endpoint in server for future use. It's just an addition, won't be visible to Syndesis user unless any connector specify CSV kind in its DataShape. Maybe user can develop an extension leveraging it?
@gaughan WDYT?

riccardo-forina: My comment was more about the typescript type. This way the url is required, and the Atlasmap component can't be consumed in Syndesis without providing one.

If it becomes optional, we have to deal with potentially undefined urls where we use this.

igarashitm: Let's provide the URL then, it's just next to the others

ui-react/packages/atlasmap-core/src/services/document-management.service.ts
rkorytkowski: Thanks! Good catch. Addressed in https://github.com/atlasmap/atlasmap/pull/2041

ui-react/packages/atlasmap/src/Atlasmap/AtlasmapProvider.tsx
rkorytkowski: @riccardo-forina, I'm having trouble finding the right place to set it in Syndesis. https://github.com/syndesisio/syndesis/search?q=jsonInspectionServiceUrl&unscoped_q=jsonInspectionServiceUrl gives me nothing...

riccardo-forina: Not your fault, [we haven't merged the PR yet](https://github.com/syndesisio/syndesis/pull/8316). I'll add the param when needed, don't worry!

rkorytkowski: I see! Thanks!

Issue #2012, opened 2020-05-06
UI: propertyDoc and targetDocs `fields` and `allFields` are out of sync


riccardo-forina:
Using the `mockdoc.adm` test file, properties and constants are listed differently in the `fields` and `allFields` collections. See screen:

![image](https://user-images.githubusercontent.com/966316/81176318-4ba98c80-8fa5-11ea-8334-aa705b5aba7b.png)

This causes some issues on the React UI like not showing the mapping preview for mapped constants/fields because the `field` collection is used to feed the ui, and in that collections the properties/constants are flagged as not part of a mapping.

PR #2044, opened 2020-05-11
fix: Avoid duplication of property/constant fields nodes.

pleacu:
Fixes: #2012

If a constant or property appeared as part of a mapping it became
duplicated.

Business logic fix.


pleacu:
@igarashitm - Hey Tomo - I know you've been busy - pls take a look when its convenient.

Issue #2067, opened 2020-05-13
UI_REACT: Add d&d support to expression box


pleacu:
Drag a field from the Source panel and drop it into the expression box at a specific location.

pleacu:
We are not planning on supporting d&d to the expression box.

PR #2070, opened 2020-05-13
fix: Make expression box tooltip disappear after 5 secs, don't drop e…

pleacu:
…xpression search dropdown if out for less than 750ms.

Fixes: #2067

** Not required for todays release

### DIFF
--- a/ui/packages/atlasmap/src/UI/ExpressionContent.tsx
+++ b/ui/packages/atlasmap/src/UI/ExpressionContent.tsx
@@ -421,6 +421,7 @@ export const ExpressionContent: FunctionComponent<IExpressionContentProps> = ({
     setTooltipVisible(true);
     tooltipTimeOut = setTimeout(() => {
       setTooltipVisible(false);
+      tooltipTimeOut = null;
     }, 5000);
   }
 
@@ -451,9 +452,14 @@ export const ExpressionContent: FunctionComponent<IExpressionContentProps> = ({
     if (expressionUpdatedSubscription) {
       expressionUpdatedSubscription.unsubscribe();
     }
+    if (tooltipTimeOut) {
+      clearTimeout(tooltipTimeOut);
+      tooltipTimeOut = null;
+    }
   };
 
   useEffect(() => {
+    tooltipTimeOut = null;
     initMappingExpression();
     return () => uninitializeMappingExpression();
   }, [initMappingExpression, mappingExpression]);
--- a/ui/packages/atlasmap/src/UI/ExpressionContent.tsx
+++ b/ui/packages/atlasmap/src/UI/ExpressionContent.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, useEffect, useCallback } from "react";
+import { FunctionComponent, useEffect, useCallback, useState } from "react";
 import React, { KeyboardEvent } from "react";
 import { Tooltip, Form, FormGroup } from "@patternfly/react-core";
 import { Subscription, Observable } from "rxjs";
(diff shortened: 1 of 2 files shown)


pleacu:
> I think you might want to read this and revisit the way you handled the timeout https://overreacted.io/making-setinterval-declarative-with-react-hooks/

I guess I can't see how adding a new FunctionComponent containing the Tooltip fragment with useEffects is better.  It's doing exactly what I want it to do now.  I guess I can't figure a way of extracting "useInterval()" out as a callable function from mouseEnter() in ExpressionContent.

riccardo-forina:
Whoa one thing at a time. Let’s focus on what I pointed out, which is about how you handled setTimeout. 

From what I see there is a single global variable to handle the timeout ref: what if there are multiple instance of the component?
I understand it can’t happen in the app, but it could in a story and it’s generally a wrong approach, so it should be fixed.

Then I see setTimeout performed on mouse enter events, but where is the cleanup for the eventual previously set timeout?

In general what I see from the code is that you are not embracing the React flow, and this leads to bug and non idiomatic code. The article I linked talks about how to reason about this family of problems the react way.

https://overreacted.io/a-complete-guide-to-useeffect/

Another great resource is this article

ui/packages/atlasmap/src/UI/ExpressionContent.tsx
```diff
@@ -421,6 +421,7 @@ export const ExpressionContent: FunctionComponent<IExpressionContentProps> = ({
     setTooltipVisible(true);
     tooltipTimeOut = setTimeout(() => {
       setTooltipVisible(false);
```
riccardo-forina, line 423: What happens if the component gets removed before this callback gets fired?

ui/packages/atlasmap/src/UI/ExpressionFieldSearch.tsx
```diff
@@ -90,34 +90,21 @@ export const ExpressionFieldSearch: FunctionComponent<IExpressionFieldSearchProp
...
 
-  function onMouseLeave(_event: any): void {
-    // mouseOverTimeOut = setTimeout(() => {
-    clearSearchMode(true);
-    candidateSrcElement = null;
-    candidateIndex = 0;
-    // }, 750);
+  function onMouseLeave(_event: React.MouseEvent<HTMLDivElement>): void {
+    mouseOverTimeOut = setTimeout(() => {
+      clearSearchMode(true);
```
riccardo-forina, line 104: Same here.

ui/packages/atlasmap/src/UI/ExpressionContent.tsx
pleacu: Then uninitializeMappingExpression() will get called on the unmount and cleartimeout() is called right?

riccardo-forina:
Oh right I didn’t notice that. But that’s my point, the code is a bit all over the place and it’s really hard to grasp what’s going on and to be sure that there are no lose treads. It’s better when features are kept small and self contained

pleacu:
> Oh right I didn’t notice that. But that’s my point, the code is a bit all over the place and it’s really hard to grasp what’s going on and to be sure that there are no lose treads. It’s better when features are kept small and self contained

I agree - but in this case the expression content is confined and the expression search is confined.  The useEffect is employed correctly (I believe) and the code works.  I'd rather not refactor the code at this moment - so does this pass as ok to push?

gashcrumb:
> I'd rather not refactor the code at this moment - so does this pass as ok to push?

Why?  What's the rush?  If not now, when?

pleacu:
> > I'd rather not refactor the code at this moment - so does this pass as ok to push?
> 
> Why? What's the rush? If not now, when?

1.  Have you read the code in question?
2.  It's about 500 lines in total and I don't think it needs to be refactored.
3.  We're 2 weeks to blocker only freeze so now's not the time.

gashcrumb:
Yep, I did look.  Since you've asked I also find it hard to follow whats
going on, and generally always think 2 or 3 times before pulling out the
setTimeout sledgehammer.  I mean, it's feedback, take it or leave it, but
if debt is added to the code eventually it needs to get repaid.

Issue #1195, opened 2019-08-30
Migrate Angular UI unit tests to React


igarashitm:
We should migrate/add UTs with using Jest, hopefully UTs for services can be migrated.

PR #2076, opened 2020-05-15
chore: Migrate Angular UI unit tests to React

igarashitm:
Fixes: #1195


ui/packages/atlasmap-standalone/src/App.test.tsx
```diff
@@ -1,9 +1,29 @@
...
 import ReactDOM from 'react-dom';
 import App from './App';
+import { AtlasmapProvider } from '@atlasmap/atlasmap';
 
-it('renders without crashing', () => {
-  const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
-  ReactDOM.unmountComponentAtNode(div);
-});
+describe('AtlasMap standalone', () => {
```
igarashitm, line 6: Hi @riccardo-forina, I managed to get tests passing other than this one. This one fails with weird error `Target container is not a DOM element.`, do you see what's going on?

riccardo-forina: Hey great effort! For that check this out, we need to enrich the test document with a static doc with an ID that will be used by the app to inject the dialog:    https://github.com/syndesisio/syndesis/blob/c815aebc1931729a7377a15a3c35b75eb052372a/app/ui-react/packages/atlasmap-adapter/tests/DataMapperAdapter.spec.tsx#L46

igarashitm: thanks!

igarashitm: Turned out it was looking for a 'modals' div in underlying component. The standalone `App` is OK being empty.

igarashitm:
CircleCI build fails, the `yarn test` process is terminated with SIGKILL. Possibly running out of memory.

igarashitm:
OK... `--runInBand` solved exit 137 problem in UI atlasmap-core test.

Issue #2072, opened 2020-05-15
UI_REACT: Add challenge modal and mapping changes for colelctions w/ expr disable.


pleacu:
![sc1](https://user-images.githubusercontent.com/2981421/82076657-2e16aa00-96ac-11ea-99f9-e59202eeb72b.png)

pleacu:
May be related to https://issues.redhat.com/browse/ENTESB-13801

PR #2078, opened 2020-05-15
fix: Toggling the expression icon off with source collection fields b…

pleacu:
…eing present in the mapping requires a challenge before removing the sourcefields in the mapping.

Fixes: #2072


ui/packages/atlasmap/src/UI/ConditionalExpressionInput.tsx
```diff
@@ -32,11 +33,12 @@ export const ConditionalExpressionInput: FunctionComponent<IConditionalExpressio
   mappingExpressionRemoveField,
   mappingExpression,
   onConditionalMappingExpressionEnabled,
+  onMappingHasSourceCollection,
   onToggleExpressionMode,
   trailerId,
 }) => {
   function onToggle() {
-    onToggleExpressionMode();
+    onToggleExpressionMode(onMappingHasSourceCollection());
```
riccardo-forina, line 41: `onMappingHasSourceCollection` is an implementation details that would be better handled inside the `onToggleExpressionMode` function. Doing this would also remove a prop from this component.

ui/packages/atlasmap/src/Atlasmap/utils/ui.ts
```diff
@@ -252,6 +252,10 @@ export function executeFieldSearch(
   );
 }
 
+export function mappingHasSourceCollection(): boolean {
+  return initializationService.cfg.mappingService.willClearOutSourceFieldsOnTogglingExpression();
```
riccardo-forina, line 256: This feels like it could be resolved to be a boolean returned by the `useAtlasmap` hook, reducing even further the need for the ui components to know about how the core works

pleacu: done.

Issue #2073, opened 2020-05-15
UI_REACT: Disable multiplicity trans dropdown if expression enabled


pleacu:
if expression is enabled, multiplicity transformation dropdown would have to be disabled

Issue #1987, opened 2020-04-29
UI_REACT: make a onConditionalMappingExpressionEnabled a var ref


pleacu:
onConditionalMappingExpressionEnabled is a function - it doesn't need to be - change to just a va rref

PR #2084, opened 2020-05-19
fix: Disable multiplicity trans dropdown if expression enabled, make …

pleacu:
…onConditionalMappingExpressionEnabled a var ref

Fixes: #2073
Fixes: #1987

### DIFF
--- a/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
@@ -68,7 +68,7 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
     mappingExpressionInsertText,
     mappingExpressionObservable,
     mappingExpressionRemoveField,
-    onConditionalMappingExpressionEnabled,
+    mappingExpressionEnabled,
     isMappingExpressionEmpty,
     trailerId,
     isFieldAddableToSelection,
@@ -112,13 +112,11 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
         mappingExpressionAddField={mappingExpressionAddField}
         mappingExpressionClearText={mappingExpressionClearText}
         isMappingExpressionEmpty={isMappingExpressionEmpty}
+        mappingExpressionEnabled={mappingExpressionEnabled}
         mappingExpressionInit={mappingExpressionInit}
         mappingExpressionInsertText={mappingExpressionInsertText}
         mappingExpressionObservable={mappingExpressionObservable}
         mappingExpressionRemoveField={mappingExpressionRemoveField}
-        onConditionalMappingExpressionEnabled={
-          onConditionalMappingExpressionEnabled
-        }
         onToggleExpressionMode={handlers.onToggleExpressionMode}
         trailerId={trailerId}
       />
@@ -292,6 +290,7 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
   ]);
 
   const renderSidebar = useSidebar({
+    mappingExpressionEnabled: mappingExpressionEnabled,
     onRemoveMapping: handlers.onDeleteSelectedMapping,
   });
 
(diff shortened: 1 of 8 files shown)


ui/packages/atlasmap/src/UI/MappingTransformation.tsx
```diff
@@ -46,6 +46,7 @@ export interface IMappingTransformationProps {
   name: string;
   transformationsOptions: ITransformationSelectOption[];
   transformationsArguments?: ITransformationArgument[];
+  mappingExpressionEnabled: boolean;
```
riccardo-forina, line 49: I suggest to change this to be something like `disableTransformations` to reflect what needs to be done instead of why.

ui/packages/atlasmap/src/Views/MappingDetailsView.tsx
```diff
@@ -30,6 +30,7 @@ export interface IMappingDetailsViewProps {
   sourceTransformationsOptions: { name: string; value: string }[];
   targetTransformationsOptions: { name: string; value: string }[];
   onClose: () => void;
+  mappingExpressionEnabled: boolean;
```
riccardo-forina, line 33: Same as above.

ui/packages/atlasmap/src/Atlasmap/AtlasmapProvider.tsx
```diff
@@ -512,7 +511,7 @@ export function useAtlasmap() {
     mappingExpressionObservable,
     mappingExpressionRemoveField,
     mappingHasSourceCollection,
-    onConditionalMappingExpressionEnabled,
+    mappingExpressionEnabled: initializationService.cfg.mappingService.conditionalMappingExpressionEnabled(),
```
riccardo-forina, line 514: 👍

ui/packages/atlasmap/src/UI/MappingTransformation.tsx
pleacu: I'm happy to change the name but I think it's important to retain that the multiplicity transformation enablement is tied to the mapping expression enablement.  so:
`isDisabled={mappingExpressionEnabled}` maks that clar in the code.

riccardo-forina: Yes that’s important indeed but the information will be retained on the outermost layer, the `MappingDetailsView`, where the `disableTransformation` prop is initially set

ui/packages/atlasmap/src/UI/MappingTransformation.stories.tsx
```diff
@@ -52,6 +52,7 @@ export const example = () => (
     onTransformationArgumentChange={action("onActionArgumentChange")}
     onTransformationChange={action("onActionChange")}
     onRemoveTransformation={action("onRemoveTransformation")}
+    disableTransformation={false}
```
riccardo-forina, line 55: It would be nice to use the boolean knob here to see on storybook how the component changes with the different values

ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
```diff
@@ -292,6 +290,7 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
   ]);
 
   const renderSidebar = useSidebar({
+    mappingExpressionEnabled: mappingExpressionEnabled,
```
riccardo-forina, line 293: Sorry didn’t notice this before. You can get  this directly in the `useSidebar` hook, which uses `useAtlasmap`. No need to make it a prop.

pleacu: ah - good catch - see if this push is okay

ui/packages/atlasmap/src/UI/MappingTransformation.stories.tsx
```diff
@@ -63,5 +64,6 @@ export const nonRemovable = () => (
     noPaddings={boolean("Hide paddings", true)}
     onTransformationArgumentChange={action("onActionArgumentChange")}
     onTransformationChange={action("onActionChange")}
+    disableTransformation={false}
```
riccardo-forina, line 67: Since there is no knob here I can’t really tell from the preview on Netlify, but shouldn’t also text inputs be disabled when this is true?

pleacu: yes - you're right.  Do you have an example for a storybook example with a knob?  Is it just an argument to the example?

riccardo-forina: Check line #64

Issue #2077, opened 2020-05-15
Average(intArray) expression mapping doesn't work


igarashitm:
xref: https://issues.redhat.com/browse/ENTESB-13801
```
java.lang.ClassCastException: java.lang.Double cannot be cast to io.atlasmap.v2.Field
	at io.atlasmap.core.DefaultAtlasExpressionProcessor.processExpression(DefaultAtlasExpressionProcessor.java:73) [atlas-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
```

PR #2090, opened 2020-05-20
fix: Average(intArray) expression mapping doesn't work

igarashitm:
Fixes: #2077


igarashitm:
Hi @rkorytkowski can you take a look at this?
Mainly the problem is that expression could return `Field` as well as raw value. It's not beautiful for sure, we might have to refactor at some point.

Issue #2046, opened 2020-05-11
Filter file types for source and target


apupier:
when trying to `Import instance or schema file` for source or target, the file picker allows to choose all types files.
But not a lot is supported. i ti snot marked in Ui what is supported. In case i ti snot supported, there is no message to the user.

it would improve user experience to filter supported extensions.

PR #2099, opened 2020-05-22
fix(ui): modals refactoring

### LINKED ISSUE #2045: Filter file types to *.adm and *.jar for import
When clicking on import, the tooltip is mentioning that only `*.adm` and `*.jar` are supported. The file picker is letting users to choose all kind of files.
In case, a not supported format is selected, there is no error, just nothing happens.

it will improve user experience to have the file picker to filter based on extensions.

riccardo-forina:
Drop react-file-picker, which is discontinued and buggy, to its
successor react-sage. Fixes #2045 #2046

Make modals normal React components, without useRef tricks.

Fixes dialogs rendering performance.

Ref ENTESB-13643 ENTESB-13644


riccardo-forina:
@pleacu could you please verify if the custom class dialog is working ok? I'm not sure how to test it.

riccardo-forina:
By the way, all dialogs have an associated story now. Eg. https://deploy-preview-2099--atlasmap.netlify.app/?path=/story/ui-dialogs--export-catalog-dialog

igarashitm:
It seems e2e test is failing, the HTML is not found with xpath `//article[@aria-label='JSONSchemaSource'` after importing adm file.

igarashitm:
https://github.com/atlasmap/atlasmap/blob/master/standalone/src/test/java/io/atlasmap/standalone/E2ETest.java#L98

riccardo-forina:
Which makes sense, I removed them and forgot to add them back but a while ago. Why is it complaining now?

igarashitm:
hmm not sure, this test is passing on both of CircleCI and local with current upstream master.

igarashitm:
Is this PR adding a modal when importing adm file? then I guess it has to push OK button or something to proceed?

riccardo-forina:
🤦‍♂️ you are right of course, I added a confirmation dialog after selecting the file to warn about the risk of losing non exported changes. That would explain it

pleacu:
@riccardo-forina - custom class dialog looks a-ok

riccardo-forina:
Cool thanks!

igarashitm:
Adjusted e2e test

PR #2100, opened 2020-05-22
fix: Validation - Field mappings should not be empty

rkorytkowski:
Fixes: ENTESB-13797

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasValidationService.java
+++ b/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasValidationService.java
@@ -68,8 +68,8 @@ public class DefaultAtlasValidationService implements AtlasValidationService {
         MAPPINGS_NOT_NULL (() ->
             new NonNullValidator(ValidationScope.MAPPING, "Field mappings must not be null")
         ),
-        MAPPINGS_NOT_EMPTY (() ->
-            new NotEmptyValidator(ValidationScope.MAPPING, "Field mappings should not be empty")
+        MAPPINGS_EMPTY(() ->
+            new NotEmptyValidator(ValidationScope.MAPPING, "Field mappings is empty")
         ),
 
         COMBINE_INPUT_NOT_NULL (() ->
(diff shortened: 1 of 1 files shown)


ui/packages/atlasmap-core/src/services/mapping-management.service.ts
```diff
@@ -775,6 +775,10 @@ export class MappingManagementService {
           if (body && body.Validations && body.Validations.validation) {
             for (const validation of body.Validations.validation) {
               const level: ErrorLevel = validation.status;
+              if (level === ErrorLevel.WARN && validation.message === 'Field mappings should not be empty') {
```
igarashitm, line 778: Doesn't this warning make any sense? is it better to remove or drop to INFO level at the runtime side then?

rkorytkowski: Well, I think it's good to communicate no mappings have been defined if you try to run it. I'm fine with WARN or INFO level.

igarashitm: Then maybe drop to INFO, as UI notification shows only above WARN level.

rkorytkowski: Done!

riccardo-forina: > Then maybe drop to INFO, as UI notification shows only above WARN level.

Actually I think the only messages I’m skipping right now are the DEBUG ones. Info are displayed, one example is the notification after importing a document

igarashitm: Is it? hmm it would be still eyesore in UI then... @rkorytkowski sorry I misunderstood.

And it just reminded me that we had same problem in the past, I think we changed UI logic to skip requesting validation if mapping is empty. @pleacu can you help finding the spot?

The other option is to remove this validation entry, then add an WARN level `Audit` before entering here
https://github.com/atlasmap/atlasmap/blob/41386ed36b1d438943d6170721d6185c4a1c9808/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java#L475

Validation is mainly for design time, then Audit is for runtime logging. In this case audit might fit better.

pleacu: validateMappings() is only triggered  if (this.cfg.mappings) in notifyMappingUpdated (mapping-management.service).

rkorytkowski: Thanks! I've fixed the UI logic in validateMappings() to skip validation if there are no complete mappings. The backend stays the same. See https://github.com/atlasmap/atlasmap/pull/2109

PR #2102, opened 2020-05-22
Added more data-testid's to facilitate easier test automation

tplevko:


ui/packages/atlasmap/src/UI/ColumnMapper/Column.tsx
```diff
@@ -13,11 +13,13 @@ const styles = StyleSheet.create({
 });
 
 export interface IColumnProps {
+  id: string;
```
riccardo-forina, line 16: There is no need for this, `Column` already accepts any `data-` attribute you pass.
We just need to set the right `data-testid` where `Column` is used

ui/packages/atlasmap/src/UI/ColumnMapper/ColumnMapper.stories.tsx
```diff
@@ -32,7 +32,7 @@ export const example = () => (
     <ColumnMapper onClick={action("onClick")}>
       <NodeRefProvider>
         <Columns>
-          <Column totalColumns={3}>
+          <Column id={"source"} totalColumns={3}>
```
riccardo-forina, line 35: Just pass `data-testid` instead of `id`.

ui/packages/atlasmap/src/Views/ColumnMapperView/Actions/EnableJavaClassAction.tsx
```diff
@@ -4,10 +4,12 @@ import { AddCircleOIcon } from "@patternfly/react-icons";
 
 export interface IEnableJavaClassActionProps {
   onCustomClassSearch: () => void;
+  id: string;
```
riccardo-forina, line 7: A pattern I found that works is instead of adding `id`s like you can simply pass the `data-testid` to the component and then pass all the "unknown" props to the root element. Example in the following lines

ui/packages/atlasmap/src/Views/ColumnMapperView/Actions/EnableJavaClassAction.tsx
```diff
@@ -4,10 +4,12 @@ import { AddCircleOIcon } from "@patternfly/react-icons";
...
 }
 
 export const EnableJavaClassAction: FunctionComponent<IEnableJavaClassActionProps> = ({
   onCustomClassSearch,
+  id,
```
riccardo-forina, line 12: Example: `...props` instead of `id`

ui/packages/atlasmap/src/Views/ColumnMapperView/Actions/EnableJavaClassAction.tsx
```diff
@@ -28,6 +30,7 @@ export const EnableJavaClassAction: FunctionComponent<IEnableJavaClassActionProp
         variant="plain"
         onClick={onEnableClassSearch}
         aria-label="Enable specific Java classes from your previously imported Java archive."
+        data-testid={`enable-specific-java-classes-${id}-button`}
```
riccardo-forina, line 33: Example: `{...props}` instead of `data-testid={...}`

Then where this component is used you can simply set the `data-testid` property as you need it

tplevko:
Thanks for the comments @riccardo-forina! Updated the old data-testid's nad added also some missing data-testid's for the newly updated dialogs

ui/packages/atlasmap/src/UI/ColumnMapper/Column.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-import { boolean, number } from "@storybook/addon-knobs";
+import { boolean, number, text } from "@storybook/addon-knobs";
```
riccardo-forina, line 3: `text` is unused and makes CI fail

tplevko: Oh, missed that one, I will check the CI results next time. Should be OK now.

PR #2125, opened 2020-05-28
fix(ui): address UXD feedback for toolbars

### LINKED ISSUE #2122: Toolbars background should be white

### LINKED ISSUE #2123: Reduce the number of icons on the main toolbar

### LINKED ISSUE #2124: Align the conditional expression control to PatternFly guidelines
`f(x)` should be a button control, the text area should be disabled when so, and look like a control when enabled.

riccardo-forina:
Toolbars background is now white, fixes #2122
The main actions (import, export, reset)  are now grouped in a menu,
fixes #2123.
The "Import catalog" doesn't allow for jar files, a new "Import Java archive" voice has been added to do that, fixes ##2110.
Align the conditional expression control to PatternFly guidelines, fixes
 #2124.

### DIFF
--- a/ui/packages/atlasmap/src/Atlasmap/toolbarItems.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/toolbarItems.tsx
@@ -1,7 +1,16 @@
 import React, { FunctionComponent, useEffect, useRef } from "react";
 import { useFilePicker } from "react-sage";
 
-import { Button, ToolbarItem, Tooltip } from "@patternfly/react-core";
+import {
+  Button,
+  ToolbarItem,
+  Tooltip,
+  Dropdown,
+  DropdownToggle,
+  DropdownItem,
+  DropdownSeparator,
+  DropdownItemIcon,
+} from "@patternfly/react-core";
 import {
   BezierCurveIcon,
   CodeIcon,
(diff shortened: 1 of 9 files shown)


riccardo-forina:
# White toolbars

image

# Grouped main actions, new action to import Java archives

image

# Conditional expression as a control

## Disabled

image

Nor the `f(x)` not the text area are clickable.

## Enabled and with content

image

Happens when a mapping with a mapping expression is selected.
Both the `f(x)` and the text area are clickable. Clicking on `f(x)` will disable the mapping expression for the selected mapping.

## Enabled but without content

image

Happens when a mapping without a mapping expression is selected.
Only the `f(x)` is clickable. Clicking on `f(x)` will enable the mapping expression for the selected mapping.

pleacu:
If I select one of the import icons in the Atlasmap pull-down menu and then I select 'Cancel' from the import dialog then the pull-down menu remains.  Not sure what the PF rules are with regard to a canceled dialog after a menu selection.

riccardo-forina:
I think it should be ok because it's less punitive to the user if he miss-clicked the item than forcing a reopen of the menu. I don't think I have seen any guideline about this, but I might have missed it.

ui/packages/atlasmap/src/Atlasmap/toolbarItems.tsx
```diff
@@ -15,17 +24,98 @@ import {
...
+    <ToolbarItem>
+      <Dropdown
+        toggle={
+          <DropdownToggle
+            id="atlasmap-toggle"
+            onToggle={onToggle}
+            iconComponent={CaretDownIcon}
+            data-testid="atlasmap-menu-button"
+          >
+            Atlasmap
```
igarashitm, line 105: guessing this is the label for the import/export dropdown, AtlasMap (capital M) might be better for user facing label?

pleacu: would be cool to have just the new AM logo - just a thought...

riccardo-forina: 🤦‍♂️ every-single-time. Thanks! I need to write a plugin for eslint to catch these...

riccardo-forina: > would be cool to have just the new AM logo - just a thought...

Makes sense, we can try that once ready.

PR #2138, opened 2020-05-29
fix(ui): decouple the MappingTableView from the business logic functions

### LINKED ISSUE #2137: Package atlasmap builds with a cyclic dependency warning

riccardo-forina:
Fixes #2137

### DIFF
--- a/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
@@ -282,7 +282,12 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
           />
         );
       case "MappingTable":
-        return <MappingTableView mappings={mappings} />;
+        return (
+          <MappingTableView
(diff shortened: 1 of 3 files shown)


riccardo-forina:
It's all about decoupling, which leads to ease of testing (and writing stories in Storybook).

The fact that the core is a singleton that can be pulled in and be used everywhere shouldn't be a shortcut to avoid wiring the components the expected way, which is by passing properties to the components.

The hierarchy is:

```
Atlasmap
  Views
	Layout
  	UI
```

Each level can depend on any component at the same level or a lower level (eg. Layout can depend on UI components, but can't depend on Views components).

jpav:
Okay, as I mentioned initially, I completely get the desire to have decoupling, but it's the rules behind that decoupling that I feel we should document, basically your last statement about what the expectation is as to which level can depend on what. Maybe it's just me, but I'm not used to seeing that level of decoupling in Java projects. Not saying there shouldn't be, but just that it's not necessarily an obvious expectation.

Issue #2127, opened 2020-05-28
Avoid save mapping request with invalid string "no data" from UI


igarashitm:
xref: https://issues.redhat.com/browse/ENTESB-13919

pleacu:
xref: https://issues.redhat.com/browse/ENTESB-13932

igarashitm:
Ugh... it sounds like an another side effect of #1946

igarashitm:
Changed the title, due to ENTESB-13932 we can't skip save mapping. But we have to avoid sending "no data" as a mapping definition from UI.

PR #2140, opened 2020-05-29
fix: Don't generate an exportMappings if no mappings exist

pleacu:
Fixes: #2127

### DIFF
--- a/ui/packages/atlasmap-core/src/services/document-management.service.ts
+++ b/ui/packages/atlasmap-core/src/services/document-management.service.ts
@@ -77,7 +77,7 @@ export class DocumentManagementService {
    * @param buffer
    */
   static generateExportMappings(buffer: string): string {
-    if (buffer === null || buffer.length === 0) {
+    if (!buffer || buffer.length === 0) {
       return '';
     }
     const metaStr =
--- a/ui/packages/atlasmap-core/src/services/file-management.service.ts
+++ b/ui/packages/atlasmap-core/src/services/file-management.service.ts
@@ -586,7 +586,7 @@ export class FileManagementService {
           if (body) {
             observer.next(body);
           } else {
-            observer.next('no data');
+            observer.next(undefined);
           }
           observer.complete();
         })


igarashitm:
I wonder why `no data` is fed instead of null|undefined|empty string tho... could there be another possibility of invalid data?

pleacu:
> I wonder why `no data` is fed instead of null|undefined|empty string tho... could there be another possibility of invalid data?

I think it's fine - it's a useful way to indicate that there's no data yet.  We did the same in angular.

igarashitm:
Generally speaking this kind of string literal flag tend to be a source of bug like this. I would suggest to change it to undefined or empty string. I did grep in ui source and didn't find anything depend on this "no data" string literal.

igarashitm:
Another option is to define a constant variable in common area, like in config.types.ts, e.g. const NO_DATA_FLAG = 'no data'. But I would defer doing it until other part requires it.

pleacu:
The observer.next() function always returns a string - if you attempt to pass en empty string you'll get '""' (a string of length two containing two double quotes).  If you pass null you'll get "null".  I guess there's no way to pass an empty string.

igarashitm:
undefined just works for me
```diff
diff --git a/ui/packages/atlasmap-core/src/services/file-management.service.ts b/ui/packages/atlasmap-core/src/services/file-management.service.ts
index 620c1a97f5..611e45fd38 100644
--- a/ui/packages/atlasmap-core/src/services/file-management.service.ts
+++ b/ui/packages/atlasmap-core/src/services/file-management.service.ts
@@ -573,7 +573,7 @@ export class FileManagementService {
    * Retrieve the current user AtlasMap data mappings from the server as a JSON object.
    */
   private getCurrentMappingJson(): Observable<any> {
-    return new Observable<any>((observer: any) => {
+    return new Observable<any>((observer) => {
       const baseURL: string =
         this.cfg.initCfg.baseMappingServiceUrl + 'mapping/JSON/';
       this.api
@@ -586,7 +586,7 @@ export class FileManagementService {
           if (body) {
             observer.next(body);
           } else {
-            observer.next('no data');
+            observer.next(undefined);
           }
           observer.complete();
         })
```
The culprit returning null string or quotations is this JSON.stringify()
https://github.com/atlasmap/atlasmap/blob/2510e8627044bdec23fb5411ef983513353e43ce/ui/packages/atlasmap-core/src/services/file-management.service.ts#L541
It requires null|undefined| check as well IMO, but luckily undefined is an exception that stringify() returns undefined
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

igarashitm:
@pleacu LGTM - can you squash the commits?

Issue #2146, opened 2020-06-01
Make field nodes distinguishable from text in the expression box.


pleacu:
It's necessary for users to be able to distinguish their field nodes vs plain text.

Make this:

![sc1](https://user-images.githubusercontent.com/2981421/83449217-39463580-a421-11ea-8382-013b5c12588b.png)

look like this:

![sc2](https://user-images.githubusercontent.com/2981421/83449247-49f6ab80-a421-11ea-951a-73729ba8cc35.png)

PR #2147, opened 2020-06-01
fix: Distinguish field nodes from text in the expression box.

pleacu:
Fixes: #2146


pleacu:
@riccardo-forina - wdyt - the angular UI used a background color to distinguish the field nodes.  I think using bold is on keeping with the new UI

pleacu:
Yes - need one more pass.

Issue #2112, opened 2020-05-26
expression fails to process primitive value


igarashitm:
xref: https://issues.redhat.com/browse/ENTESB-13900
https://issues.redhat.com/browse/ENTESB-13912 
https://issues.redhat.com/browse/ENTESB-13913

New expression possibly returns `Field` object to support handling COMPLEX parent object while it used to be always raw value, which causes `ERROR, message=Expression processing error: io.atlasmap.java.v2.JavaField cannot be cast to java.lang.Comparable` etc
```
java.lang.ClassCastException: io.atlasmap.java.v2.JavaField cannot be cast to java.lang.Comparable
	at io.atlasmap.expression.internal.ComparisonExpression.evaluate(ComparisonExpression.java:329) ~[atlas-expression-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
	at io.atlasmap.expression.internal.ComparisonExpression.matches(ComparisonExpression.java:531) ~[atlas-expression-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
	at io.atlasmap.functions.IF.lambda$create$0(IF.java:36) ~[atlas-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
	at io.atlasmap.core.DefaultAtlasExpressionProcessor.processExpression(DefaultAtlasExpressionProcessor.java:58) [atlas-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
	at io.atlasmap.core.DefaultAtlasContext.process(DefaultAtlasContext.java:491) [atlas-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
```
An another side effect of #2094

PR #2152, opened 2020-06-03
fix: Error when trying to concantenate two strings with expression

rkorytkowski:
Fixes: ENTESB-13912

### DIFF
--- a/lib/expression/pom.xml
+++ b/lib/expression/pom.xml
@@ -38,6 +38,10 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>io.atlasmap</groupId>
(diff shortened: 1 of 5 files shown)


rkorytkowski:
@igarashitm I'm not sure, if we need to apply this fix at the ArithmeticExpression level or some more general fix is needed for all expressions. You may be better informed since it seems to be related to your changes in expressions.

igarashitm:
@rkorytkowski I think this has a same root cause with ENTESB-13900 and ENTESB-13913, I'm working on it

igarashitm:
#2112

rkorytkowski:
Done in https://github.com/atlasmap/atlasmap/pull/2157

Issue #2151, opened 2020-06-03
Edit property name is not working


apupier:
tested with version 2.0.3 in Chrome

- add a property with the + button
- click on "edit property" icon button
- change name
- click "Confirm"
--> the old value is still provided

riccardo-forina:
@pleacu I briefly debugged this and I have a doubt, checking the [`editProperty`](https://github.com/riccardo-forina/atlasmap/blob/93aa3656fc191243fb6709741fd9039c444b7dc8/ui/packages/atlasmap/src/Atlasmap/utils/field.ts#L118-L134) function there is no way to change its name, just the value and type. So I wonder, is changing the name even supported?

riccardo-forina:
In any case, given that `editProperty` signature the first parameter of [this line](https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap/src/Atlasmap/useAtlasmapDialogs.tsx#L81) should be `property.name` and not `name`.

Then if the name can actually be edited, then a new param for it should be added to the `editProperty` signature.

pleacu:
The angular UI could modify the name but it has the business logic alot closer to the UI components.  I actually have a fix that extends the editProperty() api to include the original prop name.  PR coming soon.

PR #2153, opened 2020-06-03
fix: Support editing property names.

pleacu:
Fixes: #2151


ui/packages/atlasmap/src/UI/dialogs/PropertyDialog.tsx
```diff
@@ -27,6 +27,7 @@ export interface IProperty {
   name: string;
   value: string;
   valueType: string;
+  initialName: string;
```
riccardo-forina, line 30: There is no need for this, the initial data is already passed and available, check my comment in the issue

PR #2158, opened 2020-06-04
fix: Add preview capability to mapping table view

### LINKED ISSUE #2121: Add preview capability to mapping table view

jpav:
Fixes #2121

### DIFF
--- a/ui/packages/atlasmap/src/Views/MappingTableView.tsx
+++ b/ui/packages/atlasmap/src/Views/MappingTableView.tsx
@@ -19,6 +19,7 @@ import {
 
 import { MainContent } from "../Layout";
 import { IAtlasmapField, IAtlasmapMapping } from "../Views";
+import { DocumentFieldPreview } from "../UI";
 
 const emptyContent = [
   {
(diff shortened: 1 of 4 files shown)


ui/packages/atlasmap/src/Views/MappingTableView.tsx
```diff
@@ -46,25 +48,98 @@ const styles = StyleSheet.create({
...
+  };
+
   const rows =
     mappings.length === 0
       ? emptyContent
       : mappings.map((mapping) => {
-          const sources = mapping.sourceFields.map((source) => {
-            const { name } = source as IAtlasmapField;
-            return <div key={name}>{name}</div>;
+          const sources = mapping.sourceFields.map((source, index) => {
```
riccardo-forina, line 124: I'm not sure I'm following, why are you searching for the field again when the field is already available?

jpav: I assume you mean the search inside renderPreview that starts with the field from the mapping. The reason is the mapping fields are missing critical information, such as the preview value and field.mappings is always empty, which causes the shouldShowMappingPreviewForField function in Atlasmap.tsx to always return false.

riccardo-forina: Well, we don't really need `shouldShowMappingPreviewForField` in this case, a `shouldShowMappingPreview: (mapping: IAtlasmapMapping) => boolean` should be enough since then all the fields of a selected mapping that should show a preview should show a preview. 

And for the missing data I'd work on [`fromMappedFieldToIMappingField`](https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap/src/Atlasmap/utils/ui.ts#L133-L178) to sync how a field is generated, by returning an enhanced version of [`fromFieldToIFieldsNode`](https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap/src/Atlasmap/utils/ui.ts#L69-L96).

jpav: I did it this way to reuse existing code, but I can look at a simpler version that just checks whether the mapping is selected.

ui/packages/atlasmap/src/Views/MappingTableView.tsx
```diff
@@ -46,25 +47,54 @@ const styles = StyleSheet.create({
...
   mappings,
   onSelectMapping,
+  shouldShowMappingPreview,
+  onFieldPreviewChange,
 }) => {
+  const renderPreview = (
+    mapping: IAtlasmapMapping,
+    mappedField: IAtlasmapField,
+  ) =>
+    mappedField &&
```
riccardo-forina, line 66: This check isn’t really needed since the type doesn’t allow for null

jpav: Ah, that was a holdover from the previous code. One minute...

PR #2163, opened 2020-06-04
chore: Eliminate wrong upcasting to any

### LINKED ISSUE #2141: Eliminate wrong upcasting to any
This one upcasts rxjs `Subscriber` to `any` without reason
https://github.com/atlasmap/atlasmap/blob/6a314d2febd1c2e00ba454fa5b1c2793ae9c2be5/ui/packages/atlasmap-core/src/services/file-management.service.ts#L576

which prevents code assist in IDE as well as being bug prone. Same thing spreads everywhere.

igarashitm:
Partially fixes #2141, only for rxjs subscriber which was obvious.

### DIFF
--- a/ui/packages/atlasmap-core/src/services/document-management.service.ts
+++ b/ui/packages/atlasmap-core/src/services/document-management.service.ts
@@ -144,7 +144,7 @@ export class DocumentManagementService {
   }
 
   fetchClassPath(): Observable<string> {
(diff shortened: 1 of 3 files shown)


igarashitm:
This revealed implicit cast from `ArrayBuffer` to `Uint8Array` causes syntax error... better to do it after branch off

Issue #2161, opened 2020-06-04
Chore: Update custom field action maven test sources


pleacu:
existing sources are out of date

igarashitm:
@pleacu is there still something to do? seems it's already using `@AtlasActionProcessor`

pleacu:
Maybe - custom field actions are still broken - was going to wait for that to be fixed before closing this.

PR #2164, opened 2020-06-04
Correct custom field action maven tests

pleacu:
Fixes: #2161

was on the wrong branch in last PR


igarashitm:
can you avoid using wildcard import? I would have to prohibit in checkstyle...

pleacu:
actually I'll pull this for now - It can be part of my bug fix for custom field actions

Issue #2007, opened 2020-05-05
Map source elements to a specific instance in a target array


vratojr:
Hello, 
I'd like to develop the following feature.

A screenshot is worth 1000 words so here it is :) :
![image](https://user-images.githubusercontent.com/1368836/81070982-56501d00-8ee4-11ea-889f-95d608369615.png)
The idea is that of being able to map to specific instances in an array. So the idea is that of adding a + sign on each array to add a specific instance over which the mapping can be defined. And a minus to remove it.

Would you be interested in integrating this new feature? 
If yes could you provide some guidance concerning how you would implement it? And if you are ok concerning the ui (well, we'd make it a bit better of course :) ).

PS: I know the title is almost the same to issue #1418 ... Indeed it's a kind of superset of that issue (maybe you could rename 1418 to "Map multiple source elements to their K/V instances in a target array"?)

zamrokk:
It is similar to what is doing IBM WTX : https://www.youtube.com/watch?v=uVxYPH2gwM4
I vote for it !

igarashitm:
+1 This is a nice to have.
I would enhance `Mapping Details` pane to handle this kind of usecase, instead of showing it in document itself. Internally we can add a transformation to put a value into specific indexed collection item, like `putAt()`.

igarashitm:
By the way, please note that we're migrating to React version of UI and Angular version will be removed in 2.0.

zamrokk:
If you keep the mapping of items on the panel then it is not very nice to do drag n drop to the item directly on the tree. I would prefer like WTX does

zamrokk:
Btw, why react and no more angular?

zamrokk:
Also you do have usage of right click to do actions on an item of the tree. Is it a choice to? Do you prefer contextual buttons on a tree?

riccardo-forina:
We don't have any contextual menu or action in order to support assistive technologies and touch devices. In the React version we have a kind of toolbar that is shown on top of the field for mouse users with all the actions that can be performed on the field, which is also available to keyboard users moving through the app using tabs and arrows, and the enter key to focus a field.

![2020-05-06 at 16 24 53 - Emerald Mammal](https://user-images.githubusercontent.com/966316/81188954-486acc80-8fb6-11ea-94cc-7e2aa5487d6a.gif)

Atlasmap moved to React in order to use [PatternFly](https://www.patternfly.org/v4/) as the UI library.

igarashitm:
[...]

Issue #1253, opened 2019-10-01
Add copyTo one-to-many transformation


igarashitm:
Currently we only have `Split` one-to-many transformation, but sometimes we only copy the single string into the first element of target collection without splitting, say `/company/name` -> `/[0]/company`.  Having `0` as a default parameter for the index, copyTo transformation should be able to copy it into specified index in the collection.

A workaround for this is to specify a delimiter which never appears in a source String, so that the `Split` won't split anything.

zamrokk:
Please implement this feature "CopyTo" function should be added along with the actual "Split" for One-to-many cases

igarashitm:
Addressed by #2171

PR #2171, opened 2020-06-09
Map source elements to a specific instance in a target array

vratojr:
See #2007 
POC of the json implementation (backend only atm)

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/actions/CollectionActions.java
+++ b/lib/core/src/main/java/io/atlasmap/actions/CollectionActions.java
@@ -1,0 +1,17 @@
+package io.atlasmap.actions;
+
+import io.atlasmap.spi.AtlasActionProcessor;
+import io.atlasmap.spi.AtlasFieldAction;
+import io.atlasmap.v2.CopyTo;
+import io.atlasmap.v2.FieldType;
+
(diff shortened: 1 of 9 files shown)


vratojr:
Targets #1253

lib/core/src/main/java/io/atlasmap/actions/CollectionActions.java
```diff
@@ -0,0 +1,17 @@
+package io.atlasmap.actions;
```
igarashitm, line 1: Now I wonder if we need this class - since actual transformation entry in the mapping definition is removed in DefaultAtlasContext. Maybe just the `CopyTo` model class is enough?

vratojr:
Doh, I solved your comment by error... Anyway. As far as I can see, CollectionActions is necessary because otherwise the action doesn't show in the UI. More than this, it's the Processor behavior that defines the multiplicity of the action (see comment in CollectionActions) and so the fact that it's correctly displayed for one-to-many mappings.

igarashitm:
Ah that's right, it seems field action service requires logic classes to be resolved as well to build a list of action metadata used for UI. I think this is ready to merge then.

igarashitm:
@vratojr Oops, I forgot to ask adding `feat:` prefix for a commit message, here is a reference for next time

Issue #2043, opened 2020-05-11
UI_REACT: Cannot build on windows host due to prettier EOL error


vratojr:
Hello, unfortunately I'm working in a windows environment. This means that, by default, git checks out all files with CRLF endings and commit with simple LF endings.
Apparently, prettier by [default ](https://prettier.io/docs/en/options.html) expect the default line ending to be LF.
But to me what they propose seems too much since, as I said, the default git config for windows transforms CRLF to LF when committing.
If you agree, this can be easily fixed by adding:
```
"endOfLine":"auto"
```
In .prettierc

PS: I can open a PR if you agree on the fix.

riccardo-forina:
Thanks for reporting this. I'd rather fix it as suggested by them, by hinting git to use LF by default:

```
Add * text=auto eol=lf to the repo's .gitattributes file. You may need to ask Windows users to re-clone your repo after this change to ensure git has not converted LF to CRLF on checkout.
```

You'll have to reclone the repo then. Does that work for you?

vratojr:
For me it's ok. But if, on windows, git's standard it to do checkout with CRLF and commit with LF I'd leave it as it is.
I personally won't act on the git configuration just to make prettier happy.

riccardo-forina:
That setting would be specific for this project to and would ensure consistency in how the new line is handled across environments. It will not change how git behaves son other repos.

vratojr:
The consistency is guaranteed by git itself that performs the conversion at checkout/commit time.
What I meant is that, even if it's just for this repo, to me it's "strage" to alter git behavior to match what is expected by prettier rather than changing prettier. 
Because the guys from prettier DECIDED that the code should be formatted by using LF. Instead git is leaving the freedom but also ensured consistency.
Even if I normally am a linux user I don't see why we should impose a specific line ending if there is no true need to do it. I mean, it's years that linux and windows users collaborate with git. I've been always using the defautl config and it always worked flawlessy.

riccardo-forina:
AFAIK the consistency is per file, not per project. This setting would make it per project. I’m not sure why they made it default to LR instead of auto how it was, but I guess there is some edge case where auto is a pitfall and this saves from that. It’s not hard to believe either, especially if we start considering files different than sources.

Anyhow, either way works for me. We have been on auto so far and it’s been ok.

vratojr:
As you prefer :)

riccardo-forina:
Go with the `.prettierrc` :)

[...]

PR #2172, opened 2020-06-09
Fix end-of-line errors on windows: 

vratojr:
See #2043


vratojr:
@riccardo-forina ,please, why nobody is answering me on the #2007 ?

PR #2177, opened 2020-06-12
fix: SourceTargetLinks causing warning about missing keys in list

### LINKED ISSUE #2176: SourceTargetLinks causing warning about missing keys in list
Appears in browser console:
```
react.development.js:315 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `SourceTargetLinks`. See https://fb.me/react-warning-keys for more information.
    in MappingLines (created by SourceTargetLinks)
    in SourceTargetLinks (created by SourceTargetView)
```

jpav:
Fixes #2176

### DIFF
--- a/ui/packages/atlasmap/src/Views/ColumnMapperView/Links/SourceTargetLinks.tsx
+++ b/ui/packages/atlasmap/src/Views/ColumnMapperView/Links/SourceTargetLinks.tsx
@@ -1,12 +1,10 @@
+import { css, StyleSheet } from "@patternfly/react-styles";
 import React, {
   FunctionComponent,
   MouseEvent,
   ReactElement,
   useMemo,
 } from "react";
-
-import { css, StyleSheet } from "@patternfly/react-styles";
-
 import { NodesArc, useToggle } from "../../../UI";
 import { IAtlasmapMapping } from "../../../Views";
 import { SOURCES_FIELD_ID_PREFIX, TARGETS_FIELD_ID_PREFIX } from "../Columns";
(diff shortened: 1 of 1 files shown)


ui/packages/atlasmap/src/Views/ColumnMapperView/Links/SourceTargetLinks.tsx
```diff
@@ -35,11 +35,12 @@ export const SourceTargetLinks: FunctionComponent<ISourceTargetLinksProps> = ({
 
   return (
     <>
-      {sortedMappings.map((m) => (
+      {sortedMappings.map((m, index) => (
         <MappingLines
           mapping={m}
           onClick={() => onSelectMapping(m)}
           isSelected={m.id === selectedMappingId}
+          key={index}
```
riccardo-forina, line 43: It's always better to give a meaningful key if possible to help React with the reconciliation. In this case, `m.id` could be a better option.

jpav: My understanding was just that they need to be unique. I'll go ahead and make the change, but just for my own understanding, can you expound on how and why the id's would be better?

riccardo-forina: In short, a key should be stable and not change between renders. With an index in this case if a mapping gets added/remove keys will be different for some elements making React's job harder. For more details check the doc https://reactjs.org/docs/lists-and-keys.html#keys

jpav: k, that makes sense. I've updated the code with this in mind.

PR #2183, opened 2020-06-14
Transformation approach

jpav:
The storyboard demonstrates the approach of providing mapping capability to the user entirely through the use of transformations, ones that we support under the hood via the existing function expressions library. The idea is to eliminate the restraints we have with the current source-/target-based approach (documented elsewhere), the confusion added by the current expression capability's interaction with field actions (i.e., transformations), and better match the user's mental model of how they'd expect mappings to be realized.


riccardo-forina:
I don’t understand, why did you add all these stories that look like copies of the relative component?

jpav:
They all have modifications that support the new demo I'm trying to provide for a transformation approach to mapping. Have I done this the wrong way? I wasn't really sure how to deal with naming of the files, so I just added a '.story' to any file I needed to change for the main story I added in Atlasmap.stories.tsx. If this is all wrong, please let me know how I should go about this.

riccardo-forina:
I'm picking a random example:
https://github.com/jpav/atlasmap/blob/xform/ui/packages/atlasmap/src/Atlasmap/dialogs/useDeleteConstantDialog.tsx
https://github.com/jpav/atlasmap/blob/xform/ui/packages/atlasmap/src/Atlasmap/dialogs/useDeleteConstantDialog.story.tsx

These two files are identical. A story file is a Storybook file in a [component story format](https://storybook.js.org/docs/formats/component-story-format/).

I really don't understand what was the goal here.

jpav:
Well, I' probably doing this all wrong, but I thought when we last talked, you said that if I need to change the code of an existing file for the purposes of the story, I should just copy it. I'm hoping its just my naming convention that is the problem. What I've done is added a single additional story in Atlasmap.stories.tsx. This is a 'demo' level story, that interacts with all of the other code, and is thus not a 'unit test' in any sense. For what I want to demonstrate as changes using this transformation approach, I will need to modify the contents of lots of existing code. I didn't know an appropriate name for each of them since they were all very close to what now exists, and I didn't want to add a bunch of stuff that would confuse the other developers as to whether its part of our regular code base, so I decided to add '.story.' to the names to make it clear these are not production files. The example you showed above, useDeleteDialog.tsx, is actually different in the imports because it uses changes that I made to AtlasmapProvider.tsx. Hopefully that at least explains my intentions. If I've done this all wrong, please let me know how to approach this.

gashcrumb:
I would suggest to make this a draft PR so you can change what's needed on your branch and keeps it clear to other folks that it's not ready to be merged.

jpav:
@gashcrumb I could do that, but my concern is that I'll be making tons of changes to this code, and I'd really like to get feedback from @riccardo-forina and others about my contributions, as they've been very helpful thus far. I'm concerned that if it just stays as a draft that I update from my branch, no one will look at it until it's thousands of lines of code and impossible to practically review.

gashcrumb:
> until it's thousands of lines of code and impossible to practically review.

I'm confused though how making alternate copies of existing files and diverging them solves this.  Wouldn't that still mean there'd still be considerable effort to consolidate all that work if folks buy into the changes the PR introduces?

Also a bit of context in the description might help as to what this solves/introduces/changes etc.

jpav:
I updated the description to explain the intent better. The beauty of doing this via a storyboard full-app demo is that I can fully integrate this approach with the existing code base without affecting the existing behavior. It's currently implemented as a 'test' so-to-speak since it's all via storyboard, but even there I've been careful to write it so that the classes at the top layers of integration, namely AtlasmapProvider and Atlasmap, integrate the existing code right along side this new approach. Thus, if this later gets buy-in from customers, product managers, etc., it can pretty easily be integrated either as a parallel approach, probably after choosing a different naming convention than what I've done for the stories, OR as a replacement approach just by removing the 'older' classes. That's all probably still a long ways off of course, but I think the migration approach will hold up. My main concern now is making sure I'm doing the storyboard approach the 'expected' way, so I'm hoping to get more feedback from Riccardo in this regard.

riccardo-forina:
> Well, I' probably doing this all wrong, but I thought when we last talked, you said that if I need to change the code of an existing file for the purposes of the story, I should just copy it. 

I remember specifying that copying components was ok "for demo purposes". Surely it's not ok if the goal is to merge on master. So my question is, what's the goal for this PR? 

If it's to demo your idea, there is no need to review anything in the code since it will not be merged ever, not with all this code duplication.

If you plan for this to be merged and part of the product then it there should be no code duplication, to allow for a review. In the state it is, this isn't reviewable since I have no idea about what you changed. 

> I'm hoping its just my naming convention that is the problem. What I've done is added a single additional story in Atlasmap.stories.tsx. 

The change in that file is ok.

> This is a 'demo' level story, that interacts with all of the other code, and is thus not a 'unit test' in any sense. For what I want to demonstrate as changes using this transformation approach, I will need to modify the contents of lots of existing code. 

I take this as a personal failure, I spent a lot of time trying to make everything as composable and decoupled as possible. What did you have to change? I can't tell without going through all 4k lines (which I'd rather not have to do).

> I didn't know an appropriate name for each of them since they were all very close to what now exists, and I didn't want to add a bunch of stuff that would confuse the other developers as to whether its part of our regular code base, so I decided to add '.story.' to the names to make it clear these are not production files. The example you showed above, useDeleteDialog.tsx, is actually different in the imports because it uses changes that I made to AtlasmapProvider.tsx. Hopefully that at least explains my intentions. If I've done this all wrong, please let me know how to approach this.

A file with the `.story.tsx` (or `.stories.tsx`) suffix is a Storybook story file, as [documented here](https://storybook.js.org/docs/formats/component-story-format/).
The files you added aren't story files, are just copies of the existing components. This is _very_ confusing for anyone that follows Storybook's conventions.

About the changes in the `AtlasmapProvider` component, I can't tell what you changed without manual diffing the files. Anyhow, if we lack some API there, it's ok to add them even if only your demo uses them if it belongs there.

--- 

The thing is, you are basically asking your reviewers to manually diff the files hunting for the changes. This is a big ask which requires a very solid reason behind. The reason you are providing is

> but my concern is that I'll be making tons of changes to this code (...). I'm concerned that if it just stays as a draft that I update from my branch, no one will look at it until it's thousands of lines of code and impossible to practically review.

This is already impossible to review. I honestly don't think what you have in mind will require thousands of lines of code to achieve, but assuming that is the case then:
- either make smaller incremental PRs, easier to review and merge
- or else, work on a long running branch with your idea kept in sync with upstream until the feature is ready to be merged (this is what I did for the last rewrite).

jpav:
@riccardo-forina There's certainly no reason to take my misunderstanding of how I might go about developing a new approach as a 'personal failure'. The failure is on me. After a conversation with Stan, I'm now going to change how I do this.

1. I have retracted all of the .story files, and the addition to AtlasMap.stories.tsx.
1. I have removed all of the changes except an initial set of modifications that mark the beginning work towards this transformation approach. The number of changes should now be small enough to allow for a practical review.
1. I have marked this issue as a draft, and moving forward, after each requested review, will continue to update it with manageable amounts of code changes, and for which I will manually ping you and the UI team for review.
1. Once everyone is comfortable with the progress of this feature, I will mark it as Ready for review, and after a final review, merge it with master and continue thenceforth with regular PRs. I'm hoping I can get past the draft stage very quickly, but that's up to everyone else to decide.

Hopefully that all sounds reasonable.

jpav:
BTW, I know there are several changes to order of imports that were a result of an editor setting I had turned on to organize imports on save. I have turned that off to prevent this in the future.

ui/packages/atlasmap/src/Atlasmap/utils/toolbar.ts
```diff
@@ -218,3 +217,7 @@ export function toggleShowUnmappedFields(enabled: boolean) {
 
   return enabled;
 }
+
+export function toggleShowAllLinks(enabled: boolean) {
```
riccardo-forina, line 221: Is this a placeholder?

ui/packages/atlasmap/src/Atlasmap/AtlasmapProvider.tsx
```diff
@@ -278,15 +277,15 @@ export const AtlasmapProvider: FunctionComponent<IAtlasmapProviderProps> = ({
...
-      console.log(
-        "onUpdates",
-        caller,
-        "initialized",
-        initializationService.cfg.initCfg.initialized,
-        "errors",
-        initializationService.cfg.initCfg.initializationErrorOccurred,
-      );
+    function onSubUpdateCb(_caller: string) {
+      // console.log(
```
riccardo-forina, line 281: I'd rather keep this log, it's helpful to grasp what's going on with the backend sometimes.

ui/packages/atlasmap/src/Atlasmap/AtlasmapProvider.tsx
```diff
@@ -58,32 +71,18 @@ import {
...
-import {
-  INotificationsState,
-  IDataState,
-  initDataState,
-  initNotificationsState,
-  dataReducer,
-  notificationsReducer,
-  DataActionPayload,
-} from "./reducers";
+import { toggleShowAllLinks } from "./utils/toolbar";
```
riccardo-forina, line 85: I think this can be added to the import above.

ui/packages/atlasmap/src/Views/ColumnMapperView/Links/SourceMappingTargetLinks.tsx
```diff
@@ -79,7 +81,15 @@ export const SourceMappingTargetLinks: FunctionComponent<
         />
       )),
     ];
-  });
+  };
+
+  const selectedMapping = mappings.find((m) => m.id === selectedMappingId);
+  const links = showAllLinks
+    ? mappings.map((m) => createLink(m))
```
riccardo-forina, line 88: You could simplify this to `mappings.map(createLink)`

ui/packages/atlasmap/src/Atlasmap/AtlasmapProvider.tsx
jpav: Yes, I'll uncomment that. I just did that because it's annoying when I'm debugging other stuff using the console.

jpav:
I believe I addressed all of your comments, plus added icon to each mapping to remove that mapping, with the idea that the mapping details sidebar will not show in this view.

jpav:
@riccardo-forina Added ability to edit name of mapping. Not sure how much you're looking for this being flushed out before reviewing view in general, so I'm trying to add things slowly in small pieces that you can comment on.

riccardo-forina:
The Netlify build is failing.

ui/packages/atlasmap/src/Atlasmap/useContextToolbar.tsx
```diff
@@ -58,13 +61,15 @@ export function useContextToolbar({
   showColumnMapperViewToolbarItem = true,
   showMappingTableViewToolbarItem = true,
   showNamespaceTableViewToolbarItem = true,
+  showTransformationApproachViewToolbarItem = true,
```
riccardo-forina, line 64: I think this should be `false` for the moment since the view is still in progress, else it would be displayed to users.

jpav: I can do that, but my thought here was to keep it true so that the team and whatever other developers can see and try out what I'm working on as I go, with the idea that I'd change it to false if things aren't far enough along by the next release.

gashcrumb: The netlify link on this PR is probably the easiest way for folks to see how this is progressing.  I'd be wary of letting partially completed features into master.  I had a peek earlier today and while I don't know what the gaps are in terms of what's completed, it also wasn't clear to me yet where the benefit is.  I see that it restricts when the mapping lines are drawn, however what's the plan on how the user edits a given mapping, or did I miss that?

jpav: k, no problem, I'll set the flag to false. WRT to benefit of what I'm doing, I emailed a mockup to you and Riccardo a couple of weeks back that should explain the goal with this. You won't see much yet just because I've been waiting for this part to be approved and merged so I can start adding manageable reviewable commits afterwards. Let me know if you didn't receive it, and I'll resend it.

gashcrumb: Yep, I did get it, thanks!  I just don't understand why it's necessary to get bits into master before proceeding?

jpav: Working on this large of a feature isn't tenable in my own private repo. I need to get feedback from the team about the additions I make along the way, so it's important that they can see those changes and try them out as I proceed. With the flag off, the code won't affect any users in any way, but will be an easy toggle for developers to turn on the feature to try it out. I, of course, also want my contributions to be evident in this repo.

gashcrumb: Well, the other option for a long-running thing would be a separate branch in this repo alongside master that would allow you to open PRs against.

jpav: True, but again, no one will look at it that way. I guess I don't see the harm in putting this in master, but I'll do whatever you need me to do.

PR #2203, opened 2020-06-25
fix: Distinguish collection of primitive and collection of complex with icon

### LINKED ISSUE #2173: Distinguish collection of primitive and collection of complex with icon
Currently we're using same icon for "a collection of primitive" and "a collection of complex", but better distinguish them. We might be able to show both complex icon and collection icon for "a collection of complex", does it eat too much space?

jpav:
Fixes #2173 

The 'collection' portion of the icon ends up being really small, so not sure if this will be acceptable, but I can't think of a better way to do this without increasing the size of the icon components overall, which would make the entire app require more real estate.


riccardo-forina:
I'm not a fan to be honest, the icon is just too small. What if we simply add the collection icon to the right of the name as we do for the collection fields? Like this:

image

jpav:
I'm wasn't a fan either. I didn't suspect this approach would fly, but wanted to get some feedback. I was actually trying to go the opposite way of adding an additional icon, as I feel we are very icon-heavy already, but I think I agree that's probably the best option. Closing this PR and will resubmit a different fix.

Issue #2211, opened 2020-06-30
Add UI for providing CSV parameters


rkorytkowski:
Follow up to #2018

rkorytkowski:
![image](https://user-images.githubusercontent.com/1163157/86131667-399c2400-bae6-11ea-87c6-1dd527dca1fb.png)
![image](https://user-images.githubusercontent.com/1163157/86131856-7536ee00-bae6-11ea-9742-e066b367b412.png)

PR #2212, opened 2020-06-30
fix: Add UI for providing CSV parameters

rkorytkowski:
Fixes: #2211

### DIFF
--- a/ui/packages/atlasmap/src/Atlasmap/dialogs/useImportDocumentDialog.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/dialogs/useImportDocumentDialog.tsx
@@ -1,7 +1,9 @@
+import React from "react";
 import { useCallback, ReactElement } from "react";
 
 import { useAtlasmap } from "../AtlasmapProvider";
 import { useConfirmationDialog } from "./useConfirmationDialog";
+import { useParametersDialog } from "./useParametersDialog";
 
 export function useImportDocumentDialog(): [
   ReactElement,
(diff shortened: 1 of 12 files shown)


ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const handleOnConfirm = useCallback(() => {
+    onConfirm(parameters);
+  }, [parameters, onConfirm]);
+
+  const handleOnCancel = useCallback(() => {
+    onCancel();
+    reset();
+  }, [onCancel, reset]);
+
+  const handleAddParameter = useCallback(() => {
```
riccardo-forina, line 66: Following the previous comment, this would be cleared if the selected parameter is an argument of the function, since all you'd have to do is to set the new array in the state.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+      if (newParameter.name === addParameter) {
+        newParameter.hidden = false;
+        break;
+      }
+    }
+    setAddParameter(initialAddParameter);
+    setParameters(newParameters);
+  }, [parameters, addParameter]);
+
+  const handleRemoveParameter = useCallback((name) => {
```
riccardo-forina, line 78: Same as above. Unless I'm missing something, that `hidden` property with the proposed changes is not needed anymore.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+    for (const newParameter of newParameters) {
+      if (newParameter.name === name) {
+        newParameter.hidden = true;
+        break;
+      }
+    }
+    setParameters(newParameters);
+  }, [parameters]);
+
+  const handleOnChangeParameter = useCallback((value, event) => {
```
riccardo-forina, line 89: This could work like the other handlers

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+    setParameters(newParameters);
+  }, [parameters]);
+
+  const hiddenParameters = (): IParameter[] => {
+    return parameters.filter(parameter => parameter.hidden);
+  };
+
+  const visibleParameters = (): IParameter[] => {
+    return parameters.filter(parameter => !parameter.hidden);
+  };
```
riccardo-forina, line 106: This block shouldn't be required anymore with the proposed change. As a general rule, you don't need to make these function, just compute the value directly. Eg:

```
const hiddenParameters = parameters.filter(parameter => parameter.hidden)
```

If it's an expensive operation you can wrap it with `useMemo` to compute it only when the dependencies change. Eg:

```
const hiddenParameters = useMemo(() => parameters.filter(parameter => parameter.hidden, [parameters]); // runs only when parameters changes
```

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+              </SplitItem>
+              {parameter.hideable &&
+                <SplitItem>
+                  <Button variant="plain" aria-label="Action" onClick={(_event) => handleRemoveParameter(parameter.name)}>
+                    <TimesIcon />
+                  </Button>
+                </SplitItem>
+              }
+            </Split>
+          </FormGroup>
```
riccardo-forina, line 153: This block is a great candidate to become a component on its own. It doesn't have to be in a new file since it's used only here, but will make reading the code easier.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+                ))}
+              </FormSelect>
+            </SplitItem>
+            <SplitItem>
+              <Button isDisabled={addParameter === initialAddParameter} onClick={handleAddParameter} variant="link" icon={<PlusIcon />}>
+                Add parameter
+              </Button>
+            </SplitItem>
+          </Split>
+        </FormGroup>
```
riccardo-forina, line 173: Same as above.

ui/packages/atlasmap/src/Atlasmap/dialogs/useImportDocumentDialog.tsx
```diff
@@ -2,25 +2,76 @@ import { useCallback, ReactElement } from "react";
...
       } else {
-        importAtlasFile(selectedFile, isSource);
+        importFile(selectedFile, isSource);
       }
     },
-    [documentExists, importAtlasFile, openImportDialog],
+    [documentExists, importAtlasFile, openImportDialog, openParametersDialog],
   );
-  return [importDialog, onImportDocument];
+  return [importDialog, parametersDialog, onImportDocument];
```
riccardo-forina, line 76: I think we can avoid changing the return type by wrapping the two dialogs together in a fragment:

```tsx
return [<>{importDialog}{parametersDialog}</>, onImportDocument];
```

Having two dialogs is an implementation detail and it would be nice not to leak it outside this function.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+  IConfirmationDialogProps,
+} from "./ConfirmationDialog";
+import {PlusIcon, TimesIcon} from "@patternfly/react-icons";
+
+export interface IParameter {
+  name: string;
+  value: string;
+  options?: IParameterOption[];
+  hidden?: boolean;
+  hideable?: boolean;
```
riccardo-forina, line 25: I'm a bit confused by these two parameters. What do you think about renaming `hidden`  to `required` flipping the logic in the code, and dropping `hideable` completely by making all not required options to be removable (since this is the way it's set up anyway?)

rkorytkowski: From the perspective of this specific use case it's enough to have one `required` parameter. It's a bit more flexibility to be also able to specify whether to hide a parameter or not when first showing the form without making it required.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
rkorytkowski: I'm sorry for my react ignorance. I thought I needed that to access value from the select field.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
rkorytkowski: Done!

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
rkorytkowski: Changed to use only `required` parameter.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
rkorytkowski: Automated refactoring doesn't do the trick so I left it as is.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
```diff
@@ -0,0 +1,178 @@
...
+  parameters: initialParameters = [],
+  isOpen,
+  onCancel,
+  onConfirm,
+}) => {
+  const initialAddParameter = 'Select...';
+
+  const [parameters, setParameters] = useState(initialParameters);
+
+  const [addParameter, setAddParameter] = useState(initialAddParameter);
```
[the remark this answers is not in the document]
rkorytkowski: Done!

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx
rkorytkowski: I couldn't figure out how to pass it as an argument or I'm not following what you mean...

riccardo-forina: Maybe I wasn't clear, but you did what I meant :D https://github.com/atlasmap/atlasmap/pull/2212/files#diff-a22ea50982c257cee5dbd972c2a46731R134-R139

rkorytkowski: I thought you were asking to pass the value as argument to https://github.com/atlasmap/atlasmap/pull/2212/files#diff-a22ea50982c257cee5dbd972c2a46731R49 Anyway, glad to hear it's ok now.

Issue #1409, opened 2019-10-23
Support class name completion


igarashitm:
When user import Java document, ATM they have to write down FQCN which is inconvenient. We have to provide typeahead search by scanning user uploaded jars.

paoloantinori:
we did something similar in Groovy LSP, in case you want to share ideas.
To scan a classpath we have used this lib:

https://www.google.com/search?client=firefox-b-d&q=org+reflections

https://github.com/prominic/groovy-language-server/pull/14/files#diff-b1f0a074d4b3cea4a94efc94bdd1e1deR132-R136

pleacu:
Need to add tests - check PaulsFieldAction (may not be related to this push).

![sc1](https://user-images.githubusercontent.com/2981421/68683760-5672ef80-0535-11ea-878f-1ea83deb8f00.png)

pleacu:
@igarashitm - Hey Tomo - In the DM if I import a custom field action jar, the custom action defined by the custom class (`io.atlasmap.service.my.MyFieldActionsModel`) is automatically loaded into the field actions without having to 'enable' the class.
Is this an error or is the user supposed to enable `MyFieldActions`?

ref:

![sc1](https://user-images.githubusercontent.com/2981421/84810621-99200d00-afd9-11ea-9784-c2f8e389764a.png)

![sc2](https://user-images.githubusercontent.com/2981421/84810877-f7e58680-afd9-11ea-9029-12dac4c4204e.png)

igarashitm:
Custom transformations doesn't have any enable/disable option, it appears where it's eligible once it's loaded. User just decides where to apply.

pleacu:
Then do we need the Enable/Disable circleplus icon?

igarashitm:
I don't personally think so since it's redundant. ImportJar=>applyCustomTransformation and ImportJar=>addJavaDocument looks symmetric and intuitive.

pleacu:
Ok - so I'll bail on adding support for returning a list of loadable FQCNs and turn this into 'Remove Enable/Disable Class' icon from the Source/Target panels.

pleacu:
Importing a jar adds it into the classpath.  The enable class icon is needed to use the class as a document (the jar could contain any number of classes, the user must choose one from them as a document).  The enable class has nothing to do with custom field actions.

![sc3](https://user-images.githubusercontent.com/2981421/84815977-d38da800-afe1-11ea-8064-010dfc0f53ee.png)

[...]

PR #2213, opened 2020-07-02
fix: Support pull-down select menu for class name identification for …

pleacu:
…Java documents defined by user JARs

Fixes: #1409

### DIFF
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasLibraryLoader.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasLibraryLoader.java
@@ -93,7 +93,7 @@ public class AtlasLibraryLoader extends ClassLoader {
         reload();
     }
 
-    public ArrayList<String> getClassCandidates() throws AtlasException {
+    public ArrayList<String> getLibraryClassNames() throws AtlasException {
         final String classSuffix = ".class";
         ArrayList<String> classNames = new ArrayList<String>();
 
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasLibraryLoader.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasLibraryLoader.java
@@ -16,11 +16,13 @@
 package io.atlasmap.service;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashSet;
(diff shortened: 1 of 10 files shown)


pleacu:
@igarashitm @riccardo-forina - just a friendly ping.  If you guys get a chance pls review.  thkx!

lib/service/src/main/java/io/atlasmap/service/AtlasLibraryLoader.java
```diff
@@ -89,6 +93,35 @@ public void clearLibaries() {
         reload();
     }
 
+    public ArrayList<String> getClassCandidates() throws AtlasException {
```
igarashitm, line 96: I would use `getLibraryClassNames()` to align wording with UI - and avoid `candidate` unless there's specific implication, like filtering candidates for some specific purpose.

pleacu: done

pleacu:
Hey @jpav  - could I get you to take a look at the UI elements in this PR?

ui/packages/atlasmap/src/UI/dialogs/CustomClassDialog.tsx
```diff
@@ -71,24 +77,28 @@ export const CustomClassDialog: FunctionComponent<ICustomClassDialogProps> = ({
...
-            placeholder={"com.package.class"}
-            data-testid={"class-package-name-text-input"}
-          />
+            onChange={setCustomClassName}
+            data-testid={"custom-class-name-form-select"}
+          >
+            {isOpen &&
+              customClassNames &&
+              customClassNames.map((value, idx) => (
+                <FormSelectOption key={idx} value={value} label={value} />
```
jpav, line 99: I think it would be better to use value as a key here.

ui/packages/atlasmap/src/UI/dialogs/CustomClassDialog.tsx
```diff
@@ -39,25 +39,31 @@ export interface ICustomClassDialogProps {
...
+  customClassNames,
   collectionType: initialCollectionType = "",
   collectionTypeOptions,
   isOpen,
   onCancel,
   onConfirm,
 }) => {
-  const [value, setValue] = useState(initialValue);
+  const [customClassName, setCustomClassName] = useState(
+    initialCustomClassName,
```
jpav, line 51: Maybe remove comma since you're not providing a callback

ui/packages/atlasmap/src/UI/dialogs/CustomClassDialog.tsx
pleacu: either way - I used the index as the key.

ui/packages/atlasmap/src/UI/dialogs/CustomClassDialog.tsx
pleacu: I'd agree but I think prettier objects... :)

PR #2223, opened 2020-07-14
fix: Support referencing complex elements in conditional expressions.

pleacu:
Fixes: #1604

### DIFF
--- a/ui/packages/atlasmap-core/src/models/document-definition.model.ts
+++ b/ui/packages/atlasmap-core/src/models/document-definition.model.ts
@@ -237,6 +237,10 @@ export class DocumentDefinition {
     return field;
   }
 
+  getComplexFields(): Field[] {
+    return this.discoverAllComplexFields(this.fields);
+  }
+
   getTerminalFields(): Field[] {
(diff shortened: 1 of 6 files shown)


pleacu:
@igarashitm - No real React UI changes.  Please review all files.

igarashitm:
Is it populating named `inputFieldGroup` with having children like this? runtime depends on those field structure to process expression.
https://github.com/atlasmap/atlasmap/blob/master/lib/itests/core/src/test/resources/mappings/atlasmapping-expression-filter-select-java.json#L19-L33

pleacu:
> Is it populating named `inputFieldGroup` with having children like this? runtime depends on those field structure to process expression.
> https://github.com/atlasmap/atlasmap/blob/master/lib/itests/core/src/test/resources/mappings/atlasmapping-expression-filter-select-java.json#L19-L33

Yep - you're right - inputField need to be inputFieldGroup.  I'll work on that and resubmit - pls take a quick look at the rest.

pleacu:
will resubmit - working on inputField -> inputFieldGroup.

Issue #2237, opened 2020-07-20
Introduce session scoped AtlasPropertyStrategy


igarashitm:
In order to achieve #2117, AtlasPropertyStrategy should read Camel message headers for each Camel Exchange. Currently AtlasPropertyStrategy is held by AtlasContextFactory thus it could be shared among multiple transactions, which could cause race condition. Introduce session level AtlasPropertyStrategy to avoid it.

Issue #2226, opened 2020-07-16
Add runtime support for mapping target property


igarashitm:
Allow mapping into target properties. Introduce target properties on AtlasSession with changing existing session properties to source properties, then store mapped key/values as target properties.

Issue #2210, opened 2020-06-29
Add target property in UI


igarashitm:
See #2197 and #2117 

What it's trying to do with #2197 also add the needs of setting target document scoped properties - which is Camel IN message header if it runs in Camel. Add target property in UI. It also needs scope dropdown (#2209) and have a `camelExchangeProperty` scope.

pleacu:
So - the target property doc will always exist in the target panel and operate pretty much the way the existing property doc does, correct?

![sc1](https://user-images.githubusercontent.com/2981421/88928331-b71e9580-d246-11ea-8827-2b426199b306.png)

pleacu:
@igarashitm - so this seems to be the appropriate JSON for source/target properties:

`    "properties" : {`
`"sourceProperty" : [ {`
`       "name" : "rat",`
`        "value" : "cat",`
`        "fieldType" : "STRING",`
`        "scope" : "camelExchangeProperty"`
`      } ],`
`      "targetProperty" : [ {`
`       "name" : "ball",`
`        "value" : "10",`
`        "fieldType" : "INTEGER",`
`        "scope" : "JSONSchemaSource"`
`      }, {`
`        "name" : "isround",`
`        "value" : "true",`
`       "fieldType" : "BOOLEAN",`
`        "scope" : ""`
`      } ]`
`    },`

The `value` is kept in for now.  I think it makes the most sens to establish `sourceProperty` and `targetProperty` with `properties`.
That means that all server references post serialization:

`atlasMapping.getProperties().getProperty()` needs to become `atlasMapping.getProperties().getSourceProperty()`.

igarashitm:
hmm... even if we remove `/atlasMapping/properties` completely with #2255? `/atlasMapping/properties` is there only to hold mapping definition level values for fallback.

igarashitm:
If you do this before #2255, I think it's OK to share those fallback property values between source and target. But it sounds like it's better to remove values first to avoid further confusion.

pleacu:
I need the `properties` JSON  so I can establish the `Properties` documents in the `Source` and `Target` panels without having a mapping selected.

pleacu:
I can't see sharing property values between source and target panels - users still need to be able to select them so they can add to their mapping.  I can remove the `value` now but I think we'll still need a `sourceProperty` and `targetProperty`.

igarashitm:
[...]

Issue #2209, opened 2020-06-29
Add property scope in UI


igarashitm:
See #2197 and #2117 

Allow user to choose a source document for scoping properties, which retrieves the properties within associated document scope. It would a dropdown box showing a list of Document name, and fill Document ID into PropertyField. The scope should be optional. Also, the special scope `camelExchangeProperty` should be the one of available scope.

pleacu:
@igarashitm - so on the `Create Property` dialog, users would be able to choose the source document to which it applies.  By optional then it would not have a default value.  WHen they click on it they see all loaded source panel documents + `camelExchangeProperty` - is that correct?

igarashitm:
Right, and whether to show "Current Message Header" explicitly or just have a blank item would be a UXD matter. Either way scope could be omitted. Or it's even fine to populate `current` explicitly.
https://github.com/atlasmap/atlasmap/blob/master/camel3/src/main/java/org/apache/camel/component/atlasmap/CamelAtlasPropertyStrategy.java#L38

pleacu:
@igarashitm - What would be a good hover-help for this?  Also - what's a good label?  Something simple like:

Scope: _____________________________
or
Camel Header Scope: ___________________________________

hover: 'Define the source document or exchange property for Camel Atlasmap headers associated with this property.  ?????

igarashitm:
I would avoid a word "Camel" for the general description like hover, since AtlasMap runtime in general is not limited to Camel. Eventually a list of scope might have to be configurable, but it's for future.
I would think something like "Choose property scope" is sufficient, or it would have to explain longer like "If it runs with Camel, blabla...". I think it's OK to leave it to User Guide.

pleacu:
Hey @igarashitm 
   I've got the UI code working but have an issue/ question on the generated JSON for the property scope.   The class `io.atlasmap.v2.Property` defines three fields -  `value`, `name`, `fieldType`.   The class `io.atlasmap.v2.PropertyField` defines a lot more - including `scope`.   This property UI dialog:

![sc1](https://user-images.githubusercontent.com/2981421/88583649-83612700-d01e-11ea-8100-ff5de6cdc799.png)


   Generates the following fragment:

   `"properties":{"property":[{"name":"paul","value":"rat","fieldType":"STRING"}]}`
 
   When I add property `paul` to the mapping I get the following fragment:

[...]

PR #2239, opened 2020-07-20
feat: Add runtime support for mapping camel message headers to AtlasM…

igarashitm:
…ap properties

Fixes: #2197
Fixes: #2226
Fixes: #2237

Introduced `scope` on `PropertyField` which specify a Document ID to map Document scoped property. Although DefaultAtlasPropertyStrategy doesn't handle scope, when it runs in Apache Camel, CamelAtlasPropertyStrategy is used and message header is retrieved from the Camel Message which correlates with the Document ID. If scope is not specified, it is retrieved from current incoming message header.
This change also enables target property mapping, where DefaultAtlasPropertyStrategy maps them into target AtlasMap properties held by AtlasSession. In Apache Camel, target properties are populated into outgoing Message header if scope is not specified.
Both of source and target property accepts a special scope `camelExchangeProperty` to map from/to Camel Exchange property when it runs in Apache Camel.

### DIFF
--- a/camel3/src/main/java/org/apache/camel/component/atlasmap/AtlasEndpoint.java
+++ b/camel3/src/main/java/org/apache/camel/component/atlasmap/AtlasEndpoint.java
@@ -16,7 +16,6 @@
  */
 package org.apache.camel.component.atlasmap;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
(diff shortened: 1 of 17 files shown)


igarashitm:
@pleacu here is the example mapping definition of source and target properties - should help for #2209 and #2210
https://github.com/atlasmap/atlasmap/pull/2239/files#diff-59f6003020f57669c1289292f8fab7f2
https://github.com/atlasmap/atlasmap/pull/2239/files#diff-828b82569909c3407d2de01fcc16be7dR195-R244

pleacu:
The netlify preview build was never deployed.

igarashitm:
cos ui code is not changed it says - https://github.com/atlasmap/atlasmap/pull/2239#issuecomment-661425452

PR #2250, opened 2020-07-28
fix: Support property scoping in the UI.

pleacu:
Fixes: #2209


lib/model/src/main/java/io/atlasmap/v2/Property.java
```diff
@@ -12,6 +12,8 @@
 
     protected FieldType fieldType;
 
+    protected String scope;
```
igarashitm, line 15: Was it easier to put `scope` here as well? This `Property` is used to define a fallback value apart from `PropertyField`, and `scope` only makes sense on `PropertyField`.

ui/packages/atlasmap/src/Atlasmap/utils/document.ts
```diff
@@ -329,3 +329,30 @@ export function enableCustomClass(
...
+    },
+    {
+      label: "Camel Exchange Property",
+      value: "camelExchangeProperty",
+    },
+  ];
+  for (let i = 0; i < cfg.sourceDocs.length; i++) {
+    scopeOptions.push({
+      value: cfg.sourceDocs[i].id,
+      label: cfg.sourceDocs[i].id,
```
igarashitm, line 354: Maybe `cfg.sourceDocs[i].name` for label?

igarashitm: Also, wonder if it's possible to show `cfg.sourceDocs[i].description` as a tooltip?

lib/model/src/main/java/io/atlasmap/v2/Property.java
pleacu: Needed to set the scope argument when editing an existing property.

PR #2258, opened 2020-08-03
[ENTESB-14445] fix initial property and constant value type

tplevko:
The reason for this is explained in detail in [ENTESB-14445](https://issues.redhat.com/browse/ENTESB-14445). 
Also I would like to ask, whether this could be backported to 7.8. branch (of course in case the fix I provided makes sense), as this is also present in 7.7. 
I know that master will differ from 7.8., but I'm not completely sure what the 7.8. branch is.

### DIFF
--- a/ui/packages/atlasmap/src/Atlasmap/dialogs/useConstantDialog.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/dialogs/useConstantDialog.tsx
@@ -12,9 +12,10 @@ export function useConstantDialog(
   const [onConstantCb, setOnConstantCb] = useState<ConstantCallback | null>(
     null,
   );
-  const [initialConstant, setInitialConstant] = useState<IConstant | null>(
-    null,
-  );
+  const [initialConstant, setInitialConstant] = useState<IConstant | null>({
+    value: "",
+    valueType: constantTypes[0][0],
+  });
   const { state, toggleOn, toggleOff } = useToggle(false);
   const onConfirm = useCallback(
     (constant: IConstant) => {
--- a/ui/packages/atlasmap/src/Atlasmap/dialogs/usePropertyDialog.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/dialogs/usePropertyDialog.tsx
@@ -13,9 +13,12 @@ export function usePropertyDialog(
   const [onPropertyCb, setOnPropertyCb] = useState<PropertyCallback | null>(
     null,
   );
-  const [initialProperty, setInitialProperty] = useState<IProperty | null>(
-    null,
-  );
+  const [initialProperty, setInitialProperty] = useState<IProperty | null>({
+    value: "",
+    name: "",
+    valueType: propertyTypes[0][0],
+    scope: getPropertyScopeOptions()[0].value,
+  });
   const { state, toggleOn, toggleOff } = useToggle(false);
   const onConfirm = useCallback(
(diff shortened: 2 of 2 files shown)


igarashitm:
Hi @tplevko thanks for the fix! can you create a GH issue with labels (ui, bug) and put cross reference link? We'll use 2.0.x branch for Fuse 7.8.

ui/packages/atlasmap/src/Atlasmap/dialogs/useConstantDialog.tsx
```diff
@@ -33,6 +33,7 @@ export function useConstantDialog(
         value,
         label,
       }))}
+      valueType={constantTypes[0][0]}
```
riccardo-forina, line 36: We have `initialConstant` to set the default, which is set where the dialog is actually set up instead of here. I'd rather keep this logic there.

ui/packages/atlasmap/src/Atlasmap/dialogs/usePropertyDialog.tsx
```diff
@@ -33,6 +33,7 @@ export function usePropertyDialog(
         value,
         label,
       }))}
+      valueType={propertyTypes[0][0]}
```
riccardo-forina, line 36: Same for this.

tplevko:
Thanks @pleacu and @igarashitm. I created xref issue here: https://github.com/atlasmap/atlasmap/issues/2259

ui/packages/atlasmap/src/Atlasmap/dialogs/useConstantDialog.tsx
tplevko: Thanks for the feedback @riccardo-forina, I updated the PR to utilize the `initialConstant`, could you please have a look?

ui/packages/atlasmap/src/Atlasmap/dialogs/usePropertyDialog.tsx
tplevko: same change as the above

ui/packages/atlasmap/src/Atlasmap/dialogs/usePropertyDialog.tsx
```diff
@@ -13,9 +13,10 @@ export function usePropertyDialog(
   const [onPropertyCb, setOnPropertyCb] = useState<PropertyCallback | null>(
     null,
   );
-  const [initialProperty, setInitialProperty] = useState<IProperty | null>(
-    null,
-  );
+  const [initialProperty, setInitialProperty] = useState<IProperty | null>({
+    valueType: propertyTypes[0][0],
+    scope: getPropertyScopeOptions()[0].value,
```
tplevko, line 18: Added also the initial `scope`. This didn't cause error mentioned in [ENTESB-14445](https://issues.redhat.com/browse/ENTESB-14445), but omitting this caused, that in case the initial value wasn't changed, the user would expect it was set to "Current Message Header" which was visible in the dropdown, but was actually empty string in the mapping file

igarashitm:
Hi @tplevko, can you also create a backport PR for 2.0.x branch?
https://github.com/atlasmap/atlasmap/tree/atlasmap-2.0.x

tplevko:
@igarashitm sure thing, the backport PR is here: https://github.com/atlasmap/atlasmap/pull/2278

Issue #2265, opened 2020-08-05
AtlasmapProvider should accept inspectionParameters for CSV


rkorytkowski:
If I understand it correctly, AtlasmapProvider is being used by Syndesis and if we want to enable passing CSV inspectionParameters from Syndesis we need to include that field in IExternalDocumentProps.

I expect I need to adjust https://github.com/syndesisio/syndesis/blob/c55f2a94660cb2501a9ae531168165e9055b42ad/app/ui-react/packages/atlasmap-adapter/src/DataMapperAdapter.tsx as well. What's the workflow here? Do I need to wait for the release of Atlasmap, then create a PR with a new version and DataMapperAdapter change?

rkorytkowski:
@igarashitm, @pleacu, @riccardo-forina please advise.

igarashitm:
Yes, in order to make a change there we need to make atlasmap release first, then bump it up in Syndesis.

PR #2266, opened 2020-08-05
fix: AtlasmapProvider should accept inspectionParameters for CSV

rkorytkowski:
Fixes: #2265


gashcrumb:
One follow-on question on this after I +1'd it, would all usages besides Syndesis of the atlasmap component need to be updated, i.e. the app that holds atlasmap standalone, vscode extension, storybook?

riccardo-forina:
No, that interface is for Syndesis only.

gashcrumb:
awesome, tx!

Issue #2255, opened 2020-07-30
UI Property: Remove 'value' field from properties dialog.


pleacu:
The property value embedded in a mapping definition is used as a default value, if the runtime value is not found.
The runtime value could be a Java system property, ENV with default property strategy and we're adding camel header/exchange property mapping on top of it.  Consequently anything the user typed for a 'value' could only be used as a default value in a fallback scenario.

igarashitm:
I would vote for removing `value` completely from Properties, regardless of scope. That default value is not so intuitive and have been a source of confusion vs. Constants. We might be able to add generic default value support for any mappings instead.

pleacu:
merged

Issue #2253, opened 2020-07-30
UI Property: Use name instead of id for source doc pulldown labels.


pleacu:
n ui/packages/atlasmap/src/Atlasmap/utils/document.ts:

`> +    value: string;`
`+    label: string;`
`+  }[] = [`
`+    {`
`+      label: "Current Message Header",`
`+      value: "current",`
`+    },`
`+    {`
`+      label: "Camel Exchange Property",`
`+      value: "camelExchangeProperty",`
`+    },`
`+  ];`
`+  for (let i = 0; i < cfg.sourceDocs.length; i++) {`
`+    scopeOptions.push({`
`+      value: cfg.sourceDocs[i].id,`
`+      label: cfg.sourceDocs[i].id,`

Maybe cfg.sourceDocs[i].name for label?

pleacu:
Merged

PR #2271, opened 2020-08-05
fix: Add target panel 'Properties' section and remove value from pro…

pleacu:
…perties dialog.

Fixes: #2210, #2253, #2255

### DIFF
--- a/lib/model/src/main/java/io/atlasmap/v2/Property.java
+++ b/lib/model/src/main/java/io/atlasmap/v2/Property.java
@@ -14,6 +14,8 @@ public class Property implements Serializable {
 
     protected String scope;
 
+    protected DataSourceType documentType;
+
     /**
      * Gets the value of the name property.
      * 
@@ -108,6 +110,30 @@ public class Property implements Serializable {
      */
     public void setScope(String scope) {
         this.scope = scope;
+    }
+
+    /**
+     * Gets the value of the documentType property.
+     *
+     * @return
+     *     possible object is
+     *     {@link String }
+     *
+     */
+    public DataSourceType getDocumentType() {
+        return documentType;
+    }
+
+    /**
+     * Sets the value of the documentType property.
+     *
+     * @param value
+     *     allowed object is
+     *     {@link String }
+     *
+     */
+    public void setDocumentType(DataSourceType value) {
+        this.documentType = value;
     }
 
 }
--- a/lib/model/src/main/java/io/atlasmap/v2/Property.java
+++ b/lib/model/src/main/java/io/atlasmap/v2/Property.java
@@ -14,7 +14,7 @@ public class Property implements Serializable {
 
     protected String scope;
 
-    protected DataSourceType documentType;
+    protected DataSourceType dataSourceType;
 
     /**
      * Gets the value of the name property.
@@ -113,27 +113,27 @@ public class Property implements Serializable {
     }
 
     /**
-     * Gets the value of the documentType property.
+     * Gets the value of the dataSourceType property.
      *
      * @return
      *     possible object is
      *     {@link String }
      *
      */
-    public DataSourceType getDocumentType() {
-        return documentType;
+    public DataSourceType getDataSourceType() {
+        return dataSourceType;
     }
 
     /**
-     * Sets the value of the documentType property.
+     * Sets the value of the dataSourceType property.
      *
      * @param value
      *     allowed object is
      *     {@link String }
      *
      */
-    public void setDocumentType(DataSourceType value) {
-        this.documentType = value;
+    public void setDataSourceType(DataSourceType value) {
+        this.dataSourceType = value;
     }
 
 }
--- a/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
@@ -160,8 +160,8 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
         handlers.onEditConstant({ value, valueType });
       },
       onDeleteConstant: handlers.onDeleteConstant,
-      onCreateSourceProperty: () => {
-        handlers.onCreateProperty(true);
+      onCreateProperty: (isSource: boolean) => {
+        handlers.onCreateProperty(isSource);
       },
       onEditProperty: (property, isSource) => {
         const [leftPart] = property.split(" ");
@@ -221,8 +221,8 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
       canRemoveFromSelectedMapping: (f) =>
         isFieldRemovableFromSelection("target", f),
       onRemoveFromSelectedMapping: onRemoveFromMapping,
-      onCreateTargetProperty: () => {
-        handlers.onCreateProperty(false);
+      onCreateProperty: (isSource: boolean) => {
+        handlers.onCreateProperty(isSource);
       },
       onEditProperty: (property, isSource) => {
         const [leftPart] = property.split(" ");
--- a/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import { AlertGroup } from "@patternfly/react-core";
-import React, { FunctionComponent, useCallback, useMemo, useRef } from "react";
+import React, { FunctionComponent, useCallback, useMemo } from "react";
 import { CanvasControlBar, MainLayout, ViewToolbar } from "../Layout";
 import {
   CanvasProvider,
@@ -82,12 +82,9 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
     onCreateMapping,
   } = useAtlasmap();
 
-  const isSourceRef = useRef<boolean>(true);
   const { handlers, dialogs } = useAtlasmapDialogs({
     modalContainer: document.getElementById(modalsContainerId)!,
-    isSource: isSourceRef.current,
   });
-
   const {
     activeView,
     showMappingColumn,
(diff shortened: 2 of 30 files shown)


ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
```diff
@@ -86,8 +82,10 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
     onCreateMapping,
   } = useAtlasmap();
 
+  const isSourceRef = useRef<boolean>(true);
```
riccardo-forina, line 85: Please no refs unless strictly necessary. Have distinct dialogs for sources and targets instead.

pleacu: Refactored to remove useRef reference.  Will engage @gashcrumb to review to not bug you on PTO!

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
```diff
@@ -85,100 +85,126 @@ export function getConstantTypeIndex(constVal: string): number {
...
 export function createProperty(
   propName: string,
-  propValue: string,
   propType: string,
   propScope: string,
+  isSource: boolean,
 ): void {
   const cfg = ConfigModel.getConfig();
-  let field = cfg.propertyDoc.getField(propName);
+  let field = isSource
```
gashcrumb, line 93: Just a question, this pattern of checking if it's a source/target document is repeated a few times in this function, would it be a bit more clear to refactor this logic into a higher level function, so that createProperty receives the field to be modified instead of it having to find the field and continually check this boolean?

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
```diff
@@ -85,100 +85,126 @@ export function getConstantTypeIndex(constVal: string): number {
...
-export function getPropertyTypeIndex(propName: string): number {
+export function getPropertyTypeIndex(
+  propName: string,
+  isSource: boolean,
+): number {
   const cfg = ConfigModel.getConfig();
-  const field = cfg.propertyDoc.getField(
-    cfg.propertyDoc.pathSeparator + propName,
-  );
+  const field = isSource
```
gashcrumb, line 201: 'cause yeah, I scrolled down from the last comment and it's 8 checks along these same lines.  Maybe there's a nicer way?

ui/packages/atlasmap/src/Views/ColumnMapperView/Columns/SourcesColumn.tsx
```diff
@@ -146,23 +148,28 @@ export const SourcesColumn: FunctionComponent<
...
-                      content={<div>Create a property for use in mapping</div>}
+                      content={
+                        <div>Create a source property for use in mapping</div>
+                      }
                       key={"create-property"}
                     >
                       <Button
-                        onClick={onCreateProperty}
+                        onClick={() => {
+                          onCreateSourceProperty();
```
gashcrumb, line 158: Should be able to just pass the `onCreateSourceProperty` reference here instead of using the lambda, no?

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
pleacu: I thought about ir but in this case you need to establish the `field` (create it or find it) then you need to decorate it and add it to either the source property doc or target property doc.  It reads more clearly (at least to me) the way it is.

ui/packages/atlasmap/src/Views/ColumnMapperView/Columns/TargetsColumn.tsx
```diff
@@ -101,6 +127,61 @@ export const TargetsColumn: FunctionComponent<
...
+                      position={"top"}
+                      enableFlip={true}
+                      content={
+                        <div>Create a target property for use in mapping</div>
+                      }
+                      key={"create-target-property"}
+                    >
+                      <Button
+                        onClick={() => {
+                          onCreateTargetProperty();
```
gashcrumb, line 148: lambda shouldn't be necessary here.

ui/packages/atlasmap/src/Views/ColumnMapperView/Columns/SourcesColumn.tsx
pleacu: yep- good catch - will fix.

ui/packages/atlasmap/src/Views/ColumnMapperView/Columns/TargetsColumn.tsx
pleacu: yep- good catch - will fix.

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
pleacu: Gotta respin - will take a shot at cleaning this up a bit.

gashcrumb: I should have said the document, not the field as the thing to be passed in, i.e. `createProperty` where instead of `isSource` the last argument would be the property document, so you could pass either `sourcePropertyDoc` or `targetPropertyDoc` into the function.  That strikes me as a cleaner approach.

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
```diff
@@ -85,100 +85,126 @@ export function getConstantTypeIndex(constVal: string): number {
...
+    cfg.sourcePropertyDoc.addField(field);
+  } else {
+    field.docDef = cfg.targetPropertyDoc;
+    cfg.targetPropertyDoc.addField(field);
+  }
   cfg.mappingService.notifyMappingUpdated();
 }
 
-export function deleteProperty(propName: string): void {
+export function deleteProperty(propName: string, isSource: boolean): void {
```
gashcrumb, line 114: Same suggestion as above applies, if you make the document to be operated on as an argument to this, you avoid having to check if it's a source/target in all these different places.

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
```diff
@@ -85,100 +85,126 @@ export function getConstantTypeIndex(constVal: string): number {
...
-}
-
-export function getPropertyValue(propName: string): string {
-  const cfg = ConfigModel.getConfig();
-  const field = cfg.propertyDoc.getField(
-    cfg.propertyDoc.pathSeparator + propName.split(" ")[0],
-  );
-  if (!field) {
-    return "";
+  if (isSource) {
```
gashcrumb, line 158: is this `isSource` supposed to be a function argument to `getPropertyValue`?

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
```diff
@@ -85,100 +85,126 @@ export function getConstantTypeIndex(constVal: string): number {
...
+    cfg.sourcePropertyDoc.updateField(field, "");
+  } else {
+    cfg.targetPropertyDoc.updateField(field, "");
   }
-  return field.value;
+  cfg.mappingService.notifyMappingUpdated();
 }
 
-export function getPropertyType(propName: string): string {
+export function getPropertyType(propName: string, isSource: boolean): string {
```
gashcrumb, line 166: other benefit of passing in the document as a function argument is that you don't have to access the config object I guess.

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
pleacu: Unfortunately I can't pass the document in as the config model is part of the business logic from atlasmap-core.  We've isolated it into field.ts but the UI component can only pass in `isSource` (which is why`cfg` is established in each function.

gashcrumb: Ah, okay gotcha.

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
gashcrumb: oh, spotted it on a 2nd look

ui/packages/atlasmap/src/Atlasmap/utils/field.ts
pleacu: So - `field.ts` is a difficult place to operate.  I was thinking of adding a function to update the source or property doc with the added field but I'd have to pass the args as 'any' since the model types are core specific - which we don't want.  I think it's original PR form is the best we can do while keeping with the type restrictions.

pleacu:
@igarashitm - quick check of lib/model/src/main/java/io/atlasmap/v2/Property.java ?

lib/model/src/main/java/io/atlasmap/v2/Property.java
```diff
@@ -14,6 +14,8 @@
 
     protected String scope;
 
+    protected DataSourceType documentType;
```
igarashitm, line 17: If you reuse `DataSourceType` then I would suggest to use `dataSourceType` as a property name here. Otherwise, 1) Mark `DataSourceType` deprecated and 2) Create a common enum `DocumentType`

ui/packages/atlasmap/src/Atlasmap/Atlasmap.tsx
```diff
@@ -164,19 +160,22 @@ export const Atlasmap: FunctionComponent<IAtlasmapProps> = ({
...
-          valueType,
-          scope,
-        });
+        handlers.onEditProperty(
+          {
+            name: leftPart,
+            valueType,
+            scope,
+          },
+          true,
```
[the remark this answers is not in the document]
pleacu: `isSource` is needed by the `PropertiesTree` tree walker to get the lines correct so went with that and have a single createProperty() instead of createSourceProperty/ createTargetProperty.

lib/model/src/main/java/io/atlasmap/v2/Property.java
pleacu: changed to dataSourceType

pleacu:
Given no further comments and having addressed the comments to date - I will merge this PR today.

PR #2278, opened 2020-08-10
Backport atlasmap-2.0.x: [ENTESB-14445] fix initial property and constant value type

tplevko:
Backport to atlasmap-2.0.x


igarashitm:
@tplevko Ah UI build failed, it seems you need to remove `scope` for 2.0.x as it's not available for 2.0.x.

tplevko:
@igarashitm, scope removed, should be ok now.

Issue #2277, opened 2020-08-10
Unexpected exception is thrown while reading source field


pleacu:
Community version of https://issues.redhat.com/browse/ENTESB-14432

pleacu:
@igarashitm  - I've got a simple fix for this but need your opinion.  When I reference the COMPLEX array in the conditional expression the generated mapping transition is `FOR_EACH`.  A mapping from two array elements to a target string generates a `MANY_TO_ONE` mapping transition.  The fix is then just to limit the conditional expression COMPLEX array refs to `FOR_EACH`:

![sc1](https://user-images.githubusercontent.com/2981421/89792633-d47b1b80-daf2-11ea-8a89-8975c98bdb03.png)

Can `FOR_EACH` be generated by a standard mapping action?

igarashitm:
So I think this `MappingMode` is only used in UI, to determine how Mapping Details need to be shown. I think it doesn't really make sense for Conditional Mapping, or we could introduce `MappingMode.EXPRESSION` maybe?

PR #2279, opened 2020-08-11
fix: Correct JSON mapping generation for complex expressions.

pleacu:
Fixes: #2277

Also had to correct an undefined inspectionParameters JSON field.  The generated JSON looks ok - Tomo - could you verify the runtime?

### DIFF
--- a/ui/packages/atlasmap-core/src/utils/mapping-serializer.ts
+++ b/ui/packages/atlasmap-core/src/utils/mapping-serializer.ts
@@ -283,7 +283,11 @@ export class MappingSerializer {
 
     let inputField = [];
     if (mappingJson.inputFieldGroup) {
-      mapping.transition.mode = TransitionMode.MANY_TO_ONE;
+      if (mappingJson.inputFieldGroup.fieldType === 'COMPLEX') {
+        mapping.transition.mode = TransitionMode.COMPLEX_EXPR;
+      } else {
+        mapping.transition.mode = TransitionMode.MANY_TO_ONE;
+      }
       inputField = mappingJson.inputFieldGroup.field;
 
       for (const field of inputField) {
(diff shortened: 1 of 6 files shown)


ui/packages/atlasmap-core/src/utils/mapping-serializer.ts
```diff
@@ -283,7 +283,11 @@ export class MappingSerializer {
 
     let inputField = [];
     if (mappingJson.inputFieldGroup) {
-      mapping.transition.mode = TransitionMode.MANY_TO_ONE;
+      if (mappingJson.inputFieldGroup.fieldType === 'COMPLEX') {
```
igarashitm, line 286: Can we just use TransitionMode.EXPRESSION whenever if expression is enabled? Is there a reason to distinguish the "expression consumes COMPLEX field" from other expression as a separate state?

igarashitm: I got a question related to this - what do we show if expression is enabled now? if it's MANY_TO_ONE, does it show a dropdown for many-to-one transformation drondown in Mapping Details? If so I think that's a bug, many-to-one transformation should not be selectable for expression mapping.

pleacu: Since a complex field reference doesn't appear in the mapping model as a field (it's an attr of the input field group) we need to distinguish a complex reference from a simple mapping where someone enable the expression box.

pleacu: > I got a question related to this - what do we show if expression is enabled now? if it's MANY_TO_ONE, does it show a dropdown for many-to-one transformation drondown in Mapping Details? If so I think that's a bug, many-to-one transformation should not be selectable for expression mapping.

The transformation dropdown is disabled in the expression box is enabled.

igarashitm: So TransitionMode doesn't make any sense if expression is enabled isn't it? anyway, the state "if there's at least one COMPLEX field in expression" doesn't fit in TransitionMode, maybe mapping.transition.expression.hasComplexField or something?

pleacu: Removed COMPLEX_EPXR from transition mode in favor of hasCOmplexField expression boolean.

Issue #2276, opened 2020-08-10
one_to_many mapping - repeat source value based on count 


rkorytkowski:
Initial PR: https://github.com/atlasmap/atlasmap/pull/2049

from PR: "one_to_many mapping - repeat source value based on count of fieldPath and do the target transformation."

rkorytkowski:
@igarashitm so I was looking into getting rid off the dropdown. The purpose of it is to point to an array to get its length and determine how many times the source value should be repeated. It's not for selecting the source value itself.

If we get rid of the dropdown, we need a different way to determine how many times to repeat the source value. I was looking into reading that info from the target side, but it may not be available at the point of setting the target field. Is that right? The target array final length could be determined after processing of all other fields.

I'm leaning towards keeping the dropdown and limiting listed fields to those of collection type. I would also include another parameter for specifying a count as an integer. The user would have to specify one or another.

Any thoughts on how to best approach it?

shanthi4647:
I agree.. i didn't see a way to repeat source field based on target array size, as we process source fields first and then move to processing target. so went for dropdown of source fields

igarashitm:
Sounds like it's a limited version of #781 where it only uses one of single collection length at source side, correct? so like `/foo<a>/bar<b>` => `/foo<a>` while our default index mapping is `/foo<a>/bar<b>` => `/foo<a*b>`. I don't like to introduce a separate UI for this specific asynmmetric option. So if we really want to achieve it we need generic #781 I would say.
Another option in the meantime is to do it with Conditional Mapping. It should be easier than #781 - the `length` transformation we already have could participate to get the specific collection length.

rkorytkowski:
It's a bit different. It's /foo/bar -> /foo<a>/bar, where you would basically put the value of bar "a" times.

igarashitm:
Ah OK, then it's just `length(/some/array<a>)` which is used as a parameter for that repeat action, should be possible with Conditional Mapping I guess?

igarashitm:
Yet another option is, at the runtime side, internally group these kind of repeating mapping depends on target array size, postpone them until other mapping finishes and then repeat for each target item that is already established by other mappings. An idea that the target array size is determined by the other mapping.

igarashitm:
The reason why I don't like to introduce this field dropdown is that, it then have to provide dynamic variable for transformation parameter which is not currently suported. If we go this route, we have to design generic dynamic transformation parameter support head of it.

[...]

PR #2287, opened 2020-08-13
fix: one_to_many mapping - repeat source value based on count

rkorytkowski:
Fixes: #2276

### DIFF
--- a/lib/itests/core/src/test/resources/mappings/atlasmapping-multiplicity-transformation-action_repeat.json
+++ b/lib/itests/core/src/test/resources/mappings/atlasmapping-multiplicity-transformation-action_repeat.json
@@ -1,0 +1,70 @@
+{
+  "AtlasMapping": {
+    "jsonType": "io.atlasmap.v2.AtlasMapping",
+    "dataSource": [
+      {
+        "jsonType": "io.atlasmap.json.v2.JsonDataSource",
+        "id": "json-source-repeat",
+        "uri": "atlas:json:json-source-repeat",
+        "dataSourceType": "SOURCE"
+      },
+      {
+        "jsonType": "io.atlasmap.json.v2.JsonDataSource",
+        "id": "json-target",
+        "uri": "atlas:json:json-target",
+        "dataSourceType": "TARGET"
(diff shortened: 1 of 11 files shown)


lib/itests/core/src/test/resources/mappings/atlasmapping-multiplicity-transformation-action_repeat.json
```diff
@@ -0,0 +1,76 @@
...
+    "mappings": {
+      "mapping": [
+        {
+          "jsonType": "io.atlasmap.v2.Mapping",
+          "inputFieldGroup": {
+            "jsonType": "io.atlasmap.v2.FieldGroup",
+            "actions": [
+              {
+                "Expression": {
+                  "expression": "Repeat(count(${1}), ${0})"
```
igarashitm, line 27: Like we talked in the meeting, let's test with new expression instead of this deprecated field action expression:
https://github.com/atlasmap/atlasmap/blob/master/lib/itests/core/src/test/resources/mappings/atlasmapping-expression-filter-select-java.json#L18

shanthi4647: @rkorytkowski  - Modified to new expression on 2 mapping files. please take the new files from below links and commit to the PR

https://github.com/shanthi4647/atlasmap/blob/2276/lib/itests/core/src/test/resources/mappings/atlasmapping-multiplicity-transformation-action_repeat.json

https://github.com/shanthi4647/atlasmap/blob/2276/lib/itests/core/src/test/resources/mappings/atlasmapping-multiplicity-transformation-action_repeat_nested_collection.json

rkorytkowski: @shanthi4647 I'm sorry for responding late. I was away for a while. I've applied your fixes. Thanks!

igarashitm:
@rkorytkowski is this ready to merge?

rkorytkowski:
Yes, it is!

Issue #2238, opened 2020-07-20
Custom field action argument issue


pleacu:
This appears to be a continuation of https://github.com/atlasmap/atlasmap/issues/1757

I've attempted to make an easier end-to-end custom field action argument example.  Here is the user field action Java source:

![sc1](https://user-images.githubusercontent.com/2981421/87979737-85683a80-caa0-11ea-81f4-a3d159f6efe8.png)

I build it as follows, modifying the classpath to reference the locally build Atlas service libraries:

`javac -cp /home/pleacu/git-clone/atlasmap/lib/api/target/classes:/home/pleacu/git-clone/atlasmap/lib/model/target/classes:. ./io/atlasmap/service/paul/PaulsFieldActions.java`
`jar cvf ../mfa.jar *`

I created a trivial ADM with a single source mapping to two target fields.  This zip also contains the mfa.jar created in step 1 above.
[exmpl.zip](https://github.com/atlasmap/atlasmap/files/4949775/exmpl.zip)

Upon loading the JAR file, this is the log from AtlasService:
![sc2](https://user-images.githubusercontent.com/2981421/87980605-e2b0bb80-caa1-11ea-8ea1-e8a6ce0998ae.png)

It loads fine - when I actually add the transformation I get the following error:
![sc3](https://user-images.githubusercontent.com/2981421/87980963-708ca680-caa2-11ea-9df4-a452dcc20d7d.png)

Perhaps another Jackson issue?

igarashitm:
Can you try splitting `Concur` class into a separate .java file?

pleacu:
Yes - same error with them split.  Ref: 
[exmpl2.zip](https://github.com/atlasmap/atlasmap/files/4949980/exmpl2.zip)

igarashitm:
It would be helpful to compare with this test case - it goes through with a custom field action and get 200 OK for validateMappingRequest
https://github.com/atlasmap/atlasmap/blob/master/lib/service/src/test/java/io/atlasmap/service/AtlasServiceTest.java#L177

pleacu:
I'll compare that json with what I'm generating.

pleacu:
Here's the network payload info.  I tried passing ` @type: "io.atlasmap.service.paul.Concur"` as well as ` @type: "Concur"` - both fail the same way.  The AtlasServiceTest passes the fully qualified name.

[...]

PR #2310, opened 2020-08-27
fix: Cannot add custom transformation

igarashitm:
Fixes: #2238


igarashitm:
Actually there were a few more point to get over, `AtlasLibraryLoader#reload()` after ADM import, eliminate incorrect action resolver reload, etc. It should be good with current jackson version...

Issue #2308, opened 2020-08-27
Target property modal shows source document in scope dropdown


igarashitm:
e.g. `JSONInstanceSource` is shown with this ADM
https://github.com/atlasmap/atlasmap/blob/master/camel3/src/test/resources/atlasmap-mapping.adm

Rather since we don't support multiple target documents when it runs with Camel, we should not show document names for target properties, i.e. only current message header and exchange property are possible.

pleacu:
![sc1](https://user-images.githubusercontent.com/2981421/91450754-a44daf80-e84a-11ea-9c89-416c0a0b603c.png)

PR #2314, opened 2020-08-31
fix: Correct error in target property scoping in create/edit dialogs.

pleacu:
Fixes: #2308


pleacu:
@jpav - can you review? - small fix.

pleacu:
@gashcrumb - could you pls review - people must be PTOing...

PR #2325, opened 2020-09-03
fix: Split/concatenate on CSV rows does not work correclty

### LINKED ISSUE #2069: Split/concatenate on CSV rows does not work correclty
Currently when you have a CSV as a source and you want to concatenate /<>/first_name and /<>/last_name, it doesn't work as expected. It basically produces a simple field with only first row taken into account, whereas I would expect all fields to be concatenated. It seems that concatenate/split is broken for root level collections.

@igarashitm, one question here. From the Atlasmap model perspective, is it more correct to represent a CSV field as /<>/first_name or /first_name<> ? I chose the former, but apparently root level collections are unusual (saw only one test in json).

rkorytkowski:
Fixes #2069


igarashitm (approved):
:+1:

Issue #2334, opened 2020-09-03
Show property scope in document tree


igarashitm:
2 `id` source properties are scoped to different source Document. Need a visual indicator to distinguish them.
![Screenshot from 2020-09-03 11-30-17](https://user-images.githubusercontent.com/265462/92135548-e5a90680-edd8-11ea-9b12-4fd2a307f95b.png)

pleacu:
typo in PR - still an issue

Issue #2330, opened 2020-09-03
D&D doesn't work for property>property mapping


igarashitm:
The source property `id` can't be dropped to target property `order-id`. Create new Mapping > select `/id` from candidate in Mapping Details works.
![Screenshot from 2020-09-03 10-49-29](https://user-images.githubusercontent.com/265462/92130885-3fa6cd80-edd3-11ea-9e42-e2b617dd6549.png)

pleacu:
typo on PR - this is still an issue

Issue #2327, opened 2020-09-03
Source field candidate shows target properties in Mapping Details


igarashitm:
![Screenshot from 2020-09-03 10-36-54](https://user-images.githubusercontent.com/265462/92129634-c9ee3200-edd1-11ea-98ea-df3943442c5b.png)

pleacu:
This is in the mapping details panel.  The target properties also don't appear in the `select target to add to the mapping`

PR #2339, opened 2020-09-03
fix: Correct target props handling:

pleacu:
- Correct sidebar addable sources/ targets
 - Correct mapped field processing to ignore property value
 - Correct property path to include '/'
Fixes: #2327
Fixes: #2330
Fixes: #2334


pleacu:
Sent to @igarashitm for him to review so we can get these into tomorrow's build

igarashitm (approved):
:+1:

PR #2343, opened 2020-09-08
Remove .gitmodules

cunningt:
Is the .gitmodules file still necessary?   We're experiencing some trouble in building because the build system tries to parse the .gitmodules file, thinking submodules are still active.     Asked @pleacu if this was still necessary and he thought it might not be, putting this in for @igarashitm to review at a later date.


igarashitm (approved):
Thanks! no it's just a garbage left years ago when we migrated to monorepo.

Issue #2333, opened 2020-09-03
UI randomly shows white page


igarashitm:
Not sure what's the condition to reproduce, sometimes it reloads without action and show it, sometimes after `Reset All`, etc. I got a error message in browser console when I got it after `Reset All`
```
Uncaught TypeError: n.current[r] is undefined
    unsetRef NodeRefProvider.tsx:74
    Xn NodeRefProvider.tsx:99
    au React
    unstable_runWithPriority scheduler.production.min.js:19
    React 4
    unstable_runWithPriority scheduler.production.min.js:19
    React 4
    unstable_runWithPriority scheduler.production.min.js:19
    React 5
    F AtlasmapProvider.tsx:151
    r AtlasmapProvider.tsx:365
    __tryOrUnsub Subscriber.ts:266
    next Subscriber.ts:208
    _next Subscriber.ts:140
    next Subscriber.ts:100
    next Subject.ts:71
    emitUpdatedEvent error-handler.service.ts:223
    clearAllErrors error-handler.service.ts:128
    resetAll error-handler.service.ts:116
    initialize initialization.service.ts:245
    initialize initialization.service.ts:243
    d toolbar.ts:168
NodeRefProvider.tsx:74:10

```

PR #2347, opened 2020-09-10
fix: Avoid undefined reference when calling unsetRef on reset command.

pleacu:
Fixes: #2333

Wasn't able to reproduce but from call stack this should correct the issue.


riccardo-forina (approved):
I'm a bit concerned since this shouldn't happen unless we have unexpected scenario with, eg., duplicated ids. But I guess this would be enough for the moment.

Issue #2198, opened 2020-06-18
Assign unique internal ID for Document ID apart from display name


igarashitm:
xref: https://issues.redhat.com/browse/ENTESB-12361

Import document in standalone UI uses class name for Java Document and schema file name for others as a Document ID, which could cause conflict. Assign unique ID internally apart from those display name.

igarashitm:
#2242 adds `name` and `description` to the `DataSource` model object, now UI just has to populate when it imports Document.

pleacu:
@igarashitm - can you give a brief example of exactly the format you'd like to see for doc.name and doc.description?

igarashitm:
@pleacu doc.name could be simply a class name without package for Java (e.g. Person, Status, Employee) and file name for others (e.g. person for person.json, Status for Status.xsd). description could be longer, a couple of sentenses which describe about the document.

PR #2359, opened 2020-09-16
fix: Handle import of common document name (JSON/XML) by using data source ID a…

pleacu:
…nd doc ID - also add doc desc.

Fixes: #2198
ENTESB-12361


ui/packages/atlasmap-core/src/services/initialization.service.ts
```diff
@@ -650,7 +657,11 @@ ${error.status} ${error.statusText}`,
     inspectionParameters?: { [key: string]: string }
   ): DocumentDefinition {
     const model: DocumentInitializationModel = new DocumentInitializationModel();
-    model.id = name;
+    model.name = name;
+    model.id =
```
igarashitm, line 661: I'd assing a GUID to allow mutliple same-type&same-name documents.

pleacu: Are we sure we want to allow multiple documents with the exact same name to be imported into the panels?  The user wanted `address.json` and `address.xml` to be allowed.  I don't see a demand for 100 `address.json` documents and adding a global unique ID will add to the complexity.

igarashitm: Yes that has to be possible. For example retrieving address object twice but with a different condition, where schema is same.

pleacu:
Making a few more changes than just an update to the old PR - will resubmit

PR #2371, opened 2020-09-21
fix: Handle common document name (JSON/XML) by using data source ID a…

pleacu:
…nd doc ID using GUID - also add doc desc.

Fixes: #2198
ENTESB-12361


igarashitm (approved):
:+1:

Issue #2422, opened 2020-10-14
Preview contains unexpected 'Space' terms with Combine mappings


jpav:
Related to [ENTESB-14981](https://issues.redhat.com/browse/ENTESB-14981)

jpav:
Unreproducible. Initial problem due to bad mapping file.

PR #2423, opened 2020-10-15
fix: Preview contains unexpected 'Space' terms with Combine mappings

jpav:
Fixes: #2422 

This does NOT fix the issue that the delimiter name, 'Space', is being used as the delimiter rather than an actual space. I need to research the expected behavior more for this, so will create a subsequent issue for that.


lib/core/src/test/java/io/atlasmap/core/DefaultAtlasContextTest.java
```diff
@@ -578,7 +578,7 @@ public void testProcessPreviewConcatenate() throws AtlasException {
         target.setFieldType(FieldType.STRING);
         m.getOutputField().add(target);
         context.processPreview(m);
-        assertEquals("-one--two--six", target.getValue());
+        assertEquals("one-two-six", target.getValue());
```
igarashitm, line 581: This looks wrong, as indexes are skipping multiple delimiter should be there

Issue #2420, opened 2020-10-14
Implement processPreview in the runtime service based on mapping expression.


pleacu:
In this example I have a simple `sourceJsonDouble + 1.0` expression. The preview results don't show the expression result correctly (don't get 100.0).

<pre><font color="#1E8C32">2020-10-13 10:05:59.889 </font><font color="#6AC911">DEBUG</font> <font color="#75507B">15727</font> --- [  XNIO-1 task-5] <font color="#06989A">io.atlasmap.service.AtlasService        </font> : Preview request: {
  &quot;Mapping&quot; : {
    &quot;jsonType&quot; : &quot;io.atlasmap.v2.Mapping&quot;,
    &quot;expression&quot; : &quot;${sourceJson:/sourceJsonDouble}+ 1.0&quot;,
    &quot;inputField&quot; : [ {
      &quot;jsonType&quot; : &quot;io.atlasmap.json.v2.JsonField&quot;,
      &quot;value&quot; : &quot;99.0&quot;,
      &quot;docId&quot; : &quot;sourceJson&quot;,
      &quot;path&quot; : &quot;/sourceJsonDouble&quot;,
      &quot;fieldType&quot; : &quot;NUMBER&quot;,
      &quot;name&quot; : &quot;sourceJsonDouble&quot;,
      &quot;userCreated&quot; : false
    } ],
    &quot;outputField&quot; : [ {
      &quot;jsonType&quot; : &quot;io.atlasmap.json.v2.JsonField&quot;,
      &quot;value&quot; : &quot;99.0&quot;,
      &quot;docId&quot; : &quot;targetJson&quot;,
      &quot;path&quot; : &quot;/targetJsonString&quot;,
      &quot;fieldType&quot; : &quot;STRING&quot;,
      &quot;name&quot; : &quot;targetJsonString&quot;,
      &quot;userCreated&quot; : false
    } ],
    &quot;id&quot; : &quot;preview&quot;
  }
}
<font color="#1E8C32">2020-10-13 10:05:59.889 </font><font color="#6AC911">DEBUG</font> <font color="#75507B">15727</font> --- [  XNIO-1 task-5] <font color="#06989A">io.atlasmap.service.AtlasService        </font> : Preview outcome: {
  &quot;ProcessMappingResponse&quot; : {
    &quot;jsonType&quot; : &quot;io.atlasmap.v2.ProcessMappingResponse&quot;,
    &quot;mapping&quot; : {
[...]

PR #2426, opened 2020-10-15
fix: Support expression in mapping preview

igarashitm:
Fixes: #2420


pleacu (approved):
+1 - I like getting the expression arguments from with a field or field group

igarashitm:
Found a problem, the Document ID still has to be taken into account for field reading to distinguish same path field in different doc... will work on it

Issue #2442, opened 2020-10-27
Enablement of imported Java inner class without dependant JARs should give a more precise message.


pleacu:
1.  Import camel-profobuf2avro-producer-1.0.jar
2.  Enable demo.camel.TransactionProtos$Transaction

![sc1](https://user-images.githubusercontent.com/2981421/97308565-6283ba00-1837-11eb-9ab7-7a726ce0a904.png)

![sc2](https://user-images.githubusercontent.com/2981421/97308587-69aac800-1837-11eb-9f3f-b01cf6350679.png)

[camel-profobuf2avro-producer-1.0.jar.zip](https://github.com/atlasmap/atlasmap/files/5445661/camel-profobuf2avro-producer-1.0.jar.zip)

pleacu:
As Tomo has found - there's a CNFE on `message:com/google/protobuf/MessageOrBuilder`

<pre><font color="#1E8C32">2020-10-27 14:23:09.087 </font><font color="#6AC911">DEBUG</font> <font color="#75507B">27565</font> --- [  XNIO-1 task-2] <font color="#06989A">io.atlasmap.service.AtlasLibraryLoader  </font> : Class not found: [ClassLoader:java.net.URLClassLoader@5b84fbae, Class name:demo.camel.TransactionProtos$Transaction, message:com/google/protobuf/MessageOrBuilder]
<font color="#1E8C32">2020-10-27 14:23:09.087 </font><font color="#6AC911">DEBUG</font> <font color="#75507B">27565</font> --- [  XNIO-1 task-2] <font color="#06989A">io.atlasmap.service.AtlasLibraryLoader  </font> : Class not found: [ClassLoader:sun.misc.Launcher$AppClassLoader@18b4aac2, Class name:demo.camel.TransactionProtos$Transaction, message:demo.camel.TransactionProtos$Transaction]
<font color="#1E8C32">2020-10-27 14:23:09.088 </font><font color="#6AC911">DEBUG</font> <font color="#75507B">27565</font> --- [  XNIO-1 task-2] <font color="#06989A">io.atlasmap.java.service.JavaService    </font> : Class inspection response: {
  &quot;ClassInspectionResponse&quot; : {
    &quot;jsonType&quot; : &quot;io.atlasmap.java.v2.ClassInspectionResponse&quot;,
    &quot;javaClass&quot; : {
      &quot;jsonType&quot; : &quot;io.atlasmap.java.v2.JavaClass&quot;,
[...]

PR #2447, opened 2020-10-27
chore: Slightly better error message in case transitive dependency is…

igarashitm:
… not found for Java inspection (#2442)

noclassdeffounderror


pleacu (approved):
+1

Issue #2446, opened 2020-10-27
Support `java.lang.CharSequence` as a `FieldType.STRING`


igarashitm:
ATM it's handled as `FieldType.COMPLEX`

PR #2448, opened 2020-10-28
fix: Support java.lang.CharSequence and its children as FieldType.STRING

igarashitm:
Fixes: #2446


pleacu (approved):
Wow - ton of type to type conversions but mostly mechanical.  Good job.
+1

Issue #2462, opened 2020-11-03
UI: Use relative paths for post-primary arguments of select/filter.


pleacu:
Restrict the scope of subsequent fields  (dynamic path expression).  i.e.
`"expression" : "SELECT(FILTER(${sourceArrays:/jsonObjects<>},${sourceArrays:/jsonObjects<>/key}>4),${sourceArrays:/jsonObjects<>/value})",`

should be

`"expression" : "SELECT(FILTER(${sourceArrays:/jsonObjects<>},${/key}>4),${/value})",`

Ref:
https://issues.redhat.com/browse/ENTESB-14457
https://github.com/atlasmap/atlasmap/issues/1604

pleacu:
Here is an example for `filter & selec`t function that relies on dynamic path expression.
https://github.com/atlasmap/atlasmap/blob/master/lib/itests/core/src/test/resources/ma
ppings/atlasmapping-expression-filter-select-java.json#L18

Note that first path expression is absolute (include documentId) and the others are relative.

pleacu:
Updated mapping JSON:

      "mapping" : [ {
        "jsonType" : "io.atlasmap.v2.Mapping",
        "expression" : "select( filter( ${sourceArrays:/jsonObjects<>}, ${/key} > 4), ${/value})",
        "inputFieldGroup" : {
          "jsonType" : "io.atlasmap.v2.FieldGroup",
          "actions" : [ ],
          "field" : [ {
            "jsonType" : "io.atlasmap.v2.FieldGroup",
            "actions" : [ ],
            "docId" : "sourceArrays",
            "path" : "/jsonObjects<>",
            "fieldType" : "COMPLEX",
            "field" : [ {
              "jsonType" : "io.atlasmap.json.v2.JsonField",
              "docId" : "sourceArrays",
              "index" : 0,
              "path" : "/jsonObjects<>/key",
              "fieldType" : "INTEGER",
              "name" : "key",
              "userCreated" : false
            } ]
          }, {
            "jsonType" : "io.atlasmap.v2.FieldGroup",
            "actions" : [ ],
            "docId" : "sourceArrays",
            "path" : "/jsonObjects<>",
            "fieldType" : "COMPLEX",
            "field" : [ {
              "jsonType" : "io.atlasmap.json.v2.JsonField",
              "docId" : "sourceArrays",
              "index" : 1,
              "path" : "/jsonObjects<>/value",
              "fieldType" : "STRING",
              "name" : "value",
[...]

PR #2464, opened 2020-11-06
fix: Use relative paths for post-primary collection arguments in gene…

pleacu:
…rated mapping JSON.

Fixes: #2462

### DIFF
--- a/ui/packages/atlasmap-core/src/models/expression.model.ts
+++ b/ui/packages/atlasmap-core/src/models/expression.model.ts
@@ -62,7 +62,8 @@ export class FieldNode extends ExpressionNode {
     private mapping: MappingModel,
     public mappedField?: MappedField | null,
     public metaStr?: string,
-    index: number = 0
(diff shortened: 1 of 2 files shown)


igarashitm (approved):

igarashitm:
I'm not really sure which cases will be supported with this and what not. But in order to permanently solve this, we'd need to provide metadata for the functions and let UI understand which argument of the function accepts absolute path or relative path. Or we could add a REST endpoint that "normalize" the arguments, where UI just passes absolute path for everything, and backend returns by translating into relative path as needed. Either case we need to establish function metadata that describes arguments & return types.

igarashitm:
For example, could this be supported?
"expression" : "SELECT(FILTER(${sourceArrays:/jsonObjects<>},${/key/value}>4),${/value/value})",
`/key/value` is still a relative path, but key is a complex field.

pleacu:
I'll generate a new issue to track this - totally agree.

Issue #2267, opened 2020-08-05
Atlasmap CSV modal window redesign


rkorytkowski:
See https://issues.redhat.com/browse/ENTESB-14446

Basically we want to have the same workflow as for adding transformations.

PR #2468, opened 2020-11-10
fix: Atlasmap CSV modal window redesign

rkorytkowski:
Fixes: #2267

image


pleacu:
Hmmm - not a big fan of true/false pull-down menus.  Can we use a checkbox?  i.e. 
https://www.patternfly.org/v4/components/checkbox/html/#examples

rkorytkowski:
@pleacu, right, the question how to factor that in the design? Add a checkbox to the right of the dropdown with "Enabled" label? Or a checkbox to the left of the dropdown without a label? Both options feel kind of awkward to me. Maybe two radio boxes "yes" and "no" to the right of the dropdown would feel better?

pleacu:
Subjective design issues are always  tricky and getting UXD input is probably warranted *but* I'd say having the checkbox down the left preceding the CSV parameters without an `Enabled` label seems most obvious.  Also - if you have the hood up I think its better to have human rfeadable labels for the parameters - i.e.

`Allow duplicate header names'

instead of 

`allowDuplicateHeaderNames'

riccardo-forina:
+1 for human readable labels, we really need them. There is a [switch](https://www.patternfly.org/2020.08/documentation/core/components/switch) component that I think would be ok in this case.

We also have [guidelines](https://www.patternfly.org/v4/components/form/design-guidelines) regarding forms, in terms of alignment, labels, etc. to which we have to adhere.

rkorytkowski:
Updated PR!
Screenshot from 2020-11-19 12-21-37

pleacu:
Hey - I like it - it's still tweekable (maybe make the whole dialog wider so you can read the labels or possibly a smaller font side).  but it's fine with me.

gashcrumb:
I can't put my finger on what's bugging me with the trash icons.

rkorytkowski:
Increased the dialog size...

image

@gashcrumb, shoot ;) I see that the switch is slightly higher compared to the trash icon, but I couldn't find any variable to position it lower... As far as I can tell it would require using some custom positioning which may not play well with resizing etc. I'm not a css wizard so I may be missing something...

pleacu:
Hey Rafal - I think we can address the offset for the check/slider buttons with another issue.  As for the width - it's not necessary that the entire label can fit - just that a reasonable amount can fit.  I'd say cut the `Default` menu drop down widget in half - that should do it.

gashcrumb:
Yep, +1 handling the alignment in another issue.  Maybe some of the layout components could help here like [Level](https://www.patternfly.org/v4/layouts/level#basic) perhaps?

rkorytkowski:
image

Corrected width and position of the switch icon.

pleacu (approved):
+1

PR #2470, opened 2020-11-12
fix: Add option for concatenate action to skip delimiters for null values

### LINKED ISSUE #2430: Add option for concatenate action to skip delimiters for null values

jpav:
Fixes #2430

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/actions/StringComplexFieldActions.java
+++ b/lib/core/src/main/java/io/atlasmap/actions/StringComplexFieldActions.java
@@ -72,10 +72,13 @@ public class StringComplexFieldActions implements AtlasFieldAction {
         }
 
         String delim = concat.getDelimiter() == null ? "" : concat.getDelimiter();
-        StringBuilder builder = new StringBuilder();
+        boolean delimitingEmptyValues = concat.getDelimitingEmptyValues() == null
+            ? false
+            : concat.getDelimitingEmptyValues();
         boolean isFirst = true;
+        StringBuilder builder = new StringBuilder();
         for (String entry : inputs) {
-            if (!isFirst) {
+            if (!isFirst && ((entry != null && !entry.isEmpty()) || delimitingEmptyValues)) {
                 builder.append(delim);
             }
             if (entry != null) {
--- a/ui/packages/atlasmap/src/UI/MappingTransformation.tsx
+++ b/ui/packages/atlasmap/src/UI/MappingTransformation.tsx
@@ -9,6 +9,7 @@ import {
   FormGroup,
   Split,
   SplitItem,
+  Checkbox,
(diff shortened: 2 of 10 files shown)


lib/core/src/main/java/io/atlasmap/actions/StringComplexFieldActions.java
```diff
@@ -72,10 +72,13 @@ public static String concatenate(Concatenate concat, List<String> inputs) {
         }
 
         String delim = concat.getDelimiter() == null ? "" : concat.getDelimiter();
-        StringBuilder builder = new StringBuilder();
+        boolean delimitingEmptyValues = concat.getDelimitingEmptyValues() == null
```
pleacu, line 75: Isn't this just 
boolean delimitingEmptyValues = !(concat.getDelimitingEmptyValues() == null)

ui/packages/atlasmap/src/UI/MappingTransformation.tsx
```diff
@@ -144,6 +146,16 @@ export const MappingTransformation: FunctionComponent<IMappingTransformationProp
               />
             )}
           </FormGroup>
+        ) : (
```
pleacu, line 149: Can you paste a screen shot as to what this looks like.  Also I'm not sure I like `Delimit empty values`.  It's terse which is typically a good thing but maybe too much.  Maybe:
`Add the specified delimiter for empty values.`

pleacu (commented):
Adding @igarashitm so he can have a look at the lib side.  He should be resurfacing next week.  I'm concerned about the requirement to version the mapping file.  We've talked about it before - all-in-all probably a good idea - not sure this is the reason for it.

igarashitm:
No need for mapping def version as it's just a minor addition which doesn't break backward compatibility.

ui/packages/atlasmap/src/UI/MappingTransformation.tsx
jpav: I don't want to take up too much space for this, as there is already very little real estate to work with or that we are forcing user to expand to.

lib/core/src/main/java/io/atlasmap/actions/StringComplexFieldActions.java
jpav: Not sure if I understand your comment. Based on what you've suggested, if getDelimitingEmptyValues returns False, the variable would end up as true.

pleacu: You're right - no ! needed.  Just trying to avoid calling `getDelimitingEmptyValues()` twice.

pleacu (approved):
+1 - looks ok

Issue #1518, opened 2019-11-22
Ghost field added to FHIR mockdoc, which gets duplicated `@` XML attribute prefix as it reloads


igarashitm:
![reload-attr-dup](https://user-images.githubusercontent.com/265462/69424195-04308c00-0cf6-11ea-80cf-f7c049d74c4a.gif)

igarashitm:
It's entering here
https://github.com/atlasmap/atlasmap/blob/4f8901fd59b4a374c829168a89304490ffa45860/ui/src/app/lib/atlasmap-data-mapper/utils/mapping-util.ts#L94

which means this doesn't return proper field
https://github.com/atlasmap/atlasmap/blob/4f8901fd59b4a374c829168a89304490ffa45860/ui/src/app/lib/atlasmap-data-mapper/utils/mapping-util.ts#L80

So my guess is that the mapping deserialization is happening before document definition is completely built.

igarashitm:
TLDR I put this on hold until React migration is settled.

This `fetchDocuments()` is async work
https://github.com/atlasmap/atlasmap/blob/c9977f167d3926be913457e342f3ea896e084173/ui/src/app/lib/atlasmap-data-mapper/services/initialization.service.ts#L203

And it's possible that this `fetchMappings()` starts to deserialize mappings before document initialization by `fetchDocuments()` is completed.

I thought we can subscribe `initializationStatusChanged$` and wait for `ConfigModel.initCfg.initialized` comes to true, but then it doesn't happen until field action service is initialized.
https://github.com/atlasmap/atlasmap/blob/c9977f167d3926be913457e342f3ea896e084173/ui/src/app/lib/atlasmap-data-mapper/services/initialization.service.ts#L534

The field action service initialization is delayed on purpose to wait for custom field actions to be loaded when .adm is uploaded. We have to untangle this initialization statuses, basically in following order:
1. library upload and classpath refresh should happen first, then
1. document initialization
1. field action service
1. mappings

igarashitm:
A good news is that, this is standalone only issue, as in Syndesis mapping definition is passed in in UI, but not loading from .adm file.

igarashitm:
This issue is not solved yet. Note that the ghost field `/@value` is added to the FHIR document. Also the real attribute doesn't get `@` prefix it seems.
![output](https://user-images.githubusercontent.com/265462/102078545-6f438800-3e4e-11eb-8eff-90b679ada014.gif)

[...]

PR #2511, opened 2021-01-05
fix: Correct ghost field addition to COMPLEX fields.

pleacu:
Fixes: #1518

The real issue here is that the processMappingsCatalogFiles call in initialize needed to finish before the mappings are loaded.

Will issue a different issue for differentiating attributes from field elements.


igarashitm (approved):
:+1:

Issue #2512, opened 2021-01-06
Distinguish elements and attributes of the same name in XML documents.


pleacu:
While we need a way to distinguish element and attribute for XML document, I don't think adding @ prefix to the field name is the best way. Instead for example, we can add an icon that indicates it's an attribute.

/myelement/@mychild and /myelement/mychild have to show differently in UI.

`<myelement mychild="one">`
`  <mychild>the other</mychild>`
`</myelement>`

Ref: https://github.com/atlasmap/atlasmap/issues/1518#issuecomment-744394433

pleacu:
Hey @igarashitm - here is the updated icon for attributes in the canvas field display:

![sc1](https://user-images.githubusercontent.com/2981421/103819055-b2e97600-5037-11eb-9d13-edba08c6d1fb.png)

Do you have a good xsd/xml example that has both attributes and field elements?  I suck at xsd writing...

igarashitm:
Here is a generated XSD from the example in the description
https://www.freeformatter.com/xsd-generator.html
```xsd
<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:element name="myelement">
    <xs:complexType>
      <xs:sequence>
        <xs:element type="xs:string" name="mychild"/>
      </xs:sequence>
      <xs:attribute type="xs:string" name="mychild"/>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

pleacu:
Ah - cool - so does all of this look ok?

![sc2](https://user-images.githubusercontent.com/2981421/103899108-e8d43c00-50c3-11eb-9013-8199dbc1feab.png)

<pre>    &quot;mappings&quot; : {
      &quot;mapping&quot; : [ {
        &quot;jsonType&quot; : &quot;io.atlasmap.v2.Mapping&quot;,
        &quot;inputField&quot; : [ {
          &quot;jsonType&quot; : &quot;io.atlasmap.xml.v2.XmlField&quot;,
          &quot;docId&quot; : &quot;i2512-5c4f7408-a1f3-4d72-aa8f-d39da45b1ccc&quot;,
          &quot;path&quot; : &quot;/myelement/@mychild&quot;,
          &quot;fieldType&quot; : &quot;STRING&quot;,
          &quot;name&quot; : &quot;mychild&quot;,
          &quot;attribute&quot; : true
        } ],
[...]

PR #2513, opened 2021-01-07
fix: Distinguish elements and attributes of the same name in XML docu…

pleacu:
…ments.

Fixes: #2512

Also has some auto changes to developer guide adocs.


igarashitm (approved):
:+1:

Issue #2528, opened 2021-01-18
`isAttribute` on JsonField causes runtime error


igarashitm:
`isAttribute` should be placed only on `XmlField`. It doesn't exist on other field types.

> javax.ws.rs.WebApplicationException: HTTP 400 Bad Request
	at io.atlasmap.service.AtlasService.fromJson(AtlasService.java:666) ~[atlas-service-2.2.0-SNAPSHOT.jar:2.2.0-SNAPSHOT]
	at io.atlasmap.service.AtlasService.validateMappingRequest(AtlasService.java:521) ~[atlas-service-2.2.0-SNAPSHOT.jar:2.2.0-SNAPSHOT]
.......
Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "attribute" (class io.atlasmap.json.v2.JsonField), not marked as ignorable (16 known properties: "path", "arrayDimensions", "primitive", "name", "value", "index", "docId", "fieldType", "format", "collectionType", "arraySize", "actions", "required", "typeName", "status", "userCreated"])
 at [Source: (String)"{"AtlasMapping":{"jsonType":"io.atlasmap.v2.AtlasMapping","dataSource":[{"jsonType":"io.atlasmap.json.v2.JsonDataSource","id":"JSONSchemaSource-d94bd9f3-39b2-4829-9ec1-bd9fcc4158ce","name":"JSONSchemaSource","description":"Source document JSONSchemaSource type: JSON","uri":"atlas:json:JSONSchemaSource-d94bd9f3-39b2-4829-9ec1-bd9fcc4158ce","dataSourceType":"SOURCE"},{"jsonType":"io.atlasmap.json.v2.JsonDataSource","id":"JSONSchemaSource-a54feb7a-c4a3-4b05-9e76-2c7944793ee6","name":"JSONSchemaSour"[truncated 1337 chars]; line: 1, column: 964] (through reference chain: io.atlasmap.v2.AtlasMapping["mappings"]->io.atlasmap.v2.Mappings["mapping"]->java.util.ArrayList[0]->io.atlasmap.v2.Mapping["inputField"]->java.util.ArrayList[0]->io.atlasmap.json.v2.JsonField["attribute"])
.......

PR #2529, opened 2021-01-18
fix: Limit service object type 'attribute' to XML service package fie…

pleacu:
…lds.

Fixes: #2528


igarashitm (approved):
:+1: Thanks for the quick fix!

Issue #2307, opened 2020-08-26
Drag&Drop between Source/Target panels not functioning in latest Chrome


pleacu:
Attempting to d&d just marks the field element instead of dragging it.  Works o kay on FireFox and WIndows.

![sc1](https://user-images.githubusercontent.com/2981421/91319028-e7941980-e789-11ea-8bc5-4211fff38553.png)

pleacu:
Google Chrome Version 85.0.4183.83 (Official Build) (64-bit)

PR #2554, opened 2021-02-02
fix: Corrects d&d on Chrome.

pleacu:
Fixes: #2307

@riccardo-forina - please see if d&d still behaves correctly on Apple - don't have one to try.  This now works on Chrome and Firefox ok,  It's triggering through TouchBackend (enableMouseEvents = true).


igarashitm (approved):
Let's go ahead if it works for chrome and firefox - if there's a problem with safari, we can file an another issue.

riccardo-forina:
It seems to work correctly for what I can see on Safari Desktop and Safari iOS.

PR #2556, opened 2021-02-04
fix: Create new mapping button from Target panel does not work

### LINKED ISSUE #2533: Create new mapping button from Target panel does not work
Clicking this button on any target field seems to have no effect.

jpav:
Fixes #2533


pleacu (approved):
+1

Issue #2563, opened 2021-02-09
XML namespace not persisted into DataSource


igarashitm:
Seeing weird issue with this XML schema 
https://github.com/atlasmap/atlasmap/blob/master/ui/test-resources/xml/schema/po-example-schema.xsd

the target namespace `http://tempuri.org/po.xsd` should be persisted into DataSource in the mapping definition, but not
```json
{
  "jsonType" : "io.atlasmap.v2.DataSource",
  "id" : "po-example-schema-953958cd-728b-48bb-b62c-01eb54816471",
  "name" : "po-example-schema",
  "description" : "Source document po-example-schema type: XSD",
  "uri" : "atlas:xml:po-example-schema-953958cd-728b-48bb-b62c-01eb54816471",
  "dataSourceType" : "SOURCE"
}
```

Note that it does with this schema https://github.com/atlasmap/atlasmap/blob/master/ui/test-resources/xml/schema/XMLSchemaSource.xml
```json
{
  "jsonType" : "io.atlasmap.xml.v2.XmlDataSource",
  "id" : "XMLSchemaSource-857eba0d-d35b-442c-9af2-064b059c16ff",
  "name" : "XMLSchemaSource",
  "description" : "Source document XMLSchemaSource type: XML",
  "uri" : "atlas:xml:XMLSchemaSource-857eba0d-d35b-442c-9af2-064b059c16ff",
  "dataSourceType" : "SOURCE",
  "xmlNamespaces" : {
    "xmlNamespace" : [ {
      "alias" : "tns",
      "uri" : "http://syndesis.io/v1/swagger-connector-template/request"
    } ]
  }
}
```
I can see inspection returns namespace in this test, so it looks like an UI issue.
https://github.com/atlasmap/atlasmap/blob/master/lib/modules/xml/core/src/test/java/io/atlasmap/xml/inspect/XmlSchemaInspectionTest.java#L182

igarashitm:
Oh wait look at that, its `jsonType` shows `io.atlasmap.v2.DataSource` while it should be `io.atlasmap.xml.v2.XmlDataSource` like XMLSchemaSource. Something wrong.

pleacu:
We were only pushing XML files into DataSource as `io.atlasmap.xml.v2.XmlDataSource`, not XSD.

Now it looks like this:

<pre>   &quot;dataSource&quot; : [ {
[...]

Issue #832, opened 2019-03-18
Let user choose if importing Document is schema or instance


igarashitm:
I found angular.json has `/$schema` property while it still should be treated as a instance document. It might be better to look at the value as well and check if it starts with `http://json-schema.org/schema` to say it's JSON schema.

Furthermore, there could be an usecase to use JSON schema as an instance, so JSON schema is used as a payload. We should let user choose if it's schema or instance. Same for XML, XML schema could be used as an instance.

So instead of opening file dialog directly from + button, we need an intermediate modal which has a toggle to select instance/schema, and file upload.

pleacu:
Currently the XSD/XML decision is based off of the imported document suffix - ref: https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap-core/src/services/document-management.service.ts#L441

This is wrong, as `SchemaSet` would have `.xml` as an extension. But the content is a set of XML Schema. We should let the user choose if it's instance or schema.

pleacu:
Hey @igarashitm - so here's what I have for setting instance/ schema.
First - you click the import icon on the panel:

![sc1](https://user-images.githubusercontent.com/2981421/108772295-d2c2f380-752a-11eb-93c7-6d49e92b7566.png)

Then you get the instance/schema dialog:

![sc2](https://user-images.githubusercontent.com/2981421/108772347-e40c0000-752a-11eb-80d0-951d1331b5dc.png)

Notice in this example it defaults to schema since the file suffix is `xsd`.  It doesn't open the file and look for `SchemaSet` or `$schema` any more.  The import won't occur until the user clicks `OK`.  Does that look about right?  Any changes to text, etc?
thkx

igarashitm:
LGTM - I would avoid the word "instance-schema", we use the instance document to only "assume" the schema, but it might not be exactly correct.

PR #2575, opened 2021-02-10
fix: Persist XSD documents into xmlNameSpace within DataSource.

pleacu:
Fixes: #2563

### DIFF
--- a/ui/packages/atlasmap-core/src/utils/mapping-serializer.ts
+++ b/ui/packages/atlasmap-core/src/utils/mapping-serializer.ts
@@ -461,7 +461,7 @@ export class MappingSerializer {
       if (doc.locale != null) {
         serializedDoc['locale'] = doc.locale;
       }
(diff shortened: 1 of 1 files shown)


igarashitm (approved):
Approving this one, but wondering how it's decided to be DocumentType.XML or DocumentType.XSD as FHIR is XSD type for sure.

pleacu:
> Approving this one, but wondering how it's decided to be DocumentType.XML or DocumentType.XSD as FHIR is XSD type for sure.

It's based off of the imported document suffix - ref: https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap-core/src/services/document-management.service.ts#L441

igarashitm:
Ah that could be wrong, as SchemaSet would have `.xml` as an extension. But the content is a set of XML Schema. I think we should let user choose if it's instance or schema #832

Issue #2572, opened 2021-02-10
Missing namespace alias on XML collection field


igarashitm:
xref: https://issues.redhat.com/browse/ENTESB-15747

When following target field path is specified
```json
"path" : "/tns:Transaction/tns:Patient/tns:name<>/tns:given/@tns:value",
```

The output XML is missing `tns:` prefix on `name`
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<tns:Transaction xmlns:tns="http://hl7.org/fhir">
  <tns:Patient>
    <tns:id tns:value="ls1 ls2 ls3"/>
    <name>
      <tns:given tns:value="fn1"/>
      <tns:family tns:value="ln1"/>
    </name>
    <name>
      <tns:given tns:value="fn2"/>
      <tns:family tns:value="ln2"/>
    </name>
    <name>
      <tns:given tns:value="fn3"/>
      <tns:family tns:value="ln3"/>
    </name>
  </tns:Patient>
  <tns:Basic>
    <tns:id tns:value="1 2 3"/>
    <tns:language tns:value="t1 t2 t3"/>
  </tns:Basic>
</tns:Transaction>
```

Issue #2573, opened 2021-02-10
double slash (//) in field path causes INVALID_CHARACTER_ERR


igarashitm:
xref: https://issues.redhat.com/browse/ENTESB-15747

While UI should not add double slash in the path, we can normalize it at runtime side.

igarashitm:
Lowering to p1 as #2576 added a workaround to sanitize double slash.

PR #2576, opened 2021-02-10
fix: Missing namespace alias on XML collection field

igarashitm:
Fixes: #2572

Also adding a workaround for #2573, sanitize double slash to the single one


pleacu (approved):
+1

PR #2608, opened 2021-03-18
fix: Preview error on filter/select expression consuming complex coll…

### LINKED ISSUE #2607: Preview error when consuming complex collection
Although preview doesn't support collection field yet because there's no way to input multiple values (#501), it still should work as if it's non-collection (single value).
Using this schema https://github.com/atlasmap/atlasmap/blob/master/ui/test-resources/json/schema/JSONSchemaSource.json

Screenshot from 2021-03-17 13-15-40

Due to a bug in preview context

igarashitm:
…ection

It fixes #2607 at runtime side to handle collection instance for preview. We still need UI change mentioned in https://github.com/atlasmap/atlasmap/issues/2607#issuecomment-801468019 as well.


pleacu (approved):
+1

Issue #1735, opened 2020-02-01
Add `About` with version number in settings


igarashitm:
Add an easy way to see AtlasMap version in UI. It should be programmatically retrieved so that we don't have to update version string every time.

It would be better to have both of UI and runtime version. UI should retrieve runtime version via REST API.

PR #2611, opened 2021-03-19
fix: Add `About` with version number

igarashitm:
Fixes: #1735


pleacu (approved):
+1  Looks like we can make you a React UI developer!

Issue #2638, opened 2021-03-31
JSON/XML enum doesn't map at runtime


igarashitm:
JSON/XML enum mapping doesn't work at runtime.

Issue #2603, opened 2021-03-11
XML enumeration support - server.


pleacu:
Given the following XSD file:
<pre>&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot;?&gt;
&lt;xsd:schema xmlns:xsd=&quot;http://www.w3.org/2001/XMLSchema&quot; xmlns=&quot;http://hl7.org/fhir&quot; xmlns:xhtml=&quot;http://www.w3.org/1999/xhtml&quot; targetNamespace=&quot;http://hl7.org/fhir&quot; elementFormDefault=&quot;qualified&quot; version=&quot;1.0&quot;&gt;
  &lt;xsd:element name=&quot;AnEnumeration&quot;&gt;
  &lt;xsd:simpleType name=&quot;AnEnumeration&quot;&gt;
    &lt;xsd:restriction base=&quot;xsd:string&quot;&gt;
      &lt;xsd:enumeration value=&quot;G&quot;/&gt;
      &lt;xsd:enumeration value=&quot;M&quot;/&gt;
      &lt;xsd:enumeration value=&quot;T&quot;/&gt;
      &lt;xsd:enumeration value=&quot;c&quot;/&gt;
      &lt;xsd:enumeration value=&quot;d&quot;/&gt;
      &lt;xsd:enumeration value=&quot;k&quot;/&gt;
      &lt;xsd:enumeration value=&quot;m&quot;/&gt;
      &lt;xsd:enumeration value=&quot;g&quot;/&gt;
      &lt;xsd:enumeration value=&quot;n&quot;/&gt;
      &lt;xsd:enumeration value=&quot;none&quot;/&gt;
      &lt;xsd:enumeration value=&quot;p&quot;/&gt;
    &lt;/xsd:restriction&gt;
  &lt;/xsd:simpleType&gt;
  &lt;/xsd:element&gt;
&lt;/xsd:schema&gt;
</pre>

The XML inspection response from the backend server is:

<pre>body:
  executionTime: 4
  jsonType: &quot;io.atlasmap.xml.v2.XmlInspectionResponse&quot;
  xmlDocument:
    fields:
      field: Array(1)
        0:
          attribute: false
          jsonType: &quot;io.atlasmap.xml.v2.XmlField&quot;
          name: &quot;tns:AnEnumeration&quot;
          path: &quot;/tns:AnEnumeration&quot;
        __proto__: Object
          length: 1
      __proto__: Array(0)
[...]

PR #2639, opened 2021-03-31
fix: JSON/XML enum doesn't map at runtime

igarashitm:
Fixes: #2638

feat: Add enum support for XML schema inspection
Fixes: #2603


pleacu (approved):
+1

Issue #2648, opened 2021-04-05
BigDecimalConverter unnecessarily loses fractional part


igarashitm:
toDouble() Converting to BigInteger in the middle doesn't make any sense.

Issue #2647, opened 2021-04-04
Failed to convert field value '-0.96' into type 'DOUBLE'


Duvel:
I'm trying to map data from the OpenWeather API en created a mapping in the standalone application. Now when I try to process data retrieved from the API in Quarkus/Apache Camel, I get the mentioned error. The values are the negative values for the temperature in Celsius around here.

I'm using Quarkus 1.8.0 with Camel 3.9.0 which results in Atlasmap 2.1.6.

Duvel:
[...]

PR #2649, opened 2021-04-05
fix: Wrong use of float/double MIN/MAX

igarashitm:
Fixes: #2647

fix: `BigDecimalConverter#toDouble()` unnecessarily loses fractional part
Fixes: #2648


pleacu (approved):
+1

Issue #2655, opened 2021-04-07
XML Schema: Support form/elementFormDefault/attributeFormDefault


igarashitm:
xref: https://github.com/syndesisio/syndesis/pull/9228
xref: https://issues.redhat.com/projects/ENTESB/issues/ENTESB-14339

e.g. `form=unqualified` disables targetNamespace on a specific element.

igarashitm:
Changed to bug - currently schema inspection enforces targetNamespace to the subsequent elements/attributes which is against that XML Schema spec defines `elementFormDefault=unqualified` and `attributeFormDefault=unqualified` by default.

PR #2656, opened 2021-04-07
fix: XML Schema: Support form/elementFormDefault/attributeFormDefault

igarashitm:
Fixes: #2655

This fix changes the default behavior for XML Schema inspection. It used to ignore form/elementFormDefault/attributeFormDefault and always apply targetNamespace, while XML Schema defines `elementFormDefault=unqualified` and `attributeFormDefault=unqualified` by default. After this fix, subsequent children are no longer qualified with namespace unless `elementFormDefault=qualified`, `attributeFormDefault=qualified` or `form=qualified` is explicitly specified. Also note that the global element, i.e. defined at the top of the schema is always qualified.


pleacu (approved):
+1

PR #2678, opened 2021-04-21
feat: PoC for Java DSL runtime support

igarashitm:
Fixes: #2650
This introduces most basic support of AtlasMappingBuilder and AtlasField which allows user to write mapping logic in Java code. This changeset only enables simple read/write for non-collection field to bring the foundation in earlier. We can build a set of convenient API on top of this like an iterator for collection field, writing COMPLEX object in one invocation by comparing field name, apply transformations and etc.

### DIFF
--- a/lib/core/src/test/java/io/atlasmap/builder/DefaultAtlasMappingBuilderTest.java
+++ b/lib/core/src/test/java/io/atlasmap/builder/DefaultAtlasMappingBuilderTest.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (C) 2017 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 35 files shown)


pleacu:
I'll let @rkorytkowski take first look at this.

lib/core/src/test/java/io/atlasmap/builder/DefaultAtlasMappingBuilderTest.java
```diff
@@ -0,0 +1,57 @@
...
+        DefaultAtlasMappingBuilder builder = new DefaultAtlasMappingBuilder() {
+            @Override
+            public void processMapping() {
+                try {
+                    read(DEFAULT_SOURCE_DOCUMENT_ID, "/f1")
+                        .write(DEFAULT_TARGET_DOCUMENT_ID, "/f2");
+                    readConstant("c3").writeProperty(DEFAULT_TARGET_DOCUMENT_ID, "p4");
+                    readProperty(DEFAULT_SOURCE_DOCUMENT_ID, "p5").write(DEFAULT_TARGET_DOCUMENT_ID, "/f6");
+                } catch (AtlasException e) {
+                    addAudit(e);
```
rkorytkowski, line 42: It seems that we could make processMapping to rethrow this exception and handle it in DefaultAtlasMappingBuilder in 99% of cases...

igarashitm: Good catch, thanks! will do.

PR #2703, opened 2021-04-26
Reproducible iteration order in DefaultAtlasCompoundClassLoader #2702

### LINKED ISSUE #2702: Reproducible iteration order in DefaultAtlasCompoundClassLoader
HashSet makes no guarantees as to the iteration order of the set. This matters if the class sets loadable by the individual delegate class loaders overlap. DefaultAtlasCompoundClassLoader thus may return different results on different JVM implementations or versions.

A PR follows.

ppalaga:
Fix #2702

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasCompoundClassLoader.java
+++ b/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasCompoundClassLoader.java
@@ -18,7 +18,7 @@ package io.atlasmap.core;
 import java.net.URL;
 import java.util.Collections;
 import java.util.Enumeration;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


igarashitm:
Thanks for the PR, can you add `fix:` prefix to the commit message?

ppalaga:
> can you add `fix:` prefix to the commit message?

Thanks, done

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasCompoundClassLoader.java
```diff
@@ -19,6 +19,7 @@
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
```
igarashitm, line 21: It seems unused import checkstyle error is thrown for this one

ppalaga: Thanks, fixed d36f617

igarashitm (approved):
:+1:

PR #2705, opened 2021-04-26
feat: Make it possible to create a new DefaultAtlasContextFactory wit…

### LINKED ISSUE #2704: Make it possible to create a new DefaultAtlasContextFactory with a custom CompoundClassLoader
To set a custom CompoundClassLoader on a DefaultAtlasContextFactory, one currently has to do the following excercise:

```
        DefaultAtlasContextFactory cf = DefaultAtlasContextFactory.getInstance();
        cf.destroy();
        CompoundClassLoader compoundClassLoader = ...
        cf.init(compoundClassLoader);
```

I wonder if the API could be changed to allow something like the following or similar:

```
        CompoundClassLoader compoundClassLoader = ...
        DefaultAtlasContextFactory cf = DefaultAtlasContextFactory.getInstance(compoundClassLoader);
```

ppalaga:
…h a custom CompoundClassLoader 

Fix #2704

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContextFactory.java
+++ b/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContextFactory.java
@@ -75,9 +75,23 @@ public class DefaultAtlasContextFactory implements AtlasContextFactory, AtlasCon
     }
 
     public static DefaultAtlasContextFactory getInstance() {
+        return getInstance(true);
+    }
+
+    /**
+     * Returns the default singleton, possibly creating it if necessary.
+     *
+     * @param init if {@code true} the newly created {@link DefaultAtlasContextFactory} will be initialized upon
+     *            creation via {@link #init()}; otherwise, the newly created {@link DefaultAtlasContextFactory} will not
+     *            be initialized upon creation via {@link #init()}
+     * @return the singleton
+     */
+    public static DefaultAtlasContextFactory getInstance(boolean init) {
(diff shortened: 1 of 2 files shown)


lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContextFactory.java
```diff
@@ -82,6 +82,14 @@ public static DefaultAtlasContextFactory getInstance() {
         return factory;
     }
 
+    public static DefaultAtlasContextFactory getInstance(CompoundClassLoader cl) {
+        if (factory == null) {
+            factory = new DefaultAtlasContextFactory();
+            factory.init(cl);
```
igarashitm, line 88: I guess we'd want to do `init()` even if an instance already exists, following `destroy()`? since it's passing a class loader in. Or check if the class loader is same instance as existing?

igarashitm:
Can you also add unit test? would be in DefaultAtlasContextFactoryTest

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContextFactory.java
ppalaga: You are right, getting the life cycle right might be tricky. In 74d3ead I tried another option that would work for me. I have also added a test.

igarashitm (approved):
Thanks, fair enough. We might want to make `init()` separated from `getInstance()` in next major version.

Issue #2708, opened 2021-04-27
Backport CAMEL-16556 to the Camel2 camel-atlasmap component


igarashitm:
https://issues.apache.org/jira/browse/CAMEL-16556

Although Camel3 camel-atlasmap component is now maintained in Apache Camel upstream, we still have to maintain Camel2 version of it here, i.e. https://github.com/atlasmap/atlasmap/tree/master/camel

igarashitm:
https://github.com/apache/camel/pull/5472

PR #2712, opened 2021-04-27
fix: Backport CAMEL-16556 to the Camel2 camel-atlasmap component

igarashitm:
Fixes: #2708 
Moved propertiesFile parameter from endpoint to component


pleacu (approved):
+1

PR #2734, opened 2021-05-04
DataSourceMetadata: change isSource getter and setter names fixes #2722

### LINKED ISSUE #2722: DataSourceMetadata: isSource parameter name problematic while reflection
For `camel-quarkus `project, we need to register for reflection all the classes used to serialize/deserialize data on runtime.
To have better performance, we register methods or fields. Since the fields are not public, I registred only methods.
I had a problem with the `DataSourceMetadata` class, because of the name of property is `isSource` does not match with getter/setter names. 
So for this class we need to register both methods and fields, to include the `@JsonProperty("isSource")`.

zbendhiba:
Fixes: #2722

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
+++ b/lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
@@ -58,7 +58,7 @@ import io.atlasmap.v2.Json;
  * </ul>
  * </div>
  * {@link #load(Path)} {@link #export(OutputStream)}
- * 
+ *
  * <div>
  * This handler follows lazy loading strategy as much as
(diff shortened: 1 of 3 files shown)


igarashitm (approved):

igarashitm:
Oops forgot to ask - can you add `fix:` prefix to the commit message?
https://github.com/atlasmap/atlasmap/wiki/AtlasMap-Developer-Memo#examples

Also adding a second line `Fixes: #2722` would automatically close the issue when it's merged.

zbendhiba:
@igarashitm is the commit message  ok now ?

igarashitm:
:+1:

zbendhiba:
Thanks @igarashitm

Issue #2737, opened 2021-05-04
UI: Old jest is installed


igarashitm:
It seems the old version of jest is used despite of specifying newer in the package.json. It might need the yarn.lock to be regenerated.
> [INFO] @atlasmap/atlasmap: ts-jest[versions] (WARN) Version 24.9.0 of jest installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=26.0.0 <27.0.0-0). Please do not report issues in ts-jest if you are using unsupported versions.

igarashitm:
@pleacu I got this error after yarn.lock is regenerated - i.e. some dependencies might have been upgraded
> [INFO] @atlasmap/atlasmap: (typescript) Error: /home/tomo/workspace/iPaaS/atlasmap/atlasmap/ui/packages/atlasmap/src/Views/SourceMappingTargetView.tsx(71,15): semantic error TS2783: 'onMouseOver' is specified more than once, so this usage will be overwritten.
Actually the `{...mappingEvents}` contains `onMouseOver` and `onMouseOut` so it is duplicate. Which should I keep here?

pleacu:
Not sure - I'd remove the duplicates it's saying it's going to overwrite and see what happens...

gashcrumb:
I'd also make sure to align the versions in the package.json files under packages/* with the top-level package.json, install and then check with `yarn list` that the packages being pulled in are as expected.

igarashitm:
We have to stay with jest 24.9.0, see #2778

PR #2738, opened 2021-05-04
fix: regenerate yarn.lock for clean up

igarashitm:
Fixes: #2737

Opening a draft PR to get more eyes - I noticed that our yarn.lock file is in bad shape so removed and regenerated it by yarn install. But it triggered a bunch of TypeScript compile errors. It seems the newer version doesn't accept current code.


igarashitm:
@pleacu @gashcrumb @riccardo-forina do you know what this error means and how to fix?
> [INFO] @atlasmap/atlasmap: (typescript) Error: /home/tomo/workspace/iPaaS/atlasmap/atlasmap/ui/packages/atlasmap/src/UI/Canvas/Canvas.tsx(78,8): semantic error TS2322: Type '{ children: Element; onMouseDown?: ((event: MouseEvent<SVGSVGElement, MouseEvent>) => void) | ((event: MouseEvent<Element, MouseEvent>) => void) | undefined; ... 468 more ...; onTransitionEndCapture?: ((event: TransitionEvent<...>) => void) | undefined; }' is not assignable to type 'SVGProps<SVGSVGElement>'.
> [INFO] @atlasmap/atlasmap:   Types of property 'onTouchCancel' are incompatible.
> [INFO] @atlasmap/atlasmap:     Type '((event: TouchEvent<SVGSVGElement>) => void) | ((event: TouchEvent<Element>) => void) | undefined' is not assignable to type '((event: TouchEvent<SVGSVGElement>) => void) | undefined'.
> [INFO] @atlasmap/atlasmap:       Type '(event: TouchEvent<Element>) => void' is not assignable to type '(event: TouchEvent<SVGSVGElement>) => void'.
> [INFO] @atlasmap/atlasmap:         Types of parameters 'event' and 'event' are incompatible.
> [INFO] @atlasmap/atlasmap:           Type 'TouchEvent<SVGSVGElement>' is missing the following properties from type 'TouchEvent<Element>': detail, view

gashcrumb:
Is 16.14.0 the expected version of react?:

image

As that error strikes me as a version mismatch.

igarashitm:
ah OK thanks... it seems `react@>=16` specified in package.json ended up with react 17 from what I can see from yarn.lock. I'll try aligning those.

igarashitm:
Ah is it? help says "..., overwrite lockfile" so I expected to re-generate clean yarn.lock, but it doesn't seem to work as expected... I'll remove it.

igarashitm:
OK now it only shows react 16. But the property mismatch error around svg is still there...

igarashitm:
This message stinks, but no idea who is passing this incompatible `TouchEvent<Element>`
> Type '(event: TouchEvent<Element>) => void' is not assignable to type '(event: TouchEvent<SVGSVGElement>) => void'.

riccardo-forina:
I have to check this, but if I recall correctly something is augmenting the html types in a way that raises all kind of conflicts like the one you are seeing now. It's pretty typical of css-in-js libraries to create this kind of errors, and I think PatternFly in the version in use makes no exception.

Honestly, I'd not worry about the warnings and let the thing stay the way it was. I think the best way to tackle this is to first upgrade PatternFly (which is complicated enough on its own), and then try to update the toolchain

igarashitm:
I found this
```
[tomo@p1 ui]$ yarn list typescript
yarn list v1.22.10
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ tsdx@0.13.3
│  └─ typescript@3.9.9
└─ typescript@3.8.3
```
so I wonder, does the typescript 3.9 causes this? It could have introduced strict/deeper type check and causes those errors on existing code that has no problem with older version, i.e. 3.8.3. I'll look into what is bringing 3.9.9 in.

igarashitm:
hmm, `typescript@^3.7.3` specified by the tsdx resolves to 3.9.9, even if we specify `typescript@3.8.3` in devDependencies...

igarashitm:
hand edited yarn.lock to enforce typescript 3.8.3 but errors stay same. So it's not typescript version.

igarashitm:
OK, I gave up refreshing yarn.lock. It's too nasty. It'll be tracked with #2739

Issue #2744, opened 2021-05-05
UI: Imported CSV files disappear on reload.


pleacu:
ref: https://issues.redhat.com/browse/FUSEQE-12547

pleacu:
@igarashitm - Hey Tomo - having trouble with imports - I modified `file-management.service.ts` to add `DocumentType` to the imports.  It now looks like:

<pre>import {
  ErrorInfo,
  ErrorLevel,
  ErrorScope,
  ErrorType,
} from &apos;../models/error.model&apos;;

import { DocumentType, InspectionType } from &apos;../common/config.types&apos;;

import { ConfigModel } from &apos;../models/config.model&apos;;
import { DataMapperUtil } from &apos;../common/data-mapper-util&apos;;
import { DocumentManagementService } from &apos;./document-management.service&apos;;
import { Observable } from &apos;rxjs&apos;;
import { gzip } from &apos;pako&apos;;
import ky from &apos;ky&apos;;
import log from &apos;loglevel&apos;;
import { timeout } from &apos;rxjs/operators&apos;;
</pre>

But I'm getting 
<pre><font color="#34E2E2"><b>@atlasmap/core</b></font>: /home/pleacu/git-clone/atlasmap/ui/packages/atlasmap-core/src/services/file-management.service.ts
<font color="#34E2E2"><b>@atlasmap/core</b></font>:   23:1  error  Imports should be sorted alphabetically  sort-imports
</pre>

If I put the import in alphabetical order I get:
<pre><font color="#34E2E2"><b>@atlasmap/core</b></font>: /home/pleacu/git-clone/atlasmap/ui/packages/atlasmap-core/src/services/file-management.service.ts
<font color="#34E2E2"><b>@atlasmap/core</b></font>:   24:1  error  Expected &apos;multiple&apos; syntax before &apos;single&apos; syntax  sort-imports
</pre>

wdyt?

igarashitm:
multiple import should come first and then single, where each group (multiple, single) should be sorted alphabetically. Not across multiple and single.

pleacu:
Yup - needed to move `import { DocumentType, InspectionType } from '../common/config.types';` to the top - thkx!

PR #2745, opened 2021-05-05
fix: Make sure CSV docs are added to the exported meta data when esta…

pleacu:
…blishing an ADM.

Fixes: #2744

### DIFF
--- a/ui/packages/atlasmap-core/src/services/file-management.service.ts
+++ b/ui/packages/atlasmap-core/src/services/file-management.service.ts
@@ -293,13 +293,7 @@ export class FileManagementService {
         //   exportMeta - meta-data describing the instance or schema documents.
         //   exportBlockData - the actual source of the instance/schema/mappings documents or the Java class name.
         for (const doc of this.cfg.getAllDocs()) {
-          if (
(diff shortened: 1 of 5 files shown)


ui/packages/atlasmap-core/src/services/file-management.service.ts
```diff
@@ -296,7 +297,8 @@ export class FileManagementService {
             (doc.inspectionSource !== null &&
```
igarashitm, line 297: What is this if-clause checking? do we have any case we want to silently ignore on purpose? It looks like it's better to put the document metadata anyway here, or show an error if it finds any invalid document.

pleacu: Doc.Properties and DOC.Constants are skipped - we could flip the logic and just looks for those.

igarashitm: Ah OK makes sense, and yes would be better to explicitly exclude constants/props, which exactly describes what's going on.

igarashitm (approved):

Issue #2186, opened 2020-06-16
Upgrade PatternFly to latest v4


riccardo-forina:
https://medium.com/patternfly/patternflys-major-release-everything-you-need-to-know-1f9e50be81e8
https://github.com/patternfly/pf-codemods

This needs to be done in sync with Syndesis https://github.com/syndesisio/syndesis/issues/8707

Issue #2740, opened 2021-05-04
UI: Enforce import order


igarashitm:
~~See if this works for us - https://www.npmjs.com/package/eslint-plugin-import~~
eslint has built-in `sort-imports`

PR #2752, opened 2021-05-06
fix: Update to Patternfly 4

igarashitm:
Fixes: #2186
Fixes: #2231

TopologyView drastically changed in PF4 and we can't live with it anymore. Copied its layout from PF3 TopologyView into our MainLayout and removed TopologyView usage so that we can get full control on Toolbar. TopologyControlBar and TopologySideBar are still in use.

This also follows up #2740, removed unexpected ui/packages/atlasmap/.eslintrc.js which caused to disable eslint config in the package.json. Now eslint sort-imports is enabled in `@atlasmap/atlasmap` as well, so I had to fix import order errors.

### The current view with PF3 for reference
Screenshot from 2021-05-10 16-57-02

### After PF4 migration
Screenshot from 2021-05-11 13-47-15


riccardo-forina:
I see a lot of .d.ts files added by this. Those are generated by typescript, which shouldn't happen, the output should end into the dist folder. Something is wrong.

igarashitm:
They're generated through `typed-css-modules` to make them typed.
https://github.com/atlasmap/atlasmap/pull/2752/files#diff-039eae3f17c9d31a794c2acb271fdd90a0082486335e61fe86675f98a53af03fR17

igarashitm:
OK now layout looks slightly better, but still toolbar layout is broken - seems it's already functional though, import json schema, make mapping and export it works. e2e also passed.

igarashitm:
nah e2e still fails

igarashitm:
Short toolbar was caused by the duplicated `<Toolbar><ToolbarContent>`, now the last problem is the padding-top/padding-bottom of the toolbar. Posted a question to the patternfly forum.
https://forum.patternfly.org/t/pf3-pf4-migration-topologyview-and-toolbar/447

igarashitm:
Getting it back to draft... turned out storybook demo is broken. The storybook build using webpack also requires postcss setup that's different from what I did for regular build using tsdx...

igarashitm:
https://github.com/atlasmap/atlasmap/pull/2752/commits/88b33a4a46a962f2a03d613ea15f113fa9edb88f Added @storybook/addon-postcss but it doesn't work as expected. Our CSS classes all disappear - postcss `extract` option doesn't write a file, `inject` doesn't embed them. It does say PostCSS is working as a part of storybook build though. Note that it works just fine through tsdx+rollup-plugin-postcss.

igarashitm:
@pleacu storybook demo looks OK now

pleacu (approved):
+1
Looks like the at-sign expression field node inclusion dropdown menu takes the whole display - something to catch post commit

Issue #2750, opened 2021-05-06
UI: Source collection many-to-one preview mode generates extraneous concatenates.


pleacu:
This functions correctly but the preview request is incorrect:

![sc1](https://user-images.githubusercontent.com/2981421/117313461-ead0c080-ae53-11eb-924b-12fcaa5312e4.png)

<pre><font color="#1E8C32">2021-05-06 09:07:47.674 </font><font color="#6AC911">DEBUG</font> <font color="#75507B">2272</font> --- [  XNIO-1 task-1] <font color="#06989A">io.atlasmap.service.AtlasService        </font> : Preview request: {
  &quot;Mapping&quot; : {
    &quot;jsonType&quot; : &quot;io.atlasmap.v2.Mapping&quot;,
    &quot;inputFieldGroup&quot; : {
      &quot;jsonType&quot; : &quot;io.atlasmap.v2.FieldGroup&quot;,
      &quot;actions&quot; : [ {
        &quot;delimiter&quot; : &quot; &quot;,
        &quot;delimitingEmptyValues&quot; : true,
        &quot;@type&quot; : &quot;Concatenate&quot;
      } ],
      &quot;field&quot; : [ {
        &quot;jsonType&quot; : &quot;io.atlasmap.v2.FieldGroup&quot;,
        &quot;actions&quot; : [ {
          &quot;delimiter&quot; : &quot; &quot;,
          &quot;delimitingEmptyValues&quot; : true,
          &quot;@type&quot; : &quot;Concatenate&quot;
        } ],
        &quot;docId&quot; : &quot;JSONSchemaSource-5c0ba2d9-4453-4fa9-a6f1-783817a6893b&quot;,
        &quot;path&quot; : &quot;/addressList&lt;&gt;/street&quot;,
        &quot;fieldType&quot; : &quot;COMPLEX&quot;,
        &quot;field&quot; : [ {
          &quot;jsonType&quot; : &quot;io.atlasmap.json.v2.JsonField&quot;,
          &quot;value&quot; : &quot;x&quot;,
          &quot;docId&quot; : &quot;JSONSchemaSource-5c0ba2d9-4453-4fa9-a6f1-783817a6893b&quot;,
          &quot;index&quot; : 0,
          &quot;path&quot; : &quot;/addressList&lt;0&gt;/street&quot;,
          &quot;fieldType&quot; : &quot;STRING&quot;,
          &quot;name&quot; : &quot;street&quot;
        } ]
      }, {
[...]

PR #2755, opened 2021-05-07
fix: Correct extraneous concats in many-to-one preview mode.

pleacu:
Fixes: #2750

Also trigger mapping update on clearing values on preview disable.


pleacu:
Yes - the `test('collection many-to-one deserialize/serialize', (done)` should cover it but I found a typo - one second...

pleacu:
Corrected sample mappings typo - generated mapping info had to change

igarashitm:
Can we get those json file pretty printed to be human readable? I think there're a couple of ways to compare JSON in structure, not necessarily strictly equal as a string literal.
https://stackoverflow.com/questions/26049303/how-to-compare-two-json-have-the-same-properties-without-order

pleacu:
Good idea - I was wondering if there was a better way of diffing those JSON files.  Can we commit this first?  That way the new compare methodology can be a separate thing.

igarashitm (approved):

Issue #2746, opened 2021-05-06
UI: Malformed CSV processing parameters dialog


pleacu:
This dialog needs some work:

![sc1](https://user-images.githubusercontent.com/2981421/117297111-d1c01380-ae43-11eb-91c7-ef00bae61fe1.png)

pleacu:
@igarashitm - just looking for an opinion on this updated CSV dialog.  If ok then I'll issue a PR

![sca](https://user-images.githubusercontent.com/2981421/117705935-ae29ff80-b19a-11eb-8255-bcbe959680f3.png)

PR #2763, opened 2021-05-10
fix: Correct CSV dialog.

pleacu:
Fixes: #2746


pleacu:
> Can you add story and unit test?

I don't think so - I didn't add anything structural - it's all visual.  I could create a new issue to walk through all of the CSV params and verify them.

igarashitm:
There already is the story here, unit test could be added - https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.stories.tsx

pleacu:
@igarashitm - pls re-review when you get a chance :)

igarashitm:
The story looks good, how about the unit test?

pleacu:
> The story looks good, how about the unit test?

Since there's no spec.ts yet for any dialog I was going to add that as a separate issue.

Issue #2739, opened 2021-05-04
Get UI dependencies up to date


igarashitm:
Get UI dependencies up to date as much as possible. Then remove yarn.lock and regenerate it by yarn install. I tried it in #2738 but gave up as it gets too messy. It would be better to start with latest React and Patternfly.

PR #2786, opened 2021-05-19
chore: Update UI dependencies

igarashitm:
Fixes: #2739

Wrapping up this round with updating some small libraries and cleaning up dependency related sections in each package.json
ky couldn't be updated as it caused errors with Jest, even with ky 0.25
Unexpectedly webpack 4.46.0 was still in the yarn.lock, removed manually.


pleacu (approved):
+1

Issue #2756, opened 2021-05-07
UI: refactor JSON content comparision in unit tests.


pleacu:
Ref:
https://stackoverflow.com/questions/26049303/how-to-compare-two-json-have-the-same-properties-without-order

PR #2799, opened 2021-05-19
fix: Add deep-dive JSON object diff to spec tests in UI

pleacu:
Fixes: #2756

This is good for now - need to address enums and csvs in depth next.


igarashitm (approved):
human readable :tada:

Issue #2751, opened 2021-05-06
Enforce license header


igarashitm:
Both of backend and UI. Check if this one works for us https://www.npmjs.com/package/eslint-plugin-header

PR #2824, opened 2021-05-19
chore: Enabled license-maven-plugin

igarashitm:
Automatically add license header as a part of build process
Only maven/backend part of #2751. UI work will follow up later.


pleacu (approved):
+1

Issue #2821, opened 2021-05-19
UI: deleting a field node from a conditional expr removes it from expr box but not from the mapping.


pleacu:
Mapping appears as follows - note 'city' exists:

<pre>     &quot;mapping&quot; : [ {
        &quot;jsonType&quot; : &quot;io.atlasmap.v2.Mapping&quot;,
        &quot;expression&quot; : &quot;select( filter(${JSONSchemaSource-797ec3c2-3357-4bc2-8921-bc3f6d6cc575:/addressList&lt;&gt;}, ${/street}), ${/state})&quot;,
        &quot;inputFieldGroup&quot; : {
          &quot;jsonType&quot; : &quot;io.atlasmap.v2.FieldGroup&quot;,
          &quot;actions&quot; : [ ],
          &quot;field&quot; : [ {
            &quot;jsonType&quot; : &quot;io.atlasmap.v2.FieldGroup&quot;,
            &quot;actions&quot; : [ ],
            &quot;docId&quot; : &quot;JSONSchemaSource-797ec3c2-3357-4bc2-8921-bc3f6d6cc575&quot;,
            &quot;path&quot; : &quot;/addressList&lt;&gt;&quot;,
            &quot;fieldType&quot; : &quot;COMPLEX&quot;,
            &quot;field&quot; : [ {
              &quot;jsonType&quot; : &quot;io.atlasmap.json.v2.JsonField&quot;,
              &quot;docId&quot; : &quot;JSONSchemaSource-797ec3c2-3357-4bc2-8921-bc3f6d6cc575&quot;,
              &quot;index&quot; : 0,
              &quot;path&quot; : &quot;/addressList&lt;&gt;/city&quot;,
              &quot;fieldType&quot; : &quot;STRING&quot;,
              &quot;name&quot; : &quot;city&quot;
            }, {
              &quot;jsonType&quot; : &quot;io.atlasmap.json.v2.JsonField&quot;,
              &quot;docId&quot; : &quot;JSONSchemaSource-797ec3c2-3357-4bc2-8921-bc3f6d6cc575&quot;,
              &quot;index&quot; : 1,
              &quot;path&quot; : &quot;/addressList&lt;&gt;/state&quot;,
              &quot;fieldType&quot; : &quot;STRING&quot;,
              &quot;name&quot; : &quot;state&quot;
            }, {
              &quot;jsonType&quot; : &quot;io.atlasmap.json.v2.JsonField&quot;,
[...]

PR #2828, opened 2021-05-20
fix: Correct removal of field nodes/ mapped fields in conditional ex…

pleacu:
…pressions.

Fixes: #2821

The removeField() function will remove the exact mapped field associated with the specified field from the mapped fields array.


igarashitm:
Can we have unit test for the `removeField()` and `removeMappedField()` to show the usage?

igarashitm:
ah rather unit test for `ExpressionModel.removeToken()` that verifies the fix?

igarashitm (approved):

Issue #1127, opened 2019-07-19
Support java 11


apupier:
it would be nice to support running with Java 11.

use cases:
- VS Code extension relies on the Java version available on the host. it will become more and more frequent to be java 11
- Inside Eclipse Che, we will be able to choose Java 11 container image which should (in theory) be faster and more maintained.

apupier:
I think this is something relatively important

apupier:
still relevant

apupier:
still relevant, even more with Integration.Next which will be only on java 11

PR #2829, opened 2021-05-20
fix: Support building/running with java11

igarashitm:
Fixes: #1127

CircleCI will keep using java8 container and added a GitHub Action building with java11, both should run as PR checks.


pleacu (approved):
+1

Issue #2855, opened 2021-05-25
Enforce indent (2 spaces) in UI code


igarashitm:
https://eslint.org/docs/rules/indent

PR #2856, opened 2021-05-25
chore: Enforce indent and single quote

igarashitm:
Fixes: #2855
It seems 2 spaces indent is enabled by default, but added to prettier config explicitly. Also aligned prettier config between 3 sub packages.
Gave up doing indentation for Java with checkstyle, as it conflicts with Eclipse formatter


pleacu (approved):
+1

Issue #2881, opened 2021-06-01
UI: Split expression-handler.service out of mapping-management.service


igarashitm:
Or expression-management.service, etc to reduce the size of mapping-management.service. Possibly enum-handler.service as well.

igarashitm:
Or expression-util even?

Issue #2726, opened 2021-04-30
Put mapping digest file handling routine together


igarashitm:
Currently it's splitted into FileManagementService (export) and DocumentManagementService (generateExportMappings, generateExportBlockData, generateExportMetaStr), both writing raw json of mapping digest file. It would be better to put together in FileManagementService so the digest file format is defined in one place. sanitizeJSON could be moved to util then.

PR #2882, opened 2021-06-02
fix: UI: Split expression-util out of mapping-management.service

igarashitm:
Fixes: #2881
fix: UI: Move document-management.service#setLibraryToService() to file-management.service
Fixes: #2880
Also partially addresses #2726, isolated digest file handling into mapping-digest-util


pleacu (approved):
+1  Do we need the 'defined but never used' warnings? - especially for Atlasmap.tsx and AtlasmapProvider...

igarashitm:
It'd be better to remove those unused variables.

igarashitm:
Found that something is going wrong - https://github.com/atlasmap/atlasmap/issues/2886

Issue #2893, opened 2021-06-02
Remove GET version of inspection REST endpoint


igarashitm:
UI uses POST version and GET version has been abandoned for a while.

Issue #2892, opened 2021-06-02
Remove mavenclasspath REST endpoint and its usage from UI


igarashitm:
It was used for development before we implement custom class loading. No longer needed.

PR #2894, opened 2021-06-02
fix: Remove mavenclasspath REST endpoint and its usage from UI

igarashitm:
Fixes: #2892
fix: Remove GET version of inspection REST endpoint
Fixes: #2893


pleacu (approved):
+1

Issue #2917, opened 2021-06-08
UI: Encapsulate document specific inspection handling into each model classes


igarashitm:
Reduce `if (docDef.type === DocumentType.JAVA` kind of code by encapsulating them into each document specific model classes.

Issue #2915, opened 2021-06-07
UI: Don't use observable for a single event


igarashitm:
Use `Promise` instead for a single event. `Observable` is to observe continuous event stream.

Issue #2900, opened 2021-06-05
UI: serializer adds `scope` property to all fields


igarashitm:
`scope` property is allowed only for PropertyField.

Issue #2899, opened 2021-06-04
UI: Split file update task from user ADM export


igarashitm:
These are not really exporting ADM archive file for user download, but just updating digest file, mapping and schema. It has to be splitted from user export. Then `exportADMArchive()` might also invoke that.
https://github.com/atlasmap/atlasmap/blob/3bd6c328b58f0cb65089ec5680c7329f1bd1616e/ui/packages/atlasmap/src/Atlasmap/utils/document.ts#L45
https://github.com/atlasmap/atlasmap/blob/3bd6c328b58f0cb65089ec5680c7329f1bd1616e/ui/packages/atlasmap/src/Atlasmap/utils/document.ts#L136
https://github.com/atlasmap/atlasmap/blob/3bd6c328b58f0cb65089ec5680c7329f1bd1616e/ui/packages/atlasmap/src/Atlasmap/utils/document.ts#L136

PR #2921, opened 2021-06-09
fix: UI: Refactoring @atlasmap/core services

igarashitm:
fix: UI: Let initialization-service focus on initialization
Fixes: #2879
Relocated document/mapping/file handling from initialization service to the other services.

fix: Fixed wording of catalog/digest/ADM thing
Fixes: #2724

fix: Relocated digest file handling into file-management service(#2726). Also added a contract of ADMDigest file under contracts/ (#2878).
Fixes: #2726

fix: Split file update task from user ADM export
Fixes: #2899

fix: UI: serializer adds scope property to all fields
Fixes: #2900

fix: Pushed some of the logic back from @atlasmap/atlasmap utils to @atlasmap/core (#2883)

fix: UI: Don't use observable for a single event
Fixes: #2915

fix: Encapsulate document specific inspection handling into each model classes
Fixes: #2917


ui/packages/atlasmap-core/src/models/config.model.ts
```diff
@@ -204,9 +199,12 @@ export class ConfigModel {
             '&collectionClassName=' + docInitModel.collectionClassName;
         }
       }
+    } else {
+      docDef.uri = 'atlas:' + docDef.type.toLowerCase() + ':' + docDef.id;
     }
 
-    if (docInitModel.isSource) {
+    // somehow docInitModel.isSource gets always truthy... so explicitly compare with true
+    if (docInitModel.isSource === true) {
```
pleacu, line 207: That's strange - so if `docInitModel.isSource` is false it's still true?

igarashitm: yep, not sure what caused it, as if the type (boolean) is lost at this point and recognized as string

ui/packages/atlasmap-core/src/models/inspect/xml-inspection.model.ts
```diff
@@ -0,0 +1,187 @@
+/*
```
pleacu, line 1: I like the models/inspect/... architecture - much easier to find.

ui/packages/atlasmap-core/src/services/field-action.service.ts
```diff
@@ -49,13 +48,10 @@ export class FieldActionService {
...
   async fetchFieldActions(): Promise<boolean> {
-    return new Promise<boolean>((resolve, reject) => {
+    return new Promise<boolean>((resolve) => {
       if (this.cfg.preloadedFieldActionMetadata) {
         this.clearActionDefinitions();
-        if (
-          this.cfg.preloadedFieldActionMetadata &&
-          this.cfg.preloadedFieldActionMetadata.ActionDetails
-        ) {
+        if (this.cfg.preloadedFieldActionMetadata.ActionDetails) {
```
pleacu, line 54: Does this want to be 'if (this.cfg.preloadedFieldActionMetadata?.ActionDetails) {'

igarashitm: It is already checked at L52 I think?

pleacu: yup - missed that

ui/packages/atlasmap-core/src/models/config.model.ts
igarashitm: @gashcrumb @riccardo-forina do you know anything about this one? `docInitModel.isSource` is a typed boolean property and it is working fine on master, but after this refactoring `if(docInitModel.isSource)` evaluates to truthy even if it's false.

pleacu (approved):
This looks good to me - I'd say push it asap and we can deal with any issues.

igarashitm:
@pleacu thanks! I'll debug a bit more and make sure at least import/export path works.

ui/packages/atlasmap-core/src/models/config.model.ts
igarashitm: nevermind, turned out it's caused by the type cast from JSON.parse() outcome into the contract interface. If the value in the JSON is string (e.g. {"isSource": "true"}) then a string value is delivered through a boolean property. It had to be cast to `string` anyway. crappy type system.

Issue #2925, opened 2021-06-10
E2ETest failure


igarashitm:
> no such element: Unable to locate element: {"method":"xpath","selector":"//article[@aria-label='io.atlasmap.service.my.MyFieldActionsModel']"}
  (Session info: headless chrome=91.0.4472.77)
For documentation on this error, please visit: https://www.seleniumhq.org/exceptions/no_such_element.html
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:17:03'
System info: host: 'p1', ip: '192.168.12.121', os.name: 'Linux', os.arch: 'amd64', os.version: '5.12.9-200.fc33.x86_64', java.version: '11.0.11'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Capabilities {acceptInsecureCerts: false, browserName: chrome, browserVersion: 91.0.4472.77, chrome: {chromedriverVersion: 90.0.4430.212 (e3cd97fc771b..., userDataDir: /tmp/.org.chromium.Chromium...}, goog:chromeOptions: {debuggerAddress: localhost:38975}, javascriptEnabled: true, networkConnectionEnabled: false, pageLoadStrategy: normal, platform: LINUX, platformName: LINUX, proxy: Proxy(), setWindowRect: true, strictFileInteractability: false, timeouts: {implicit: 0, pageLoad: 300000, script: 30000}, unhandledPromptBehavior: dismiss and notify, webauthn:extension:largeBlob: true, webauthn:virtualAuthenticators: true}
Session ID: 1086ae1f00babe06eee6fb6e24b8b785
*** Element info: {Using=xpath, value=//article[@aria-label='io.atlasmap.service.my.MyFieldActionsModel']}
> ...
> at io.atlasmap.standalone.E2ETest.test(E2ETest.java:133)

igarashitm:
It'd be a timing issue, interestingly it tends to succeed in GH action, but fail on CircleCI and my local build.

igarashitm:
Not really, it used to be using FQDN as a Java Document label but now it's a simple name. Using simple name is correct as a default display name. cf. #2919

Issue #2928, opened 2021-06-10
Add data-testid on the Document card with Document ID


igarashitm:
Currently E2ETest looks for an `aria-label` that is using Document name, but Document name is not unique and could be changed by the user. It would help to add more e2e test if `data-testid` could identify the Document.

Issue #2919, opened 2021-06-08
UI: Assign GUID to Java Document ID as well, and use simple class name for default Document name


igarashitm:
Currently class name is used as a Document ID, so you can't import multiple Java Document from a single class
https://github.com/atlasmap/atlasmap/blob/3bd6c328b58f0cb65089ec5680c7329f1bd1616e/ui/packages/atlasmap-core/src/services/initialization.service.ts#L591
It'd be better to assign GUID just like others. Then default Document name for Java could be just a simple name (without package) so that it appears nicer as a Document display name in the canvas.

PR #2929, opened 2021-06-10
fix: E2ETest failure

igarashitm:
Fixes: #2925
It turned out the FQDN was used as a Document label for Java Document, and somehow changed to simple name after refactoring. Using simple name as a Document display name is right (#2919), but it might be better to use Document ID for data-testid to be unique (#2928).


pleacu (approved):
+1

Issue #2721, opened 2021-04-29
Add jandex index for Atlasmap model beans discovery


zbendhiba:
Add jandex index for Atlasmap model beans discovery. This is needed to discover all classes and enums needed  to serialize/deserialize data that needs reflection.

The idea is to add the jandex index for those modules, using [jandex-maven-plugin](https://github.com/wildfly/jandex-maven-plugin) :

- atlas-model
- atlas-csv-model
- atlas-dfdl-model
- atlas-java-model
- atlas-json-model
- atlas-xml-model

zbendhiba:
I'm working on this one !

zbendhiba:
@igarashitm  I've updated the issue to set the design. Let me know if it's okay

I did a first version of jandex, for the modules we want to have all classes indexed (atlas-model, atlas-csv-model ....) : see here https://github.com/zbendhiba/atlasmap/commit/978c952f60099bd26be63b4dc9979c05e3f2de10

When I build those modules, the jandex index is generated successfully in target/classes/META-INF/jandex.idx

But when the Jar is generated, this index is not included at all. I tried to understand the maven configuration of the project.

I need help on this one

(FYI, If I do the same in a very simple java project, the jandex index is presented in the generated JAR)

zbendhiba:
putting this here if someone is facing the same problem.
maven-bundle-plugin substitutes maven-jar-plugin when the packaging is bundle

We need to add the IncludeResource in maven-bundle-plugin, so that the jandex.idx is included in the jar.

`
<Include-Resource>META-INF/jandex.idx=target/classes/META-INF/jandex.idx</Include-Resource>
`

PR #2966, opened 2021-06-15
fix: Add jandex index for atlasmap model beans discovery

zbendhiba:
Fixes: #2721

FYI, if I put this configuration in `parent/pom.xml`:  
`<Include-Resource>${osgi.include.resource},META-INF/jandex.idx=target/classes/META-INF/jandex.idx</Include-Resource>`

The build doesn't work for modules that don't have the jandex.idx. This is the reason why I've overwrote this maven plugin section everywhere, I've created the jandex.
If you have an idea of how I can improve this, let me know.

### DIFF
--- a/lib/model/pom.xml
+++ b/lib/model/pom.xml
@@ -30,6 +30,10 @@
 
   <properties>
     <osgi.export.pkg>io.atlasmap.v2</osgi.export.pkg>
+    <!-- override include resource to include the jandex resource -->
+    <osgi.include.resource.atlasmap>${osgi.include.resource},META-INF/jandex.idx=target/classes/META-INF/jandex.idx</osgi.include.resource.atlasmap>
+    <!-- override skipping jandex maven plugin -->
+    <jandex.skip>false</jandex.skip>
   </properties>
 
   <dependencies>
--- a/lib/modules/csv/model/pom.xml
+++ b/lib/modules/csv/model/pom.xml
@@ -30,6 +30,10 @@
 
   <properties>
     <osgi.export.pkg>io.atlasmap.csv.v2</osgi.export.pkg>
+    <!-- override include resource to include the jandex resource -->
+    <osgi.include.resource.atlasmap>${osgi.include.resource},META-INF/jandex.idx=target/classes/META-INF/jandex.idx</osgi.include.resource.atlasmap>
+    <!-- override skipping jandex maven plugin -->
+    <jandex.skip>false</jandex.skip>
   </properties>
 
   <dependencies>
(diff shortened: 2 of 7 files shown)


lib/model/pom.xml
```diff
@@ -56,4 +56,46 @@
...
+            <Bundle-SymbolicName>${osgi.symbolic.name}</Bundle-SymbolicName>
+            <Bundle-DocURL>${project.url}</Bundle-DocURL>
+            <Export-Package>${osgi.export.pkg}</Export-Package>
+            <Import-Package>${osgi.import.pkg}</Import-Package>
+            <Export-Service>${osgi.export.service}</Export-Service>
+            <Implementation-Title>${project.description}</Implementation-Title>
+            <Implementation-Version>${project.version}</Implementation-Version>
+            <Embed-Dependency>${osgi.embed.dependency}</Embed-Dependency>
+            <Embed-Transitive>${osgi.embed.transitive}</Embed-Transitive>
+            <Include-Resource>${osgi.include.resource},META-INF/jandex.idx=target/classes/META-INF/jandex.idx
```
igarashitm, line 81: Couldn't it be like `<osgi.include.resource>${osgi.include.resource},META-INF/jandex.idx=target/classes/META-INF/jandex.idx</osgi.include.resource>` under properties?

igarashitm: Or, in parent/pom.xml
```xml
<properties>
  <osgi.include.resource.atlasmap>
    ${osgi.include.resource}
  </osgi.include.resource.atlasmap>
</properties>
...
<Include-Resource>${osgi.include.resource.atlasmap}</Include-Resource>
```
then in children which needs to override
```xml
<properties>
  <osgi.include.resource.atlasmap>
    ${osgi.include.resource},META-INF/jandex.idx=target/classes/META-INF/jandex.idx
  </osgi.include.resource.atlasmap>
</properties>

Comment (zbendhiba) [reply lib/model/pom.xml:81]: @igarashitm this sounds a good idea.

File: lib/modules/csv/model/pom.xml [hunk 2]
```diff
@@ -56,4 +56,47 @@
...
+            <Implementation-Title>${project.description}</Implementation-Title>
+            <Implementation-Version>${project.version}</Implementation-Version>
+            <Embed-Dependency>${osgi.embed.dependency}</Embed-Dependency>
+            <Embed-Transitive>${osgi.embed.transitive}</Embed-Transitive>
+            <Include-Resource>${osgi.include.resource},META-INF/jandex.idx=target/classes/META-INF/jandex.idx
+            </Include-Resource>
+          </instructions>
+        </configuration>
+      </plugin>
+      <plugin>
```
igarashitm, line 86: This could be moved to `PluginManagement` in parent/pom.xml with having `skip` to be true by default, maybe with a property like `jandex.skip`. Then only in children that needs jandex, set the `jandex.skip` property to true.

zbendhiba:
@igarashitm I took in consideration your remarks. I prefer the new version.
Thanks

igarashitm (approved):
:+1:

Issue #2694, opened 2021-04-23
Improve UI test coverage


igarashitm:
Add more UI tests. At least atlasmap-core has to be well tested.

One parameter is the coverage, but not always helpful. Covering known usecases comes first.
https://6984-106613119-gh.circle-artifacts.com/0/workspace/coverage/ui/atlasmap-core/index.html
https://6984-106613119-gh.circle-artifacts.com/0/workspace/coverage/ui/atlasmap/index.html
https://6984-106613119-gh.circle-artifacts.com/0/workspace/coverage/ui/atlasmap-standalone/index.html

igarashitm:
Note that storybook is counted as tests that earn coverage. For visual components it's better to add stories instead of tests.
https://docs.atlasmap.io/developer-guide/coverage/ui/atlasmap/src/UI/ColumnMapper/index.html

igarashitm:
Refactorings like #2883 also contributes to reduce untested code in @atlasmap/atlasmap and instead test it in @atlasmap/core which doesn't have any React component restriction.

PR #2967, opened 2021-06-15
chore: Add UT for SourceTargetView (#2694)

igarashitm:
Also limited preview input box to active mapping mapped field


pleacu (approved):
+1

Issue #2878, opened 2021-06-01
UI: Establish contract between backend


igarashitm:
Currently UI receives all the objects sent/received to/from backend as `any`. We should instead have model objects that represent typed JSON payload, which correspond to Java model objects at backend.

BTW we do have `models/` subdir in UI atlasmap-core src, but they are basically implementations. I suggest to keep models as is and introduce contract interfaces under `contracts/`.
https://github.com/atlasmap/atlasmap/tree/master/ui/packages/atlasmap-core/src/contracts

PR #3021, opened 2021-06-30
fix: UI: Establish contract between backend

igarashitm:
Fixes: #2878


pleacu (approved):
+1
Nice cleanup of `any` types and a much better class inspection contract.  A lot of changes on the serializer/deserializer - looks good but we'll have to beat on it.

Issue #2977, opened 2021-06-18
Mapping preview fails


pleacu:
Mapping preview is generating this error:
<pre>      &quot;audit&quot; : [ {
        &quot;message&quot; : &quot;Expression processing error [if ( ${DOC.Properties.704068:/JSONSchemaSource/prop-city} == ${DOC.Constants.801708:/Boston}, ${JSONInstanceSource:/order/address/city}, ${JSONInstanceSource:/order/address/state})]: null&quot;,
        &quot;docId&quot; : &quot;if ( ${DOC.Properties.704068:/JSONSchemaSource/prop-city} == ${DOC.Constants.801708:/Boston}, ${JSONInstanceSource:/order/address/city}, ${JSONInstanceSource:/order/address/state})&quot;,
        &quot;docName&quot; : &quot;Preview&quot;,
        &quot;status&quot; : &quot;ERROR&quot;
      } ]
</pre>

ref:
![err1](https://user-images.githubusercontent.com/2981421/122560639-d402b980-d00e-11eb-8855-c0e177480e7c.png)

mapping:
<pre>    &quot;mapping&quot; : {
      &quot;jsonType&quot; : &quot;io.atlasmap.v2.Mapping&quot;,
      &quot;expression&quot; : &quot;if ( ${DOC.Properties.704068:/JSONSchemaSource/prop-city} == ${DOC.Constants.801708:/Boston}, ${JSONInstanceSource:/order/address/city}, ${JSONInstanceSource:/order/address/state})&quot;,
      &quot;inputFieldGroup&quot; : {
        &quot;jsonType&quot; : &quot;io.atlasmap.v2.FieldGroup&quot;,
        &quot;actions&quot; : [ ],
        &quot;field&quot; : [ {
          &quot;jsonType&quot; : &quot;io.atlasmap.v2.PropertyField&quot;,
          &quot;value&quot; : &quot;&quot;,
          &quot;docId&quot; : &quot;DOC.Properties.704068&quot;,
          &quot;path&quot; : &quot;/prop-city&quot;,
          &quot;fieldType&quot; : &quot;STRING&quot;,
          &quot;name&quot; : &quot;prop-city&quot;,
          &quot;scope&quot; : &quot;JSONSchemaSource&quot;
        }, {
          &quot;jsonType&quot; : &quot;io.atlasmap.v2.ConstantField&quot;,
          &quot;value&quot; : &quot;Boston&quot;,
[...]

PR #3031, opened 2021-07-01
fix: scope is missing in property field path

igarashitm:
Fixes: #2977
Expression mapping needs corresponding field entry in `inputFieldGroup` or `inputField` where it's identified with Document ID and path. Same for property, path has to match. No need to handle property scope separately as it's supposed to be put in path.
atlasmap


pleacu (approved):
+1

Issue #2990, opened 2021-06-21
Clicking `Cancel` in `ParametersDialog` looses previously choosen parameters


zregvart:
This is from the integrated dialog in Syndesis:

![Peek 2021-06-21 15-36](https://user-images.githubusercontent.com/1306050/122771235-886a3d00-d2a6-11eb-89f8-e7b2789d2ffb.gif)

Seems like the call to `reset` could be removed here:

https://github.com/atlasmap/atlasmap/blob/d645eacf777139797715d2e28b3f4121cf04ef12/ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.tsx#L86

rkorytkowski:
In Atlasmap UI clicking "Cancel" simply cancels importing of a CSV sample, thus the reset is used. Something to have in mind when working on that dialog so it still works for Atlasmap standalone...

igarashitm:
Ah interesting, it's the first usecase to use document import related feature in Syndesis, as in Syndesis documents are loaded from backdoor when AtlasMap initializes. It might need some advanced exposed control to support it.

rkorytkowski:
It might be as simple as adding onReset handle with a default as is now (reset to a specific CSV format) and a custom one to reset to a previous state.

pleacu:
sorry - grabbed by mistake

pleacu:
Ref: Syndesis usage: https://mail.google.com/mail/u/0/?tab=wm#inbox/FMfcgzGkZGdGbJNvlFmlrgltHcRXHksz?projector=1

pleacu:
I've created an edit CSV parameters dialog (here's the screen capture):

![ok1](https://user-images.githubusercontent.com/2981421/124661563-a8d4f280-de75-11eb-8fd3-0441ac39bd7e.png)

(still playing with my gnome recorder).  You can edit or cancel as often as you like.  Ok on restart and with both source and target panels.  I'll issue a PR for this and we'll see if you need further glue with the Syndesis/ atlasmap app interface.

Issue #2975, opened 2021-06-17
UI: Migrate data model contracts from @atlasmap/atlasmap to @atlasmap/core


igarashitm:
https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap/src/Views/models.ts
These are data model contracts between @atlasmap/atlasmap and @atlasmap/core and better resides in @atlasmap/core. Same direction with #2883. Then existing model objects should implement these interfaces so that @atlasmap/core can pass them directly to the @atlasmap/atlasmap without converter.

I think for now it's OK to coexist in https://github.com/atlasmap/atlasmap/tree/master/ui/packages/atlasmap-core/src/contracts and then we can separate into `backend` and `presentation` subdir whenever we want in the future.

Issue #2883, opened 2021-06-02
UI: Minimize logic in @atlasmap/atlasmap and keep them in @atlasmap/core


igarashitm:
There're even duplicate code, e.g.
https://github.com/atlasmap/atlasmap/blob/b46d59218952e7e126bff881a2a4ae0bb3172e39/ui/packages/atlasmap-core/src/models/expression.model.ts#L27-L203
and
https://github.com/atlasmap/atlasmap/blob/b46d59218952e7e126bff881a2a4ae0bb3172e39/ui/packages/atlasmap/src/Atlasmap/utils/expression.ts#L30-L93
They already look different, maybe only one of them is in use? Ideally minimize logic in @atlasmap/atlasmap to those have to stay in presentation layer, and move everything else into @atlasmap/core and connect them through simple public method invocation, i.e. clean up contract between @atlasmap/core and @atlasmap/atlasmap.

PR #3052, opened 2021-07-06
fix: Add support for editing CSV processing parameters.

pleacu:
Fixes: #2990

As stated in the issue - this adds full CSV parameter edit support and solves the Atlasmap cancel issue.  We can open a new issue or reopen this one if there are Syndesis integration problems.

### DIFF
--- a/ui/packages/atlasmap-core/src/contracts/documents/csv.ts
+++ b/ui/packages/atlasmap-core/src/contracts/documents/csv.ts
@@ -14,7 +14,7 @@
     limitations under the License.
 */
 
-import { IDocument, IField } from '../common';
+import { IDocument, IField, IParameter } from '../common';
 
 /**
  * The CSV inspection data model contracts between frontend and backend.
@@ -65,4 +65,101 @@ export interface ICsvField extends IField {
  */
 export interface ICsvFields {
   csvField: ICsvField[];
+}
+
+export class ICsvParameterOptions {
+  // TODO: Retrieve from the backend CSV module
+  static getCsvParameterOptions(): IParameter[] {
+    return [
+      {
+        name: 'format',
+        label: 'CSV File Format',
+        value: 'Default',
+        options: [
+          { label: 'Default', value: 'Default' },
+          { label: 'Excel', value: 'Excel' },
+          { label: 'InformixUnload', value: 'InformixUnload' },
+          { label: 'InformixUnloadCsv', value: 'InformixUnloadCsv' },
+          { label: 'MongoDBCsv', value: 'MongoDBCsv' },
+          { label: 'MongoDBTsv', value: 'MongoDBTsv' },
+          { label: 'MySQL', value: 'MySQL' },
+          { label: 'Oracle', value: 'Oracle' },
+          { label: 'PostgreSQLCsv', value: 'PostgreSQLCsv' },
+          { label: 'PostgreSQLText', value: 'PostgreSQLText' },
+          { label: 'RFC4180', value: 'RFC4180' },
+          { label: 'TDF', value: 'TDF' },
+        ],
+        required: true,
+      },
+      {
+        name: 'allowDuplicateHeaderNames',
+        label: 'Allow Duplicate Header Names',
+        value: 'true',
+        boolean: true,
+        required: false,
+      },
+      {
+        name: 'allowMissingColumnNames',
+        label: 'Allow Missing Column Names',
+        value: 'true',
+        boolean: true,
+        required: false,
+      },
+      {
+        name: 'commentMarker',
+        label: 'Comment Marker',
+        value: '',
+        required: false,
+      },
+      {
+        name: 'delimiter',
+        label: 'Delimiter',
+        value: '',
+        required: false,
+      },
+      { name: 'escape', label: 'Escape', value: '', required: false },
+      {
+        name: 'firstRecordAsHeader',
+        label: 'First Record As Header',
+        value: 'true',
+        boolean: true,
+        required: false,
+      },
+      { name: 'headers', label: 'Headers', value: '', required: false },
+      {
+        name: 'ignoreEmptyLines',
+        label: 'Ignore Empty Lines',
+        value: 'true',
+        boolean: true,
+        required: false,
+      },
+      {
+        name: 'ignoreHeaderCase',
+        label: 'Ignore Header Case',
+        value: 'true',
+        boolean: true,
+        required: false,
+      },
+      {
+        name: 'ignoreSurroundingSpaces',
+        label: 'Ignore Surrounding Spaces',
+        value: 'true',
+        boolean: true,
+        required: false,
+      },
+      {
+        name: 'nullString',
+        label: 'Null String',
+        value: '',
+        required: false,
+      },
+      {
+        name: 'skipHeaderRecord',
+        label: 'Skip Header Record',
+        value: 'true',
+        boolean: true,
+        required: false,
+      },
+    ];
+  }
 }
--- a/ui/packages/atlasmap/src/Atlasmap/dialogs/useParametersDialog.tsx
+++ b/ui/packages/atlasmap/src/Atlasmap/dialogs/useParametersDialog.tsx
@@ -13,8 +13,9 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 */
-import { IParameter, ParametersDialog } from '../../UI';
-import React, { ReactElement, useCallback, useState } from 'react';
+import React, { ReactElement, useCallback, useEffect, useState } from 'react';
+import { IParameter } from '@atlasmap/core';
+import { ParametersDialog } from '../../UI';
 
 import { useToggle } from '../../Atlasmap/utils';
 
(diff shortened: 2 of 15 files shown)


ui/packages/atlasmap/src/Atlasmap/dialogs/useParametersDialog.tsx
```diff
@@ -20,6 +20,100 @@ import { useToggle } from '../../Atlasmap/utils';
 
 type ParametersCallback = (parameters: IParameter[]) => void;
 
+export function getCSVParameterOptions(): IParameter[] {
```
igarashitm, line 23: Could this be moved into `contracts/` as an interface, e.g. `ICSVParameterOptions`? maybe IParameter as well.

igarashitm (commented):
Is it possible to add a story demonstrates showing preconfigured values, possibly in `ParametersDialog.stories.tsx`?

ui/packages/atlasmap/src/Atlasmap/dialogs/useParametersDialog.tsx
igarashitm: It might have to be a class to have values, possibly with `static readonly` - https://stackoverflow.com/questions/37265275/how-to-implement-class-constants

igarashitm: And ideally these parameter list should be retrieved from backend as the original source is the backend CSV module. But it would need a bigger surgery to establish a procedure to retrieve inspection parameters beforehand.

pleacu: @igarashitm - do you see this as an addition to the atlasmap-core contracts or as  a new folder atlasmap/src/contracts?

pleacu: Actually it probably makes more sense in atlasmap-core/src/contracts since we could use them later if/when we retrieve the inspection params from the backend

igarashitm: The existing `@atlasmap/core:/contracts/documents/csv.ts` I would say, in principle @atlasmap/atlasmap shouldn't declare any contracts but just use the one from @atlasmap/core. We'll even migrate existing (#2975, #2883).

pleacu: Made stories run the Parameters dialog for both select and edit.
Established ICsvParameterOptions static class in contracts.

ui/packages/atlasmap-core/src/contracts/documents/csv.ts
```diff
@@ -66,3 +66,115 @@ export interface ICsvField extends IField {
...
+  name: string;
+  label: string;
+  value: string;
+  boolean?: boolean;
+  options?: IParameterOption[];
+  enabled?: boolean;
+  required?: boolean;
+}
+
+export class ICsvParameterOptions {
```
igarashitm, line 85: ah maybe even
```typescript
const CSV_PARAMETER_OPTIONS: IParameter[] = [
  {
    name: 'format',
...
```

pleacu: I guess if we're going to eventually get this from the BE then having a class there will mean less changes going forward.  But we can certainly go with a const rather than static class.  wdyt?  I'll squash once all is decided.

igarashitm: I'd vote on const - I think retrieving from backend won't be implemented soon.

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.stories.tsx
```diff
@@ -127,12 +127,26 @@ const parameters = [
...
+ */
 export const parametersDialog = () => (
-  <ParametersDialog
-    title={text('Title', 'Select CSV Processing Parameters')}
-    isOpen={boolean('Is open', true)}
-    onCancel={action('onCancel')}
-    onConfirm={action('onConfirm')}
-    parameters={parameters}
-  />
+  <span>
```
igarashitm, line 137: Not sure why it's not showing both, but would be better to have separate story anyway, see https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap/src/Views/MappingTableView.stories.tsx#L29-L45 for having multiple in a same file

ui/packages/atlasmap-core/src/contracts/documents/csv.ts
pleacu: ok - using const

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.stories.tsx
pleacu: corrected

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.stories.tsx
```diff
@@ -20,7 +20,7 @@ import React from 'react';
 import { action } from '@storybook/addon-actions';
 
 const obj = {
-  title: 'UI|Dialogs',
+  title: 'Story UI|Dialogs',
```
igarashitm, line 23: Is there a reason to have a separate `Story UI` category in storybook?

ui/packages/atlasmap-core/src/contracts/documents/csv.ts
pleacu: const arrays are still mutable in typescript interestingly enough.  Consequently with
 `const CSV_PARAMETER_OPTIONS: IParameter[] = [...`
you can legally set `CSV_PARAMETER_OPTIONS[3]`.  This resulted in cross-dialog element corruption where the constant array from csv,ts gets changed.  I even tried
`const CSV_PARAMETER_OPTIONS: ReadonlyArray<IParameter> = [` and it still got munged.
I think the best way is to return a fresh static class array.  That works fine in all cases.

igarashitm: `as const` might help? - https://stackoverflow.com/questions/43120022/i-want-totally-immutable-object-in-ts

ui/packages/atlasmap/src/UI/dialogs/ParametersDialog.stories.tsx
igarashitm: Can you get this back to `UI|Dialogs`? It created a new upper category `Story UI` in storybook separated from other dialogs, I don't think that's necessary.

pleacu: done

ui/packages/atlasmap-core/src/contracts/documents/csv.ts
```diff
@@ -66,3 +66,115 @@ export interface ICsvField extends IField {
 export interface ICsvFields {
   csvField: ICsvField[];
 }
+
+export interface IParameterOption {
```
igarashitm, line 70: oops overlooked at first, can we move these `IParameterOption` and `IParameter` to ../common.ts? They look like generic inspection parameter types that could be used for other document formats.

ui/packages/atlasmap-core/src/contracts/documents/csv.ts
pleacu: We can't use `as const` with `'IParameter[]` since it's a mutable type.  We'd have to create another local immutable type reference or cast `as IParameter` on all references.  I'd like to keep as is now if possible...

ui/packages/atlasmap-core/src/contracts/documents/csv.ts
pleacu: done

ui/packages/atlasmap-core/src/contracts/documents/csv.ts
igarashitm: I mean isn't it possible to have `const CSV_PARAMETER_OPTIONS: IParameter[] = [..] as const`?

pleacu: No - you get this error:
`const CSV_PARAMETER_OPTIONS: IParameter[]`
`The type 'readonly [{ readonly name: "format"; readonly label: "CSV File Format"; readonly value: "Default"; readonly options: readonly [{ readonly label: "Default"; readonly value: "Default"; }, { readonly label: "Excel"; readonly value: "Excel"; }, ... 9 more ..., { ...; }]; readonly required: true; readonly enabled: true; ...' is 'readonly' and cannot be assigned to the mutable type 'IParameter[]'.ts(4104)`

igarashitm: So it doesn't enforce on existing types... pity. LGTM then, can you squash the commits?

igarashitm (approved):

Issue #3051, opened 2021-07-06
Remove double slash from Java inspection field path


igarashitm:
Java inspection has been returning invalid double slash but is corrected in UI due to #3030

Issue #3030, opened 2021-07-01
UI: Remove field path recalculation


igarashitm:
It doesn't make sense to recalculate field path at UI side, all fields must receive complete path through inspection and UI just has to use it.
https://github.com/atlasmap/atlasmap/blob/704eeb763eb01357bd2e2572d95bad24763f2b0a/ui/packages/atlasmap-core/src/models/document-definition.model.ts#L442-L473

PR #3053, opened 2021-07-06
fix: UI: Remove field path recalculation

igarashitm:
Fixes: #3030

Remove double slash from Java inspection field path
Fixes: #3051


pleacu (approved):
+1

Issue #3127, opened 2021-07-19
[TASK] Remove junit4


igarashitm:
Migrate all tests to junit5 jupiter

PR #3131, opened 2021-07-20
fix: Remove junit4

igarashitm:
Fixes: #3127
Also fixed deprecated JDK API usage


pleacu (approved):
+1  Wow - looks like JUnit5 Jupiter touched everything

Issue #3135, opened 2021-07-22
[BUG] expression field search doesn't establish mapped field as expected


igarashitm:
**Describe the bug**
It seems to be a regression after refactoring
![atlasmap-fail](https://user-images.githubusercontent.com/265462/126645706-c5a3684d-84a8-41d8-b8da-e0b68325c148.gif)

It works fine if mapping is established through dnd and convert it to conditional mapping
![atlasmap](https://user-images.githubusercontent.com/265462/126646335-a081ba2f-c951-48fb-9f79-b8515b10fc63.gif)

pleacu:
Working on this now

pleacu:
Next PR addresses this:

![sc0](https://user-images.githubusercontent.com/2981421/126776477-f172dbd3-f7ce-46fc-bdd8-02429e7946ca.png)

![sc1](https://user-images.githubusercontent.com/2981421/126776808-b7e612d0-97d8-4837-8afb-19dec67ef955.png)

Issue #3029, opened 2021-07-01
UI: expression can't distinguish same path fields in different documents


igarashitm:
This won't work if same path fields exist in multiple documents, should check the Document ID as well.
https://github.com/atlasmap/atlasmap/blob/704eeb763eb01357bd2e2572d95bad24763f2b0a/ui/packages/atlasmap-core/src/services/mapping-expression.service.ts#L156-L161

PR #3142, opened 2021-07-23
fix: Added a doc ID to mapping model queries and corrected adding exp…

pleacu:
…ression field nodes to the active mapping.

Fixes: #3135
Fixes: #3029


ui/packages/atlasmap-core/src/models/config.model.ts
```diff
@@ -268,6 +268,14 @@ export class ConfigModel {
     return this.getDocs(isSource).find((d) => d.id === documentId)!;
   }
 
+  getDocForName(
```
igarashitm, line 271: IIRC we allow duplicated document name, how do we distinguish them?

ui/packages/atlasmap-core/src/services/mapping-expression.service.spec.ts
```diff
@@ -168,4 +174,56 @@ describe('MappingExpressionService', () => {
     const mapping1 = cfg.mappings!.mappings[1];
     expect(service.createMappingExpression(mapping1)).toContain('Concatenate');
   });
+
+  const mockstr = () => {
```
igarashitm, line 178: Is there a special reason to have this as a function?

ui/packages/atlasmap-core/src/models/expression.model.ts
```diff
@@ -93,7 +93,16 @@ export class FieldNode extends ExpressionNode {
           )!;
         }
         if (!this.mappedField) {
-          this.mappedField = mapping.getReferenceField(fieldParts[1]);
+          const docDef = ConfigModel.getConfig().getDocForName(
```
igarashitm, line 96: The field expression is `<Document ID>:<Field path>`, not the document name I think

ui/packages/atlasmap-core/src/models/config.model.ts
pleacu: removed - this function is not needed

ui/packages/atlasmap-core/src/services/mapping-expression.service.spec.ts
pleacu: removed mockstr - just added inline nameless fcns

ui/packages/atlasmap-core/src/models/expression.model.ts
pleacu: Ah - you're right - can use fieldParts[0] as is since it's already docId str

pleacu:
@igarashitm - can you tell what's wrong with the windows build?  It's complaining about 

`[INFO] @atlasmap/atlasmap: (typescript) Error: D:/a/atlasmap/atlasmap/ui/packages/atlasmap/src/impl/utils/ui.ts(344,33): semantic error TS2339: Property 'addFieldToExpression' does not exist on type 'MappingExpressionService'.
[INFO] @atlasmap/atlasmap: Error: D:/a/atlasmap/atlasmap/ui/packages/atlasmap/src/impl/utils/ui.ts(344,33): semantic error TS2339: Property 'addFieldToExpression' does not exist on type 'MappingExpressionService'.
`
But `addFieldToExpression` is declared and I have a clean maven build from Linux.

ui/packages/atlasmap-core/src/services/mapping-expression.service.spec.ts
```diff
@@ -168,4 +174,59 @@ describe('MappingExpressionService', () => {
...
+    source.parentField = new Field();
+    source.parentField.isCollection = true;
+    source.parentField.parentField = new Field();
+    source.parentField.parentField.isCollection = true;
+    jsonSchemaSource.addField(source);
+
+    expect(mapping).toBeDefined();
+    cfg.mappingService.selectMapping(mapping);
+
+    const textNode: ExpressionNode = {
```
igarashitm, line 204: This could be
```javascript
const textNode: ExpressionNode = {
      uuid: '0',
      str: 'mockstr',
      getUuid: () => '',
      toText: () => '',
      toHTML: () => '',
    };
```

pleacu:
For the windows build issue - I changed `addField` to `addFieldToExpression` in the `mapping-expression` service.  Could the windows build not be cleaning up the generated .d.ts files?

ui/packages/atlasmap-core/src/services/mapping-expression.service.spec.ts
pleacu: sure - np

igarashitm:
Maybe because mapping-expression.service is not exported in the index.ts? not sure why it's OK on other platforms.

pleacu:
I updated the index.ts with no success on Windows.  I can move the name back to `addField` and squash so this PR can go through - wdyt?

igarashitm:
Let's squash and merge and then file an issue for windows build failure.

Issue #3141, opened 2021-07-23
[BUG] Wrong path for Java collection field


igarashitm:
**Describe the bug**
Java collection field gets wrong field path in the mapping. It doesn't reproduce with JSON.
 
**To Reproduce**
1. Import java test-model jar (https://github.com/atlasmap/atlasmap/tree/master/lib/modules/java/test-model)
2. Add `io.atlasmap.java.test.TargetTestClass` as a Document
2. Map `contactList<>/firstName`
3. See mapping definition sent to backend, field path is `/contact/firstName` while it's supposed to be `contactList<>/firstName`
```json
{
  "AtlasMapping" : {
    "jsonType" : "io.atlasmap.v2.AtlasMapping",
    "dataSource" : [ {
      "jsonType" : "io.atlasmap.v2.DataSource",
      "id" : "io.atlasmap.java.test.TargetTestClass",
      "name" : "TargetTestClass",
      "description" : "Java document class io.atlasmap.java.test.TargetTestClass",
      "uri" : "atlas:java:io.atlasmap.java.test.TargetTestClass?className=io.atlasmap.java.test.TargetTestClass",
      "dataSourceType" : "SOURCE"
    }, {
      "jsonType" : "io.atlasmap.v2.DataSource",
      "id" : "io.atlasmap.java.test.TargetTestClass",
      "name" : "TargetTestClass",
      "description" : "Java document class io.atlasmap.java.test.TargetTestClass",
      "uri" : "atlas:java:io.atlasmap.java.test.TargetTestClass?className=io.atlasmap.java.test.TargetTestClass",
      "dataSourceType" : "TARGET"
    } ],
    "mappings" : {
      "mapping" : [ {
        "jsonType" : "io.atlasmap.v2.Mapping",
        "inputFieldGroup" : {
          "jsonType" : "io.atlasmap.v2.FieldGroup",
          "actions" : [ ],
          "field" : [ {
            "jsonType" : "io.atlasmap.java.v2.JavaField",
            "docId" : "io.atlasmap.java.test.TargetTestClass",
            "path" : "/contact/firstName",
            "fieldType" : "STRING",
            "name" : "firstName"
          } ]
        },
        "inputField" : [ ],
        "outputField" : [ {
          "jsonType" : "io.atlasmap.java.v2.JavaField",
          "docId" : "io.atlasmap.java.test.TargetTestClass",
          "path" : "/contact/firstName",
          "fieldType" : "STRING",
[...]

PR #3146, opened 2021-07-23
chore: Add TargetTestClass for an UI test resource

igarashitm:
@pleacu #3141 seems to be rather Java side issue, Java inspection doesn't have collection suffix on the field path as expected. This PR will add TargetTestClass to be inspected and generate `ui/test-resources/inspected/atlasmap-inspection-io.atlasmap.java.test.TargetTestClass.json`


pleacu (approved):
+1

Issue #3213, opened 2021-08-17
[BUG] Fail to create constant field


sevenbamboos:
**Describe the bug**
Constants can't be created on mapper UI.
 
**Attach debug logs**
This is the error from browser's console:
```
hostReportError.ts:8 Uncaught TypeError: Cannot read property 'split' of undefined
    at e.n.getFieldLabel (field.model.ts:246)
    at Rt (ui.ts:107)
    at _t (ui.ts:134)
    at Array.map (<anonymous>)
    at kt (ui.ts:152)
    at AtlasmapProvider.tsx:266
    at AtlasmapProvider.tsx:337
    at t._next (AtlasmapProvider.tsx:400)
    at t.__tryOrUnsub (Subscriber.ts:266)
    at t.next (Subscriber.ts:208)
```
 
**To Reproduce**
1. Start to create a new mapping.
2. Click '+' to create a (String) constant.
3. Fill in the name and value fields, and then click 'Confirm' button.

**Expected behavior**
A new constant should be created and shown on mapper UI.

**Actual behavior**
Nothing shown on UI. There is javascript error in browser's console (see above).

**Environment (please complete the following information):**
 - OS: Mac
 - Browser: chrome
 - Version: 2.3.0-SNAPSHOT

**Additional context**
`getFieldLabel` in field.model.ts is expecting a path property, but is undefined for constant fields.

On the java backend, however, the mapping json has been updated correctly with the new constant.

It's working fine on 2.2.3.

It seems something to do with this change - https://github.com/atlasmap/atlasmap/commit/31e9c96a550a95864736f370c0714eec8f572d38
The path property isn't got populated during `addField` in document-definition.model.ts but is still used on UI.

PR #3216, opened 2021-08-20
fix: Fail to create constant field

igarashitm:
Fixes: #3213
I'll add an UT later with those routines migrated to document-management.service (#2883)


pleacu (approved):
+1

Issue #3218, opened 2021-08-20
[BUG] UI: editConstant() doesn't update the path, makes it out of sync


igarashitm:
The original path is used if constant name is modified.
```json
{
  "AtlasMapping" : {
    "jsonType" : "io.atlasmap.v2.AtlasMapping",
    "dataSource" : [ ],
    "mappings" : {
      "mapping" : [ {
        "jsonType" : "io.atlasmap.v2.Mapping",
        "inputField" : [ {
          "jsonType" : "io.atlasmap.v2.ConstantField",
          "value" : "test",
          "docId" : "DOC.Constants.269771",
          "path" : "/test",
          "fieldType" : "STRING",
          "name" : "testmod"
        } ],
        "outputField" : [ {
          "jsonType" : "io.atlasmap.v2.PropertyField",
          "docId" : "DOC.Properties.47778",
          "path" : "/current/test",
          "fieldType" : "STRING",
          "name" : "test",
          "scope" : "current"
        } ],
        "id" : "mapping.499735"
      } ]
    },
    "lookupTables" : {
      "lookupTable" : [ ]
    },
    "constants" : {
      "constant" : [ {
        "name" : "testmod",
        "value" : "test",
        "fieldType" : "STRING"
      } ]
    },
    "properties" : {
      "property" : [ {
        "name" : "test",
        "fieldType" : "STRING",
        "scope" : "current",
        "dataSourceType" : "TARGET"
      } ]
    },
    "name" : "UI.0"
  }
```

PR #3219, opened 2021-08-20
fix: Move constant/property handling routine to @atlasmap/core (#2883)

igarashitm:
Also add UTs for #3213

fix: editConstant() doesn't update the path, makes it out of sync
Fixes: #3218


pleacu (approved):
+1

Issue #3197, opened 2021-08-12
Inherited Properties Duplicated Within UI


ronMilne:
The UI Editor duplicates properties from the base class (i.e. extends) In the screen shot below the properties are private the setter/getter public, but the properties are duplicated (UserIdentity, TransactionId, SessionId) are defined within the base
![duplicateUI](https://user-images.githubusercontent.com/24907744/129275867-ed85bf3c-bb27-43f6-bbd8-b9c18ca9237e.png)

the attached zip includes the loaded type
[searchNniName-model.zip](https://github.com/atlasmap/atlasmap/files/6978794/searchNniName-model.zip)

AtlasMap 2.2.1

igarashitm:
Can you upload ADM file?

PR #3230, opened 2021-08-23
fix: Hidden Java Field is shwon duplicated

igarashitm:
Fixes: #3197
Since Java hidden field is generally a bad practice and doesn't work well together with AtlasMap path expression, we'll simply ignore it.


pleacu (approved):
+1

PR #3236, opened 2021-08-27
Fixed issue where a source Java Enum that is null generates an Audit ERROR

johnathani:
The issue occurs when the enum in the source Java Object is null and is required to be mapped to an enum in a destination Java Object.
This generates a mapping error as it tries to create enum using the writerUtil.createComplexChildObject which tries to call a constructor on the Java Enum class.

The fix ignores this branch of code if the value complex and not a JavaEnum Field and null.
The null enum now branches to use the writerUtil.setChildObject which sets the destination as null.

There README.md explains the issue.


igarashitm:
Can you
* Add a testcase to verify the issue is fixed
* Correct the commit message to follow the guideline

johnathani:
Hi
I am deleting this pull request and creating a new pull request with the correct commit messages and test case

Issue #3239, opened 2021-08-27
[BUG] Mapping a source enum when its null generates and mapping error 


johnathani:
**Describe the bug**
Using Java POJO source to Java POJO destination mapping, when mapping a source enum that is null to a destination enum, the mapping fails with a mapping error.
This was replicated in the Camel environment.

**Attach .adm file and source document examples**
Please see publicly available repo with source and ADM demonstrating the issue.
https://github.com/johnathani/AtlasMap-Enum-Error.git
 
**Attach debug logs**
Exception is raised with mapping error.
 
**To Reproduce**
Please see README.md for https://github.com/johnathani/AtlasMap-Enum-Error.git

**Expected behavior**
It is expected that id an enum is null on the source side, that the null is mapped to the enum on the destination side.

**Screenshots**
Please see README.md for https://github.com/johnathani/AtlasMap-Enum-Error.git

**Environment (please complete the following information):**
 - OS: Windows and Mac OS
 - Version 2.2.3

**Additional context**

PR #3253, opened 2021-08-31
Fixed issue where a source Java Enum that is null generates an Audit ERROR due to underlying JavaWriter exception

johnathani:
Fixes: #3239

The issue occurs when the enum in the source Java Object is null and is required to be mapped to an enum in a destination Java Object.
This generates a mapping error as it tries to create enum using the writerUtil.createComplexChildObject which tries to call a constructor on the Java Enum class.

The exception is:
io.atlasmap.api.AtlasException: Unable to create value for segment: statesLong parentObject: io.atlasmap.java.test.TargetTestClass

The fix does not write the complex value if its null and rather uses the writerUtil.setChildObject which sets the destination as null.

There README.md explains the issue.


igarashitm:
Looks good - can you squash the 2 commits into one?

johnathani:
Closing pull request and creating a new one for single commit

Issue #3275, opened 2021-09-08
[BUG] Java to Java null source complex type creates an empty target complex type instead of setting the target to null  


johnathani:
**Describe the bug**
With a one to one mapping.
If the source has a complex type with fields mapped to a target complex type fields then
   if the source complex type is null, it still creates an instance of the target type with all fields null instead of the target complex type being set to null.

This is really problematic in mappings where the target POJO is used in web services such as JSON or SOAP payloads as the created target complex type will add XML (example for SOAP) for outer structure without the inner values instead of leaving the XML out totally.

The expected behaviour is that a null instance of a source complex type does not created an instance on the target. 

**Attach .adm file and source document examples**
Please refer to publish project
https://github.com/johnathani/AtlasMap-Enum-Error.git
This demonstrates the expected null to null behaviour for mapping in the following test methods
            testObjectNullSubObjPartialMap(context);
            testObjectAllNullMap(context);
            testObjectNullMap(context);
 
**Expected behavior**
The expected behaviour is that a null complex type on the source side does not create an instance of a complex type on the target side. Technically this is if all source field values are null that are to be mapped to all target fields will not instantiate the parent complex type containing the target fields.

**Environment (please complete the following information):**
 - OS: MacOS, Windows
 - Version 2.2.3 and 2.3.0-SNAPSHOT

PR #3276, opened 2021-09-08
fix: Fixed mapping error for null complex type to null complex type i…

johnathani:
Fixes: #3275

The issue occurs when a source complex type in a POJO has mappings targeting fields in the target POJO, but the source complex type is null.

This ends up creating an instance of the target POJO with all fields null.
The expected behaviour is that the target remains null of the source is null.

The fix applies a concept of lazy object instantiation, only instantiating an instance of the target object when a source mapped field value is non null
  UNLESS there is an explicit mapping to create the object instance (complex type) such as can be contained in JSON -> Java mapping
  Example: 
     ...
    {
        "jsonType" : "io.atlasmap.v2.Mapping",
        "mappingType" : "MAP",
        "inputField" : [ {
          "jsonType" : "io.atlasmap.json.v2.JsonField",
          "path" : "/Contact"
        } ],
        "outputField" : [ {
          "jsonType" : "io.atlasmap.java.v2.JavaField",
          "path" : "/contact",
          "fieldType" : "COMPLEX",
          "className" : "io.atlasmap.java.test.TargetContact"
        }

The test case was modified to test the output for the expected behaviour.


igarashitm (approved):
Thank you for the contribution, LGTM

PR #3280, opened 2021-09-09
Include ui/dist folder in .gitignore

apupier:
it avoids having a lot of modified files listed after a full build

### DIFF
--- a/ui/.gitignore
+++ b/ui/.gitignore
@@ -1,6 +1,6 @@
 node
 _netlify
-
+dist
 # compiled output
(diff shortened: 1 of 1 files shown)


igarashitm:
I've never seen `ui/dist` dir is created, but I do see ``ui/packages/*/dist` which is alaredy in the ingore list. Something wrong in your build environment maybe?

apupier:
effectively, it was froma very old build in January 2020, I do not work often on AtlasMap.

Sorry for the disturbance

PR #3323, opened 2021-09-29
ENTESB-15784 - support space in path to atlasmap workspace when loading

apupier:
jars

fixes: #3324 ENTESB-15784

URL.getPath should not be used directly. it needs to go through toURi

see Javadoc https://docs.oracle.com/javase/7/docs/api/java/net/URL.html

The URL class does not itself encode or decode any URL components
according to the escaping mechanism defined in RFC2396. It is the
responsibility of the caller to encode any fields, which need to be
escaped prior to calling URL, and also to decode any escaped fields,
that are returned from URL. Furthermore, because URL has no knowledge of
URL escaping, it does not recognise equivalence between the encoded or
decoded form of the same URL

Note, the URI class does perform escaping of its component fields in
certain circumstances. The recommended way to manage the encoding and
decoding of URLs is to use URI, and to convert between these two classes
using toURI() and URI.toURL().

Fixes: #

~~I failed so far to write a test but creating PRs to start showing work and test that i ti snot breaking something else~~ updated with a modified test

### DIFF
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasLibraryLoader.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasLibraryLoader.java
@@ -21,6 +21,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Modifier;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


igarashitm:
Can you file a GH issue (maybe with a cross reference link between JIRA) and follow this?
https://github.com/atlasmap/atlasmap/wiki/AtlasMap-Developer-Memo#pull-request

apupier:
seems there is another place which is not supporting space in path of the workspace:

```
Error:  Errors: 
Error:    AtlasServiceTest.testGetMapping:113 NullPointer
```

apupier:
I restrained using workspace with space in path to the jar upload test case to get rid of previous issue whcih was a problem with Test only.

I created 2 cases for jarUpload test which requires to have a clean library folder. it seems that the AtlasMapService.clearLibraries is not working on Windows based on the error:
```
Error:  Failures: 
Error:    AtlasServiceTest.testJarUploadWithSpaceInWorkspacePath:145->testJarUpload:226 expected: <1> but was: <2>
```

igarashitm (approved):

PR #3330, opened 2021-09-30
fix: null index causes NPE in field action service

igarashitm:
Fixes: #3329
Implicit unboxing causes NPE


lib/core/src/main/java/io/atlasmap/core/DefaultAtlasFieldActionService.java
```diff
@@ -799,7 +799,7 @@ private void extractFlatListValuesFromFieldGroup(AtlasInternalSession session,
                 // Collection field
                 List<Field> fields = ((FieldGroup) subField).getField();
                 if (fields != null && !fields.isEmpty()) {
-                    if (fieldGroup.getField().size() == 1 && (index == 0 || index == null)) {
+                    if (fieldGroup.getField().size() == 1 && (index == null || index == 0)) {
```
pleacu, line 802: Just trying to grok this - testing for 0 or null with the parenthetical expression shouldn't matter - how did this make a difference?

igarashitm: So if index is null, the former tries to "unbox" `Integer` into primitive `int` to evaluate `index == 0`, causes NPE.

pleacu: ah - sneaky - good catch...

pleacu (approved):
+1

Issue #3328, opened 2021-09-30
[TASK] Add xalan dependency explicitly on atlas-xml-core


igarashitm:
The default XML transformer xalan is not always available through JDK, e.g. in camel-k
cf. https://github.com/apache/camel-k/issues/2534

Add it explicitly in atlas-xml-core.

igarashitm:
It turned out xalan 2.7.2 behaves a bit different from JDK default one.

PR #3332, opened 2021-09-30
chore: Add xalan dependency explicitly on atlas-xml-core

igarashitm:
Fixes: #3328


igarashitm:
Closing - xalan 2.7.2 behaves a bit different, it doesn't add `standalone="no"` to the header while JDK built-in one does. Not a problem practically, but potentially there's more.

Issue #3303, opened 2021-09-24
[BUG] Constant can be saved without any name/value or name even though they are required.


mkralik3:
**Describe the bug**
The modal dialog for creating constant in DataMapper shows that Name/Values fields are required but I am able (Create button is enabled) to create constant without any Name/Value (*)
![image](https://user-images.githubusercontent.com/16251792/134657417-a2be750e-97d0-453e-b099-6f336c876ec4.png)
or without Name.
![image](https://user-images.githubusercontent.com/16251792/134657505-a65c3e66-c943-4781-8cb9-478cf4df0648.png)
* only right after open the dialog, when I start typing name, the button is disabled.
![image](https://user-images.githubusercontent.com/16251792/134657436-9da018cd-1038-4842-a15e-90140446350c.png)

Original JIRA: https://issues.redhat.com/browse/ENTESB-17377

PR #3394, opened 2021-10-13
fix: Correct validation of constant and property input (update)

pleacu:
Correct where a legal name is specified then backed over to an empty name
Fixes: #3303


pleacu:
Not sure how  the stories appear in the log.  Added your suggestion - all passes.

igarashitm:
So... `onValidation` doesn't affect the confirm button enabled/disabled state at all, isn't it? It works as expected even in a story, i.e. if I type name and value then button is enabled, and disabled if I remove it. It's built-in in ConstantDialog.

Issue #3399, opened 2021-10-14
[TASK] Better error message when META-INF/services/* files for custom action are missing


igarashitm:
e.g. getting this when uploading jar. Even a debug log doesn't really explain what is wrong.
```
javax.ws.rs.WebApplicationException: Could not read file part: Invalid type: class io.atlasmap.service.my.PaulsFieldActions$Concur
```

PR #3402, opened 2021-10-14
fix: Better error message when META-INF/services/* files for custom a…

igarashitm:
…ction are missing

Fixes: #3399

fix: Reset all doesn't dispose jar on windows

Fixes: atlasmap#3403
Since URLClassLoader keeps file handler open, on windows, loaded jar file can't be removed until URLClassLoader#close() is invoked and file handlers are released.
Unfortunately even URLClassLoader#close() didn't help on windows+java8. Thus tests are conditionally skipped.


igarashitm:
ClassLoader doesn't reset on windows... investigating

Issue #3404, opened 2021-10-15
[BUG] [VSCode] Shortcut keys are ignored when Source or Target Column have received the last event


apupier:
**Describe the bug**

In VS Code, when clicking inside the Source or Target column, shortcut keys are no more available (for instance Ctrl+Shift+P to open command palette which is the central action in VS Code).

**Attach .adm file and source document examples**
The .adm archive file you can export from AtlasMap UI contains all metadata to reproduce your mapping. Also don't forget to provide source document examples to reproduce the runtime behavior exactly what you see.
 
**Attach debug logs**
If it's a design time issue, attach a backend log with debug enabled
`java -Dlogging.level.io.atlasmap=debug -jar atlasmap-standalone.jar`
or for runtime issue, set the log level for `io.atlasmap` package to debug in your log4j config
 
**To Reproduce**
1. In VS Code, call "Open Atlasmap" command
2. Wait that the AtlasMap Ui is loaded
3. Do one of these actions:
   - click to collapse or expand Properties, or Constants panel
   - navigate with "Tab" until reaching an element inside source or target column
   - Add a constant or a properties
   - a lot of other actions inside Source and a target column outside of the search bar
4. Try to call Ctrl+Shift+P --> nothing is happening

workaround consists in clicking elsewhere than the source/Target column.

**Expected behavior**

Shortcut keys should still be working.

**Screenshots**

one of the use case demoed here:
https://user-images.githubusercontent.com/1105127/137482310-04860753-f988-4796-b2c3-017c110df207.mp4

**Environment (please complete the following information):**
 - OS:Fedora 34
 - Browser [e.g. chrome, safari] VS Code
 - Version [e.g. 22] tried with 2.2 and 2.3

**Additional context**

With the move of Atlasmap in an editor in VS Code, this is more visible and impacting as it is affecting the `Ctrl+s` functionality which is key in the workflow usage of it.

apupier:
My newbie investigation leads me to think that there is a problem with a [ForwardRef](https://reactjs.org/docs/forwarding-refs.html) given that:
* found several mentions of event problems with ref and frowardref in different forums
* the issue seems to affect everythign which is inside the [nodeRef/ColumnBody](https://github.com/atlasmap/atlasmap/blob/cea516186d279c8ba1ad0299333a6cbb4b8530cf/ui/packages/atlasmap/src/Views/ColumnMapperView/Columns/SourcesColumn.tsx#L164)
[...]

PR #3457, opened 2021-11-03
fix: Remove event stop-propagation calls that were interfering with v…

pleacu:
…scode
Modified the tree focus provider and document keyboard handler to not stop event propogation.
Fixes: #3404


igarashitm (approved):

pleacu:
@igarashitm - Looks like the experimental Java 15 maven build is stuck.  All else is okay - should I squash and merge?

igarashitm:
yep, just ignore experimentals

Issue #3470, opened 2021-11-10
[BUG] Same name element with a different inline type in the XSD is wrongly `cached`


igarashitm:
**Describe the bug**

Say there's `/foo/bar/foo/else`, the 2nd `foo` is wrongly inspected as `cached` and UI disables it.
[Not_working.xsd.txt](https://github.com/atlasmap/atlasmap/files/7514478/Not_working.xsd.txt)

**Expected behavior**

Should not be cached with just having a same name. Check the type.

PR #3471, opened 2021-11-10
fix: Same name element with a different inline type in the XSD is wro…

igarashitm:
…ngly cached

Fixes: #3470
Now I can't remember why I thought we had to cache with element name in #2374. Put a real usecase if it's really needed.


pleacu (approved):
+1  Looks like a bad merge from before?

Issue #3507, opened 2021-11-22
[BUG] Mappings aren't restored in 1st .adm import attempt


igarashitm:
**Describe the bug**

The .adm import doesn't restore mappings in 1st loading attempt, although it succeeds in 2nd attempt. The 1st attempt here means 1st time after backend starts. Once backend is restarted then the symptom reproduces again.

**Attach .adm file and source document examples**

 https://github.com/atlasmap/atlasmap/blob/main/standalone/src/test/resources/json-schema-source-to-xml-schema-target.adm

**Expected behavior**

Mappings to be restored in 1st attempt.

**Screenshots**

![atlasmap](https://user-images.githubusercontent.com/265462/142891280-3c85a2d0-61a5-440d-9a4e-e984d08cfde3.gif)

PR #3508, opened 2021-11-22
fix: Mappings aren't restored in 1st .adm import attempt

igarashitm:
Fixes: #3507
This happens if there's a Java Document and the jar has to be loaded before mapping definition is deserialized in ADMArchiveHandler#persist()


pleacu (approved):
+1

Issue #3564, opened 2021-12-10
[BUG] Inconsistent field addition ban on DnD


igarashitm:
**Describe the bug**

Dragging a source field which is already in some mapping onto a free target field is banned if there's no active mapping, but allowed if unrelated mapping is active, which is inconsistent. Creating a new mapping in this case would be natural here.

![atlasmap](https://user-images.githubusercontent.com/265462/145588031-8d158146-921c-4f5a-aefb-dc1ce4e68a51.gif)

**Expected behavior**

Create a new mapping

Issue #3559, opened 2021-12-07
[BUG] Mutiple mappings to a non-collection target field should not be allowed


igarashitm:
cf. #3554 
It should **NOT** be allowed for non-collection target field as opposed to a collection target field case in #3554. Currently adding a target field through incremental search allows it. 
![atlasmap](https://user-images.githubusercontent.com/265462/145051929-16da4d1b-c1bd-4009-998c-9b9c389609cf.gif)

Issue #3554, opened 2021-12-06
[BUG] Multiple mappings to a collection target field should be allowed


igarashitm:
**Describe the bug**

~~The target field can participate only one mapping and D&D prohibits it successfully, but it seems adding a target field from incremental search is not. It would be better to exclude the target fields those are already in another mapping rather than showing an error to be symmetric with D&D.~~
I totally forgot we once allowed it for handling collection item, see https://github.com/atlasmap/atlasmap/issues/2289
So, the bug is in D&D behavior, it should be allowed if target field is in collection.

**Attach .adm file and source document examples**

 https://github.com/atlasmap/atlasmap/files/7659078/AtlasmapIssue.zip

**To Reproduce**

![atlasmap](https://user-images.githubusercontent.com/265462/144909985-0956e0af-9bb7-4701-92f8-e6e8cd5ed10f.gif)

**Expected behavior**

The target fields those are already in another mappings are excluded from candidates.

PR #3565, opened 2021-12-10
fix: Mutiple mappings to a non-collection target field should not be …

igarashitm:
…allowed

Fixes: #3559

Multiple mappings to a collection target field should be allowed
Fixes: #3554

Inconsistent field addition ban on DnD
Fixes: #3564


pleacu (approved):
+1

Issue #3560, opened 2021-12-07
[TASK] Fix unhandled promise rejection in UI tests


igarashitm:
It's only a warning with node 14 but it does fail with node 16, it seems. Currently we let frontend-maven-plugin use node 14 and CI build is fine, but better to fix those and bump it to node 16. Note that Fedora 35 provides node 16.

PR #3566, opened 2021-12-10
fix: Fix unhandled promise rejection in UI tests

igarashitm:
Fixes: #3560
Also updated to the latest node 16 and npm 8.3.0
node_modules would have to be rebuilt after this update. Remove it and run yarn install (or mvn install)


pleacu (approved):
+1

Issue #579, opened 2018-09-21
Refactor inspection services to be "on demand" style


igarashitm:
xref: https://issues.redhat.com/browse/ENTESB-12410

Today inspection services are single request per Document. It consumes whole schema and build entire Document model at once.

In order to handle huge Document model, we should investigate to refactor inspection services to be "on demand" style, which request an inspection chunk by chunk.

For example, if we have
```xml
<root>
  <child name="foo">
    <grand-child name="bar"/>
  </child>
</root>
```
The initial inspection would only returns `child`, i.e. first level sub-fields. When user expand `/root/child` in the UI, then it requests a second inspection to look into sub-fields of `child`, which returns `grand-child`. In other words, it requests inspections for each complex type, but only when user try to expand.

**The biggest hurdle** is the fact that Syndesis depends on offline inspection for Java Document to avoid Java classes to be required by AtlasMap design runtime. 2 options:

* Avoid on demand inspection if offline inspection result is fed
  * quicker but then Syndesis can't address #255 for Java Documents
* Migrate inspection service into syndesis-meta and provide online inspection even for Java Document
  * Introducing a non-flat class loader is not preferred. We have to wait for syndesis-meta to be completely serverless, so that the each invocation can build their own classpath

igarashitm:
The biggest hurdle is the fact that Syndesis depends on offline inspection for Java Document to avoid Java classes to be required by AtlasMap design runtime. 2 options:

* Avoid on demand inspection if offline inspection result is fed
  * quicker but then Syndesis can't address #255 for Java Documents
* Implement #561 and let Syndesis use online inspection by uploading required jars
  * more work

igarashitm:
An option to implement server side query against inspected document metadata
https://github.com/eiiches/jackson-jq

Worth a investigate. It also benefits server side field search #603

PR #3604, opened 2021-12-20
feat: Refactor inspection services to be "on demand" style

rkorytkowski:
Fixes: #579

### DIFF
--- a/lib/model/src/main/java/io/atlasmap/v2/Document.java
+++ b/lib/model/src/main/java/io/atlasmap/v2/Document.java
@@ -49,5 +49,4 @@ public class Document extends Field {
     public void setFields(Fields value) {
         this.fields = value;
     }
-
 }
--- a/lib/modules/json/model/src/main/java/io/atlasmap/json/v2/JsonInspectionRequest.java
+++ b/lib/modules/json/model/src/main/java/io/atlasmap/json/v2/JsonInspectionRequest.java
@@ -20,11 +20,12 @@ import java.io.Serializable;
 import com.fasterxml.jackson.annotation.JsonRootName;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
+import io.atlasmap.v2.BaseInspectionRequest;
 import io.atlasmap.v2.StringList;
 
 @JsonRootName("JsonInspectionRequest")
 @JsonTypeInfo(include = JsonTypeInfo.As.PROPERTY, use = JsonTypeInfo.Id.CLASS, property = "jsonType")
-public class JsonInspectionRequest implements Serializable {
+public class JsonInspectionRequest extends BaseInspectionRequest implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
@@ -42,11 +43,11 @@ public class JsonInspectionRequest implements Serializable {
 
     /**
      * Gets the value of the fieldNameExclusions property.
-     * 
+     *
      * @return
      *     possible object is
      *     {@link StringList }
-     *     
+     *
      */
     public StringList getFieldNameExclusions() {
         return fieldNameExclusions;
@@ -54,11 +55,11 @@ public class JsonInspectionRequest implements Serializable {
 
     /**
      * Sets the value of the fieldNameExclusions property.
-     * 
+     *
      * @param value
      *     allowed object is
      *     {@link StringList }
-     *     
+     *
      */
     public void setFieldNameExclusions(StringList value) {
         this.fieldNameExclusions = value;
@@ -66,11 +67,11 @@ public class JsonInspectionRequest implements Serializable {
 
     /**
      * Gets the value of the typeNameExclusions property.
-     * 
+     *
      * @return
      *     possible object is
      *     {@link StringList }
-     *     
+     *
      */
     public StringList getTypeNameExclusions() {
         return typeNameExclusions;
@@ -78,11 +79,11 @@ public class JsonInspectionRequest implements Serializable {
 
     /**
      * Sets the value of the typeNameExclusions property.
-     * 
+     *
      * @param value
      *     allowed object is
      *     {@link StringList }
-     *     
+     *
      */
     public void setTypeNameExclusions(StringList value) {
         this.typeNameExclusions = value;
@@ -90,11 +91,11 @@ public class JsonInspectionRequest implements Serializable {
 
     /**
      * Gets the value of the namespaceExclusions property.
-     * 
+     *
      * @return
      *     possible object is
      *     {@link StringList }
-     *     
+     *
      */
     public StringList getNamespaceExclusions() {
         return namespaceExclusions;
@@ -102,11 +103,11 @@ public class JsonInspectionRequest implements Serializable {
 
     /**
      * Sets the value of the namespaceExclusions property.
-     * 
+     *
      * @param value
      *     allowed object is
      *     {@link StringList }
-     *     
+     *
      */
     public void setNamespaceExclusions(StringList value) {
         this.namespaceExclusions = value;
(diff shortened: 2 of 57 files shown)


pleacu (commented):
This looks pretty good to me.  A lot of async layering but it probably needed that anyway.  If you can get this to pass the tests completely then I'd say your close to pushing it.  See what @igarashitm thinks about the lib side as well.

lib/model/src/main/java/io/atlasmap/v2/Document.java
```diff
@@ -54,4 +59,62 @@ public void setFields(Fields value) {
...
     }
 
+    /**
+     * Exclude fields which are not on requested paths.
+     *
+     * If includePaths is null or empty it does nothing and all fields are included.
+     *
+     * @param includePaths
+     */
+    public void excludeNotRequestedFields(List<String> includePaths) {
```
igarashitm, line 69: This is better to be in atlas-core, maybe as a static util method or something (e.g. io.atlasmap.core.InspectionUtil next to AtlasUtil). The model objects are better to be only with properties getter/setter and minimize the logic.

lib/model/src/main/java/io/atlasmap/v2/Field.java
```diff
@@ -437,4 +439,15 @@ public String getName() {
...
     }
+
+
+    /**
+     * A shortcut to get fields of a collection as a list.
+     *
+     * @return list of fields
+     */
+    @JsonIgnore
+    public List<? extends Field> getCollectionFields() {
```
igarashitm, line 450: What kind of job will `getCollectionFields()` do? currently it looks like only returning children in the complex field?

lib/modules/json/model/src/main/java/io/atlasmap/json/v2/JsonInspectionRequest.java
```diff
@@ -40,150 +42,162 @@
 
     protected InspectionType type;
 
+    protected List<String> inspectPaths;
```
igarashitm, line 45: We can add `BaseInspectionRequest` abstract class (and maybe remove InspectionRequest interface if not needed) and add getter/setter for this there if it's data format agnostic.

lib/model/src/main/java/io/atlasmap/v2/Document.java
igarashitm: Also, I wonder if we can smoothly transition when we make backend side inspection on-demand as a 2nd phase. I expect the complex fields are marked with something like `FieldStatus.DEFERRED` and have no children. Will this method be just removed at the 2nd phase?

lib/model/src/main/java/io/atlasmap/v2/Field.java
igarashitm: FYI, `Collection Field` in AtlasMap is a special term for the array/list field, like `/employee<>/email<>`

rkorytkowski: Yes, it's a convenient method to determine if the field has any children. Otherwise I'd need to check for an implementation, cast and get children. It allows to abstract that out.

rkorytkowski: Maybe rename it to getChildrenFields() then?

rkorytkowski:
> This looks pretty good to me. A lot of async layering but it probably needed that anyway. If you can get this to pass the tests completely then I'd say your close to pushing it. See what @igarashitm thinks about the lib side as well.

Thanks! Hopefully the failing UI test is now fixed. The issue with not loading mappings properly is also resolved so it's all functional. I'll address other items that @igarashitm has spotted now.

lib/model/src/main/java/io/atlasmap/v2/Field.java
igarashitm: Hmm OK, that's a downside of having complex field classes (*ComplexType) separated from the primitive fields (*Field).  Although we might be able to clean them up in next major version, I'd like to avoid returning null for this from the primitive fields as its intention is not clear at a glance, and it looks confusing as we already have complex field types. Can we also move it as a static utility method?

lib/modules/json/model/src/main/java/io/atlasmap/json/v2/JsonField.java
```diff
@@ -36,7 +36,7 @@
      * Gets the value of the primitive property.
      *
      * @return
-     *     possible object is
+     *     possible object isget
```
igarashitm, line 39: typo?

lib/model/src/main/java/io/atlasmap/v2/Document.java
igarashitm: And can we get a unit test once it's relocated?

lib/modules/json/model/src/main/java/io/atlasmap/json/v2/JsonInspectionRequest.java
rkorytkowski: Fixed!

lib/model/src/main/java/io/atlasmap/v2/Field.java
rkorytkowski: I made getChildrenFields return empty list by default. Not sure about the static utility method. I would still need to cast to different types in that method and make it aware of all implementations. Where should I put it?

lib/model/src/main/java/io/atlasmap/v2/Document.java
rkorytkowski: I've moved it to AtlasUtil and unit tested it. I didn't assign FieldStatus.DEFERRED. It would require rewriting the method as I would need to mark parent as deferred and currently I don't pass parent around... just the collection of children. If you think it's important to mark parent as FieldStatus.DEFERRED then I can try to do that.

lib/model/src/main/java/io/atlasmap/v2/Field.java
igarashitm: Maybe to AtlasUtil just like `excludeNotRequestedFields()`? with same reason, keep model objects from having a logic other than getter/setter.
We can avoid casting to the module specific complex type like `JsonComplexType` by adding an interface, e.g. `ComplexType` which only declares the getter/setter for the children. Does it help?

rkorytkowski: If we are to add an interface to ComplexTypes then there's no need for a util method as it's simply a getter. I'm okay with doing that instead of adding it to the Field.

igarashitm: Sounds good :+1:

rkorytkowski: Done!

lib/model/src/main/java/io/atlasmap/v2/Document.java
rkorytkowski: @igarashitm how about this one? Is it okay to not set FieldStatus.DEFERRED? It's a bit more work and UI doesn't need that. UI just checks if there are children loaded for a complex type and if not then it requests them.

igarashitm: Sounds good, FieldStatus.DEFERRED is not necessary if it doesn't make something better.

igarashitm:
@rkorytkowski did you see the parent complex fields are now expanded by default in the stories? You can see it from the Chromatic "UI Review" GitHub check
https://www.chromatic.com/pullrequest?appId=60c26c6784efc9003bf6c3bb&number=3604

Maybe just because the children are already filled in the static stories, but wanted to make sure.

igarashitm (approved):

Issue #3583, opened 2021-12-15
[BUG] - Correct field node offsets when deleting OOB and UD function references.


pleacu:
**Describe the bug**
If a user has a conditional expression fragment like this:

`   lowercase(state)`

and then decides to delete the `lowercase` portion then the `state` field node reference must have its position adjusted.  We can identify the function references internally as they are preceded by `action.` when parsed.  We can add the same semantics as we have with field refs where if you delete any portion of the name it deletes the whole thing automatically or just have it delete one char at a time.  This must also not break deserialization.

Use mockdoc.adm

pleacu:
The real issue here was that bs/ del didn't inhibit the monaco editor standard processing via `stopPropagation`/`preventDefault`.

PR #3642, opened 2022-01-06
fix: Correct field node offsets when deleting OOB and UD function ref…

pleacu:
…erences.  Also correct auto '+' insertion.

Fixes: #3583


pleacu (commented):
Changes to expression support only

Issue #3234, opened 2021-08-23
[RFE] UI: Support importing kafka schema


igarashitm:
The 2nd option in https://github.com/atlasmap/atlasmap/discussions/3231#discussioncomment-1224009

Once #3233 is implemented, add an option to import a kafka schema as a Document in UI.

kafka-connect module at backend side has been implemented here:
https://github.com/atlasmap/atlasmap/tree/main/lib/modules/kafka-connect

Currently only JSON and AVRO are the supported Kafka Connect schema type.
https://github.com/atlasmap/atlasmap/blob/main/lib/modules/kafka-connect/service/src/main/java/io/atlasmap/kafkaconnect/service/KafkaConnectService.java#L117-L122

Note that Kafka Connect JSON schema is different from JSON schema.

There're a few examples of Kafka Connect JSON/AVRO schema here:
https://github.com/atlasmap/atlasmap/tree/main/lib/modules/kafka-connect/test-model/src/main/resources

Since Kafka Connect requires to put the schema along the data, Document metadata is required to run mapping for Kafka Connect Document. It expects `DataSourceMetadata.specification` to hold schema payload
https://github.com/atlasmap/atlasmap/blob/main/lib/modules/kafka-connect/module/src/main/java/io/atlasmap/kafkaconnect/module/KafkaConnectModule.java#L106
and the other inspection parameters to be in `DataSourceMetadata.inspectionParameters`
https://github.com/atlasmap/atlasmap/blob/main/lib/modules/kafka-connect/module/src/main/java/io/atlasmap/kafkaconnect/module/KafkaConnectModule.java#L108-L110
The required parameters are:
* schemaType - JSON or AVRO
* isKey - boolean flag, true for key, false for value

pleacu:
@igarashitm - so do you see user selecting the existing import document icon and then being presented with a select menu:
<pre>
* Import Kafka Aware Schema
* Import Standard Instance/ Schema
</pre>

Or do you see two separate icons on the panel - one for Kakka and one for standard?

pleacu:
Does the Kafka Connect AVRO schema have a specific suffix?

igarashitm:
Or Java class even could be migrated as an option if we can introduce dynamic form, like when Java class is selected, file selector is replaced with a textbox to enter a class name.

igarashitm:
> Does the Kafka Connect AVRO schema have a specific suffix?

it's `.json`, even the structure is very similar to Kafka Connect JSON

[...]

PR #3725, opened 2022-02-04
fix: UI: Support importing kafka connect schema

pleacu:
Fixes: #3234


lib/modules/kafka-connect/module/src/main/java/io/atlasmap/kafkaconnect/module/KafkaConnectValidationService.java
```diff
@@ -77,10 +77,12 @@ protected AtlasModuleDetail getModuleDetail() {
     @Override
     protected void validateModuleField(String mappingId, KafkaConnectField field, FieldDirection direction,
             List<Validation> validations) {
-        validatorMap.get("kafkaconnect.field.type.not.null").validate(field, validations, mappingId, ValidationStatus.WARN);
-
-        if (field.getPath() == null) {
-            validatorMap.get("kafkaconnect.field.path.not.null").validate(field.getPath(), validations, mappingId);
+        AtlasValidator aVal = validatorMap.get("kafkaconnect.field.type.not.null");
```
igarashitm, line 80: I guess the real fix is on L62-L63, replace `json.field.type.not.null` to `kafkaconnect.field.type.not.null` and like that?

ui/packages/atlasmap-core/src/contracts/common.ts
```diff
@@ -83,12 +83,19 @@ export enum FieldType {
 }
 
 export enum DocumentType {
+  CORE = 'Core',
+  CSV = 'CSV',
   JAVA = 'JAVA',
+  JSON = 'JSON',
+  JSON_INSTANCE = 'JSON_INSTANCE',
```
igarashitm, line 90: Are the `SCHENA` and `INSTANCE` are required for something? We have `InspectionType` enum apart from this `DocumentType` at runtime.

ui/packages/atlasmap/src/Views/ColumnMapperView/Actions/ImportAction.tsx
```diff
@@ -13,20 +13,72 @@
...
+  onImport: (selectedFile: File, docType: DocumentType) => void;
 }
-export const ImportAction: FunctionComponent<IImportActionProps> = ({
-  id,
-  onImport,
-}) => {
+
+type importDocumentType = {
+  documentType: DocumentType;
+  typeName: string;
```
igarashitm, line 35: Can you add `inspectionType` here? in that way `DocumentType` would be able to stay like before, isolated from schema/instance. Otherwise we need to remove `inspectionType` completely from both of UI and runtime to avoid further confusion.

igarashitm: Considering that the Syndesis is using atlasmap-maven-plugin to perform offline inspection, removing `InspectionType` from runtime is more risky. So we'd be better to keep it.

pleacu:
resubmitted - see https://github.com/atlasmap/atlasmap/pull/3746

PR #3745, opened 2022-02-07
fix: replace existing CVS inspection parameters

zregvart:
When changing the CVS parameters the document URI accumulates previously
set parameters. This leads to unwanted parameters remaining, or more
than one value for a parameter being used.

This makes sure that when setting the CVS inspection parameters the
existing parameters are first removed. This way only the desired new
parameters are used.


igarashitm (approved):
Thank you for the fix :+1:

PR #3756, opened 2022-02-10
fix: notify of mapping changes in initialization

### LINKED ISSUE #3755: [BUG] Document URI containing previous parameters
**Describe the bug**

The document URI of a CVS document can end up having previous version if the CSV inspection parameters change and no other change is performed on the mapping.

**To Reproduce**

I've experienced this in Syndesis.

1. Create a mapping from a CVS data shape to a (for example) JSON data shape
2. Change the CVS data shape's parameters
3. Enter the mapping editor
4. Save the mapping without changing the mapping

**Expected behavior**

Even though mapping wasn't altered, the changes made in the CVS parameters should cause mapping to be changed.

zregvart:
When initializing with a preloaded mapping JSON, if the mapping JSON
changed, say because the inspection parameters that have changed and
this is reflected in the document URI, we need to notify that the
mapping changed.
A scenario in which this issue is encountered is when CSV parameters
change and no other change is performed. The saved mapping in that case
contains the old URI in the CVS document.

Fixes #3755

### DIFF
--- a/ui/packages/atlasmap-core/src/services/initialization.service.spec.ts
+++ b/ui/packages/atlasmap-core/src/services/initialization.service.spec.ts
@@ -29,6 +29,7 @@ import atlasmapInspectionMockXmlInstance1Json from '../../../../test-resources/i
 import atlasmapInspectionMockXmlSchema1Json from '../../../../test-resources/inspected/atlasmap-inspection-mock-xml-schema-1.json';
 import atlasmapInspectionOldActionSourceJson from '../../../../test-resources/inspected/atlasmap-inspection-old-action-source.json';
 import atlasmapInspectionOldActionTargetJson from '../../../../test-resources/inspected/atlasmap-inspection-old-action-target.json';
+import atlasmappingCvsToJson from '../../../../test-resources/mapping/atlasmapping-cvs-to-json.json';
 import atlasmappingOldActionJson from '../../../../test-resources/mapping/atlasmapping-old-action.json';
 import ky from 'ky/umd';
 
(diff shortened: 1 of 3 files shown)


zregvart:
After struggling a bit trying to `yarn link` between Syndesis and Atlasmap (I have this change on 2.3.x branch as well), and finally giving up and just doing a filesystem symbolic link of the `atlasmap-core/dist` in Syndesis's `node_modules/...` I've now found that this causes issues and needs to be rethought. An exception is raised at runtime:

```
TypeError: mappingDefinition is null
serializeMappings
src/utils/mapping-serializer.ts:96

  93 | // TODO: check this non null operator
  94 | const mappingDefinition: MappingDefinition = cfg.mappings!;
  95 | let jsonMappings: IMapping[] = [];
> 96 | for (const mapping of mappingDefinition.mappings.filter((m) =>
     | ^  97 |   m.isFullyMapped()
  98 | )) {
  99 |   try {
```

```
next
src/services/initialization.service.ts:152

  149 | this.cfg.documentService.inspectDocuments().subscribe({
  150 |   next: () => {
  151 |     if (this.cfg.preloadedMappingJson) {
> 152 |       const maybeUpdatedMappingJson = JSON.stringify(
      | ^  153 |         MappingSerializer.serializeMappings(this.cfg)
  154 |       );
  155 |       // inspection might change the mapping, for example the preloaded
```

I'll try to find a better solution.

igarashitm:
IIRC you can specify the local directory like `.../atlasmap/ui/packages/atlasmap/dist/` as a dependency instead of version number, that should resolve it directly from build artifacts, bypassing node_modules... this one? https://stackoverflow.com/questions/14381898/local-dependency-in-package-json

zregvart:
This version seems to do the trick, the issue was that in order for the mappings to be set on `cfg`, `updateStatus` needed to finish, so this now waits for that to complete.

igarashitm (approved):

PR #3780, opened 2022-02-14
should make @atlasmap/atlasmap consistent with @atlasmap/core version？

polly-zou:
#3783

### DIFF
--- a/ui/packages/atlasmap/package.json
+++ b/ui/packages/atlasmap/package.json
@@ -26,7 +26,7 @@
     "storybook": "start-storybook -p 6006"
   },
   "dependencies": {
-    "@atlasmap/core": "^2.4.0-SNAPSHOT"
+    "@atlasmap/core": "2.4.0-SNAPSHOT"
   },
   "devDependencies": {
     "@babel/core": "^7.12.10",


igarashitm:
What's the problem that needs this fix?

polly-zou:
> What's the problem that needs this fix?

In my project, @atlasmap/core version was automatically updated from 2.4.0-M.3 to 2.4.0-M.4, but the back-end version was not updated, only the front-end ui version was updated, for example, @atlasmap/core added inspectPaths in version 2.4.0-M.4, but the back-end was not updated because version can't receive this parameter properly, so the page can't be opened

Or is there any way to allow users to specify the version of @atlasmap/core in the host environment?

igarashitm:
I think you can explicitly specify the transitive dependency but I'm OK doing it here - can you file an issue about it - would be a "Task" - and fix the commit message?

polly-zou:
> I think you can explicitly specify the transitive dependency but I'm OK doing it here - can you file an issue about it - would be a "Task" - and fix the commit message? https://github.com/atlasmap/atlasmap/wiki/AtlasMap-Developer-Memo#commit-message

the issue and commit message have done :satisfied: ; coule you tell me how explicitly specify the transitive dependency,and I tried many ways to no avail :disappointed_relieved: Many thanks

igarashitm (approved):

Issue #3781, opened 2022-02-15
[BUG] JSON schema import fails


igarashitm:
**Describe the bug**

JSON schema import as a Document silently fails, even the dropdown stays. Sometimes it succeeds, could be some timing issue?

**To Reproduce**

1. Import JSON schema file as a source Document

**Screenshots**

![atlasmap](https://user-images.githubusercontent.com/265462/154126689-484c8038-68b8-4bf2-b06c-7b8339d17728.gif)

igarashitm:
It worked once I reload the whole page (F5)

pleacu:
I went to my local `atlasmap/standalone` and `mvn clean spring-boot:run` followed by `http://localhost:8585/` in a separate browser window in various sequences and can not get the schema source to fail to load (still using the fix here). I also tried the fully standalone `java -jar ~/git-clone/atlasmap/standalone/target/atlasmap-standalone.ja` as well as the standard storybook atlasmap instance - also unable to get it to fail for me. Perhaps a browser instance issue? `Version 90.0.4430.212 (Official Build) (64-bit)`

igarashitm:
Same thing happens with XML instance... now it only succeeds like once in 10 attempts. I'm with `98.0.4758.102 (Official Build) (64-bit)`

pleacu:
Hey - good news and bad news - I updated Chrome to your version and it fails for me now too!

PR #3782, opened 2022-02-15
fix: UI: Correct useEffect() re-rendering during file import.

pleacu:
Fixes: #3781

@igarashitm - please try your import with this change.  I could not reproduce your situation but corrected useEffect rerendering triggering which should catch your issue.

### DIFF
--- a/ui/packages/atlasmap/src/Views/ColumnMapperView/Actions/ImportAction.tsx
+++ b/ui/packages/atlasmap/src/Views/ColumnMapperView/Actions/ImportAction.tsx
@@ -102,14 +102,15 @@ export const ImportDocumentTypeItem: FunctionComponent<{
 
   useEffect(() => {
     if (previouslyUploadedFiles.current !== files) {
-      previouslyUploadedFiles.current = files;
       if (files?.length === 1) {
-        previouslyUploadedFiles.current = null;
         onFile(
           files[0],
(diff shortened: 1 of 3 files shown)


igarashitm:
Seems it contains a leftover from dependabot PRs in yarn.lock, let me clean it up first.

igarashitm:
Nope it doesn't work - even reload didn't help this time.

igarashitm:
Hmm, it's inconsistent... after it reproduced, I closed the browser tab, restarted the stasndalone with `mvn clean spring-boot:run`, then opened a new tab, then it worked. Then yet again, I closed the browser tab, restarted the standalone, opened a new tab, then the issue reproduced again.

pleacu:
Will close this PR for now since it seems to have no effect (but does appear to be more efficient).  I went to my local `atlasmap/standalone` and `mvn clean spring-boot:run` followed by `http://localhost:8585/` in a separate browser window in various sequences and can not get the schema source to fail to load (still using the fix here).  I also tried the fully standalone `java -jar ~/git-clone/atlasmap/standalone/target/atlasmap-standalone.jar` as well as the standard storybook atlasmap instance - also unable to get it to fail for me.  Perhaps a browser instance issue? `Version 90.0.4430.212 (Official Build) (64-bit)`

Issue #3138, opened 2021-07-22
[TASK] Consider building expression framework on language server protocol


igarashitm:
Current expression editor is a quick hack from PoC and won't scale. UI has some parsing logic inside, which is not ideal. We should rebuild it probably on lauguage server protocol so that we can push all parsing logic out to the backend.

We should finish this research before #982, #1139, #2093

Issue #603, opened 2018-10-03
Implement server side field search


rkorytkowski:
Today field search is calculated at client side, i.e. it’s JavaScript running in browser. It just iterates all the Documents/fields and set visible flag to `true` if name matches. Thus field search won't work at all once inspection services are refactored to on-demand style (#579), as then client side no longer have whole Document.

igarashitm:
Hi @rkorytkowski , can you elaborate it a bit? what would you throw as an query and what is expected to be returned? And maybe how is it leveraged from UI in combination with inspection service?

igarashitm:
Ah now I reminded the `Search` feature in the UI. Right, that field search requires to ask server side once #579 is implemented. It would be a separate API to return a list of field paths that matches with query. If user select one of it which is not yet expanded, on-demand inspection should run underneath to get that chunk.

igarashitm:
Now that we merged UI side on-demand inspection, field search doesn't work anymore until all parents are expanded. Changing to a bug.

rkorytkowski:
There's a quick fix for that i.e. request the whole document before running the query (passing no includePaths to inspect API). However, I didn't go that route since my finding was that even before the dynamic tree, the tree needed to be entirely expanded for the search to work. Otherwise it would not show search results for closed nodes.

Anyway, it should be straightforward to implement an endpoint, which only return nodes matching the query. I could help with it.

igarashitm:
While we need to keep backward compatibility - i.e. importing old .adm file only contains adm-catalog-files.gz, creates the new DataSourceCatalog with inspected JSON files, then after that .adm export would contain those new metadata.

igarashitm:
reopening, not yet completed

PR #3814, opened 2022-03-02
feat: LSP server for conditional mapping expression (#3138)

igarashitm:
Checkpoint: Established LSP server websocket endpoint and mocked up the server impl. E2ETest tests the websocket connection and initialization.

TODO:
* Implement LSP server logic for the conditional mapping expression syntax support
  * Establish function metadata registry so that the LSP function completion can leverage it
  * Integrate with server side field search (#603) for the field reference in expression
* Connect from UI, probably with using https://github.com/TypeFox/monaco-languageclient

### DIFF
--- a/lib/service/pom.xml
+++ b/lib/service/pom.xml
@@ -65,16 +65,31 @@
       <artifactId>swagger-annotations</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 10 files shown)


pleacu (approved):
+1 : seems straightforward

PR #3830, opened 2022-03-10
fix: Implement server side field search

rkorytkowski:
Fixes: #603

It implements the API side by introducing searchPhrase into InspectionRequest sent to /inspect. 
It should be straightforward to modify UI search to first call the /inspect with the searchPhrase provided and then do a usual client side search to update the tree.


rkorytkowski:
@igarashitm I managed to fix the algorithm to support searchPhrase by the /inspect endpoint as we discussed. It is mergeable as it doesn't break anything. I'll try to look into modifying UI to call /inspect when entering search text.

igarashitm (approved):

Issue #3510, opened 2021-11-23
Showing not available routes (fields) in expression dropdown[BUG] 


YevhenTarashchyk:
Imagine the mapper has 2 source documents: doc1 and doc2, and they both have similar structure or fields in common. Like this 
<img width="653" alt="Screen Shot 2021-11-23 at 14 56 30" src="https://user-images.githubusercontent.com/29583265/143028025-e42631b2-dc13-4b91-964a-29646f1134fd.png">
As u can see from dropdown we have fields (routes) from 1 of them 

https://user-images.githubusercontent.com/29583265/143028673-3b6b97a3-9dd3-47a7-b6f8-8e213edc56d7.mp4

igarashitm:
Please follow the issue template and provide all infos so we can reproduce. Also you'd want to try with main branch as we migrated the expression input box to monaco-editor.

pleacu:
As @igarashitm said - we've migrated away from a primitive HTML editor to the monaco editor and we not longer use `@` to insert field references into the conditional expression - we use a pull-down menu (ctrl-enter or mouse-3).

pleacu:
I just looked at this again.  The issue here is that the number of elements listed in the search window is limited to 20.  The other issue is that there was no indication that more field elements were available but were past the 20 element threshold.  The last thing is that the number of elements selectable is dependent on the fields that are visible in the source/target panels (i.e. expanded).  So - I've got a fix as follows:

1.  I've increased the number of elements in the search window to 500.  Again - the searchable fields are based on the visible fields in the panel (expanded).  Mush past 500 and the display becomes sluggish.
2.  I've added a message in the number of searchable fields exceeds 500:
   `** Displayable field limit reached **`

ref:
![sc1](https://user-images.githubusercontent.com/2981421/157966954-15ce7e9f-b42f-4fb4-8122-d8b7c0e6f14b.png)

@igarashitm - I've got a PR ready if this sounds ok.

pleacu:
This is fixed on the atlasmap-2.3.x branch.

PR #3849, opened 2022-03-14
Atlasmap 2.3.x

pleacu:
fix: UI: Extend expr search element max.
Fixes: #3510


pleacu:
*ugh* - I hate github...

Issue #3812, opened 2022-03-01
[RFE] Version Compatibility Check


igarashitm:
**Is your feature request related to a problem? Please describe.**

Sometimes user uses the newer version of AtlasMap UI (maybe through vscode extension) and let the older version of AtlasMap runtime comsume it, which often causes a conflict. We should embed a version number into Mapping Definition and let runtime check the compatibility. The rule of thumb is it doesn't work if major.minor version is different, i.e. the .adm created with v2.3 doesn't work with v2.2 runtime. Note that we try to keep backward compatiblity for minor version, so the .adm created with v2.3 should work with v2.4 runtime.

pleacu:
@igarashitm, @lhein 

   On AtlasMap startup:
   - UI and RT version is the same:
     Ping/ pong messages to INFO.  Messages will contain the versions just as a visual verification.
ref:
<pre>
2022-03-15 15:14:32.635 DEBUG 1590558 --- [  XNIO-1 task-2] io.undertow.request.security             : Authentication result was ATTEMPTED for /v2/atlas/ping
2022-03-15 15:14:32.664 DEBUG 1590558 --- [  XNIO-1 task-2] io.atlasmap.service.AtlasService         : Ping...  responding with 'pong'.
</pre>

   - UI version > RT version
     UI Error dialog - `Incompatible Atlasmap runtime (2.13.12) detected (2.13.17 expected) at port 8585.`
     RT log error stating the same
    It's important to reference the port in case they have multiple instances of the AtlasMap RT running.
   - UI version < RT version
     INFO - `Atlasmap runtime (2.13.12) detected (2.13.11 expected) at port 8585.`
     This will keep things working if VSCode AtlasMap UI is a little older.
 
   ADM version field on init (JSON deserialization):
   - `undefined` works with RT and UI.  UI will update ADM with the UI version field (e.g. '2.3.17')
   - '2.3.12'
     UI or RT version >= 2.3.12 will work (newer UI/RT can handle older ADM)
     UI or RT version < 2.3.12 will fail (older UI/RT can't handle future ADM)

Ref: 
Will add `version?: string;`
https://github.com/atlasmap/atlasmap/blob/44745c3b1e3d8b9250b6ceb3cb4602a97a841b59/ui/packages/atlasmap-core/src/contracts/mapping.ts#L44-L56

Will add `version: string | null = null;`
[...]

PR #3862, opened 2022-03-23
fix: Add version info to mapping - check in UI and in default Atlas c…

pleacu:
…ontext.

Fixes: #3812

Fixes: #

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
+++ b/lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
@@ -44,6 +44,7 @@ import io.atlasmap.api.AtlasConversionException;
 import io.atlasmap.api.AtlasException;
 import io.atlasmap.api.AtlasSession;
 import io.atlasmap.mxbean.AtlasContextMXBean;
+import io.atlasmap.spi.AtlasInternalSession;
 import io.atlasmap.spi.AtlasModule;
 import io.atlasmap.spi.AtlasModuleInfo;
 import io.atlasmap.spi.AtlasModuleInfoRegistry;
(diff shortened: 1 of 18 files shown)


lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
```diff
@@ -167,7 +191,18 @@ protected synchronized void init() throws AtlasException {
                     , this.atlasMappingUri);
             return;
         }
-
+        AtlasMapping atlasMapping = this.admHandler.getMappingDefinition();
+        String version = factory.getProperties().get(AtlasContextFactory.PROPERTY_ATLASMAP_CORE_VERSION);
+        String mappingVersion = atlasMapping.getVersion();
+        if (!validateVersion(version, mappingVersion)) {
+            throw new AtlasException(
```
igarashitm, line 198: Can we only do `LOG.error(msg, e)` but not throwing it? maybe with saying like `The version "n.n.n" of Mapping Definition was detected. It might not work as expected with the runtime version "m.m.m"`

ui/packages/atlasmap-core/src/utils/mapping-serializer.ts
```diff
@@ -292,6 +293,30 @@ export class MappingSerializer {
...
+      cfg.errorService.addError(
+        new ErrorInfo({
+          message: `Mappings file version mismatch.  Expected ${currentVersion}, detected ${mappingVersion}`,
+          level: ErrorLevel.ERROR,
+          scope: ErrorScope.APPLICATION,
+          type: ErrorType.USER,
+        })
+      );
+    } else {
+      cfg.mappings!.version = currentVersion;
```
igarashitm, line 316: I think we can update the version even if the version mismatch is detected. Then the error is shown only at the first time, but if the deserialization succeeds, the version is overridden and the next adm export will create a compatible one. Let user decide if they continue with it.

lib/core/src/main/java/io/atlasmap/core/DefaultAtlasContext.java
igarashitm: And can we add some test cases, would be in DefaultAtlasContextTest

lib/core/src/test/java/io/atlasmap/core/DefaultAtlasContextTest.java
```diff
@@ -61,6 +61,31 @@
 
 public class DefaultAtlasContextTest extends BaseDefaultAtlasContextTest {
 
+    @Test
+    public void testMappingsVersion() throws AtlasException {
```
igarashitm, line 65: I suspect this doesn't trigger version check? would need to invoke `context.init()`...
or invoke version check also from `context.processValidation()` and put the warn level `Audit`

pleacu: I went the nvoke version check also from context.processValidation() and put the warn level Audit way.

Issue #3750, opened 2022-02-09
[TASK] Refactor to store inspection result into DataSourceMetadata instead of inspection source


igarashitm:
Currently we store all the inspection seeds into `DataSourceMetadata` in the .adm file, and the inspections are performed everytime when it's loaded into UI. We should instead inpect only when it's imported as a Document and on update once #3553 is implemented, and store the inspection result JSON into `DataSourceMetadata`. This way even if the inspection seed is a huge repeating object like discussed in https://github.com/atlasmap/atlasmap/issues/3392, we use it only once on inspection and the stored inspection result won't contain unnecesary repeating items.
#1476 has to come before this one or be done together.

Issue #1577, opened 2019-12-09
UI_REACT: migrate multiple mapping definition routing


igarashitm:
Port #930 to ui-react
https://github.com/atlasmap/atlasmap/tree/master/ui-react

igarashitm:
@AbbassFaytaroony are you interested in working on this?

AbbassFaytaroony:
Sure! I have a feature proposal in mind, ill create a separate issue for it, to discuss and decide priorities

riccardo-forina:
This can be tricky. With the Angular UI we used to embed Atlasmap as a standalone application through iframes and webviews. With the React UI in some host environments like Syndesis we don't run in an iframe anymore because Atlasmap is consumed as a React component, so we can't set up a full routing solution automatically because it would clash with what Syndesis uses.

One option is to use the same routing solution Syndesis uses (react-router), coupling Atlasmap to Syndesis a bit because versions must match. For example, there is a newer major version of react-router that Syndesis is not using yet (nor is planning to).

A second option could be to not use url-based routing at all and handle it programmatically. This would ease embedding Atlasmap, but would be harder to code and a downgrade in the user experience.

A third option could be to offer this feature only when using Atlasmap standalone and not when embedded as a React component. Syndesis wouldn't make use of this feature anyhow. But what about other potential consumers of the app?

igarashitm:
Third is the one, this is a standalone only feature and have to be disabled in Syndesis. We can reconsider about supporting multiple mapping definition for embedding mode once we get any other user than Syndesis.

igarashitm:
Reopening - multiple mapping definition ID is still taken into account at the backend. We'll still need to make an enhancement at UI side. It doesn't necessary to be URL routing, we just need to allow initializing the other instance by specifying mapping definition ID, which is hardcoded to `0` in UI ATM.
Cc: @lhein

pleacu:
So specifically how would a user specify the mapping ID?  If they drag a source field to a target field to establish a new mapping should a dialog pop up asking for the ID?  Should the mapping details pain have an ID field which is editable?

igarashitm:
As a first step, user won't notice that they are using Mapping Definition ID = 0, which is already hardcoded in the UI. It would be the tool like vscode AtlasMap plugin, which specifies other than 0 for the Mapping Definition ID, and manage multiple mapping projects with a single backend instance.

igarashitm:
[...]

PR #3864, opened 2022-03-25
fix: Pre clean up toward #3750 #1476 #603 #3138 #1577

igarashitm:
* Removed mapping definition from `ADMDigest` which has been broken due to duplicated stringify. It's not neccessary as the separate mapping definition JSON file is always used. `ADMDigest` will be eventually removed through #3750 and #1476. The `ADMArchiveHandler` at backend side would only keep supporting extracting `ADMDigest` for backward compatibility, but only convert it to the newer structure.
 * Adjusted REST API URLs to be a bit more "RESTful", and avoid response body for HTTP PUT/DELETE which is prohibited
 * Created `MappingService` and `DocumentService` to split the `AtlasService` into 3. `DocumentService` will provide the backend Document store for #3750 and #1476
 * Introduced `ModuleService` to interface common operations that core services would delegate to while decoupling module formats
 * Introduced a word `project` to represent Documents+MappingDefinition, which is still identified by Mapping Definition ID. Deleting a project includes deleting correlated Documents as well as a Mapping Definition. #1577 would be then switching between projects with Mapping Definition ID as a key. ~~The library (jar files) is on the other hand global, shared across the projects ATM. Although .adm export is per project. This might have to change in order to leverage #1577 for vscode plugin, maybe let libarary belong to the project.~~ oops not a good idea... it's too cumbersome to switch the custom class loader between projects. The library should stay global.


lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -347,343 +230,138 @@ public Response resetAllMappings() {
...
-            LOG.error("Error saving Mapping Definition file.\n" + e.getMessage(), e);
-            throw new WebApplicationException(e.getMessage(), e, Status.INTERNAL_SERVER_ERROR);
+            getLibraryLoader().addJarFromStream(requestIn);
+        } catch (Exception e) {
+            if (LOG.isDebugEnabled()) {
+                LOG.error("", e);
+            }
+            StringBuilder buf = new StringBuilder();
+            buf.append("Failed to import a jar file. This error occurs when:\n")
+                    .append(("\t1. The jar file is not compatible with the JVM AtlasMap backend server is running on\n"))
```
pleacu, line 270: The jar file is not compatible with the JVM AtlasMap backend server is running on is not a correct sentence.  I'd just shorted in to:
The jar file is not compatible with the JVM AtlasMap backend server

ui/packages/atlasmap-core/src/services/file-management.service.ts
```diff
@@ -160,25 +160,23 @@ export class FileManagementService {
...
 
   /**
-   * Delete mapping files on the runtime.
+   * Delete mapping project on the runtime.
    */
   resetMappings(): Promise<boolean> {
     return new Promise<boolean>((resolve) => {
-      const url = this.cfg.initCfg.baseMappingServiceUrl + 'mapping/RESET';
-      this.cfg.logger!.debug('Reset Mappings Request');
+      const url = this.cfg.initCfg.baseAtlasServiceUrl + 'project/0';
```
pleacu, line 167: Is reset really always delete in this context?

pleacu (commented):
I  like the into of moduleService - also the split between AtlasService and MappingService.  Looks good - just that one text change and its +1 for me

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
igarashitm: added "which", `the JVM which AtlasMap backend server is running on`

ui/packages/atlasmap-core/src/services/file-management.service.ts
igarashitm: It looks like so, the backend just deletes the directory

pleacu (approved):
+1 - looks good

Issue #3865, opened 2022-03-25
[RFE] Add "Delete mappings"


igarashitm:
**Describe the solution you'd like**

Add "Delete mappings" menu in the UI. Currently we only have "Reset All", which removes all mapping definitions, Documents as well as libraries. If the user want to start creating mappings from start, but with the same libraries and Documents, "Delete mappings" is more helpful.
The REST target would be
`DELETE /v2/atlas/project/{mappingDefinitionId}/mapping`
on `MappingService`

pleacu:
![sc1](https://user-images.githubusercontent.com/2981421/160888893-b932af2b-3b6a-4bf7-8db5-82776e11d904.png)
![sc2](https://user-images.githubusercontent.com/2981421/160888916-8e9b1206-0a01-4fd2-8eb7-a079d4f6d520.png)
![sc3](https://user-images.githubusercontent.com/2981421/160889035-2aff6461-54e2-4ee4-8ebd-1f344bd92860.png)
![sc4](https://user-images.githubusercontent.com/2981421/160889021-a4499fba-304d-41ad-94a7-fd10afdfb415.png)

pleacu:
Hey @igarashitm - having difficulty with the `mapping-management.service.spec.ts`.  For `removeAllMappings` test the build fails with:
<pre>
@atlasmap/core: FAIL src/services/mapping-management.service.spec.ts (9.803s)
@atlasmap/core:   ● MappingManagementService › removeAllMappings()
@atlasmap/core:     TypeError: globals.Request is not a constructor
@atlasmap/core:       171 |         '/mapping';
@atlasmap/core:       172 |       this.api
@atlasmap/core:     > 173 |         .delete(url)
@atlasmap/core:           |                ^
@atlasmap/core:       174 |         .then((res: Response) => {
@atlasmap/core:       175 |           this.cfg.logger!.debug(`Delete Mappings Response: ${res.ok}`);
@atlasmap/core:       176 |           this.notifyMappingUpdated();
@atlasmap/core:       at new Ky (../../node_modules/ky/umd.js:243:19)
@atlasmap/core:       at Function.ky.<computed> [as delete] (../../node_modules/ky/umd.js:456:37)
@atlasmap/core:       at src/services/mapping-management.service.ts:173:16
</pre>
The new `removeAllMappings`  function looks like this:
<pre>
  removeAllMappings(): Promise<boolean> {
    return new Promise<boolean>((resolve) => {
[...]

Issue #3866, opened 2022-03-25
[RFE] Add "Delete Documents and mappings"


igarashitm:
**Describe the solution you'd like**

Add "Delete Documents and mappings" menu in the UI. Currently we only have "Reset All", which removes all mapping definitions, Documents as well as libraries. If the user want to start creating mappings and Documents from start, but with the same libraries, "Delete Documents and mappings" is more helpful.
The REST target would be
`DELETE /v2/atlas/project/{mappingDefinitionId}/`
in `AtlasService`

igarashitm:
Yes, `Reset All` deletes everything, if there're multiple mapping projects it's the only way to reset at once, while `Delete Documents and Mappings` is per project operation and lib is not deleted. That's why it expects `mappingDefinitionId` as a path parameter.

PR #3879, opened 2022-03-30
fix: UI: Added 'Delete Mappings'

pleacu:
Fixes: #3865


igarashitm:
Can you make these invoke backend service as I described in #3865 and #3866?

pleacu:
I thought I'd use existing infrastructure rather than write more core UI code.

igarashitm:
The purpose is to centralize the logic to the backend, then UI only asks backend to manipulate the mappings/documents/libraries. It eventually removes the logic from UI code, even if it adds the REST invocations in short term.

igarashitm:
Actually #3866 might be better to wait for #3750 and #1476 as it involves removing all persisted resources (inspection seed, inspection result, and from data source catalog) correlated to the Document.

pleacu:
Ok - so we only want `Remove Mappings` - I'll leave the UI side infrastructure in place but not visible.  The remove mappings core will `DELETE /v2/atlas/project/{mappingDefinitionId}/mapping` - correct?

igarashitm:
yep and it's already here
https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/MappingService.java#L106
Not sure if this works OOTB, it removes the mapping definition file at file level. But instead, it might need to remove `Mapping` entries from `MappingDefinition` instead, to keep `DataSource`, `Constants` and `Properties`. `LookupTable` would be removed as it's tightly coupled with mappings.

ui/packages/atlasmap-core/src/services/mapping-management.service.ts
```diff
@@ -156,17 +156,33 @@ export class MappingManagementService {
...
+        this.cfg.initCfg.baseAtlasServiceUrl +
+        'project/' +
+        this.cfg.mappingDefinitionId +
+        '/mapping';
+      this.api
+        .delete(url)
+        .then(async (res: Response) => {
+          this.cfg.logger!.debug(`Delete Mappings Response: ${res.ok}`);
+          for (const mapping of this.cfg.mappings!.getAllMappings(true)) {
+            this.cfg.mappings!.removeMapping(mapping);
```
igarashitm, line 173: Could this be instead fetch the mapping definition from backend and refresh? isn't there already a routine doing it?

pleacu: I'd still need to clear the mappings from the mappings model - otherwise it would appear like the mappings are being reintroduced.  I could fetch the mappings after clearing the UI side mappings model just to double check.

igarashitm: yeah only fetching mapping definition isn't enough, then it needs to refresh from what is received - deserialize the MappingDefinition and reflect to the ConfigModel.mappings. I thought we're already doing at somewhere, at least we do on initialiization.

pleacu: Yes - `notifyMappingUpdated()` will cause a full refresh.

igarashitm: The reason is to reduce the logic from UI side and always rely on the backend for data consistency. Eventually every single action on UI should be notified to backend side and UI only asks updated mappings/documents to backend side and display them. We have to reduce those UI side state manipulating one-by-one, that is the primary motivation of the series of refactorings I've been working on recently.

igarashitm: And once we have all the data&logic migrated to the Java side, we can reuse them in a offline tool as well - i.e. without UI, programmatically build an ADM file in Java.

pleacu: Okay - so specifically with this change - its calling the new backend endpoint to clear out just the mappings.  It's then clearing out the UI-side mapping info and refreshing.  Do I add a fetchMappings backend call to verify that the mappings are gone - then do the refresh?

igarashitm: 1. invoke backend to remove mappings
2. fetch updated (i.e. empty) mappings from backend
3. reflect the empty mappings
I thought the deserialize method in UI do #3 automatically? Then #2 and #3 could be reused for any operation thereafter. Does notifyMappingUpdated() already do that?

igarashitm: If the UI side modification is needed, it means that the UI side of change overrides the backend change. Then backend remove mapping invocation is just a waste. Then later time it could be a source of a bug, hard to nail down which logic wins.

pleacu: `notifyMappingUpdated()` serializes, validates and pushes the mappings down to the backend, triggering a UI refresh.  It's necessary to clear out the UI mappings model.  I could extend the change I have to fetch the mappings from the backend (which we know to be empty).  The existing `notifyMappingUpdated()` will reflect the empty mappings back to the UI.  So from your list:
1.  done
2. not done currently but could add but this function's job is to delete all mappings so would be redundant with clearing mapping model
3. done via  `notifyMappingUpdated()`

pleacu: Ultimately the UI needs to tell the backend what the mappings are as defined by the mapping model (created by the user setting mappings).

igarashitm: ah so `notifyMappingUpdated()` is actually to reflect from UI to backend. Then we should eventually replace it with the opposite, {No.2 + No.3} . The {No.2 + No.3} routine will be eventually shared with all the operations, like add mapping entry, add/delete field to/from mapping, etc. You might think it's redundant specifically for delete mappings, but doing same for every operation in the future will reduce the chance of bug significantly.

pleacu: Okay - so changing `notifyMappingUpdated()` from a push to a pull {No.2 + No.3} .  So - for this PR I'd would say it is correct and working with the given semantics and I can create a new issue that will tackle changing `notifyMappingUpdated()` semantics across the UI.  wdyt?

igarashitm: I mean I'm trying to get this refactoring gradually. This time we add {No.2 + No.3} and only delete mappings will use it. Since other operations stay using notifyMappingUpdated(), it's temporary redundant anyway. But eventually we remove notifyMappingUpdated() and rely on {No.2 + No.3} at all time, then redundancy will be also removed.

igarashitm: I'm OK with deferring the refactoring to the next issue, but can you work on it before another UI logic is added? :-D

pleacu: sure - it probably needs some more thought than a quick change anyway.

igarashitm: OK then let this one goes in as-is.
Note that changing `notifyMappingUpdated()` from push to pull at once will be too big to do it at once, rather it'd be better to
1. Add pull routine
2. Let "remove mappings" use the pull routine and verify UI state modification is not necessary
3. Continue for other operations... many of existing mapping manipulation operations will need to migrate the logic from UI to the backend
4. Finally remove `notifyMappingUpdated()`

igarashitm (approved):

ui/packages/atlasmap-core/src/services/mapping-management.service.ts
pleacu: ok - will create a new issue.

Issue #3905, opened 2022-04-05
[TASK] - Refactor removeAllMappings() to use Backend service.


pleacu:
Refactor `removeAllMappings()` to use a new UI core mapping service pull routine and to use the backend service to retain the current mappings.

pleacu:
@igarashitm - Hey Tomo - so here's my issue.  The backend service function `removeMappingRequest` does its job correctly.  However, when I perform a `getMappingRequest` call I get the original mappings because it calls `ADMArchiveHandler admHandler = this.atlasService.loadExplodedMappingDirectory(mappingDefinitionId);` - ref https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/MappingService.java#L135
It appears I need to modify the contents of the explodedMappingsDirectory to remove just the mappings so subsequent `get` mappings will not find the old info.  wdyt?

igarashitm:
That sounds like a bug in backend, this `persist()` should persist to the mapping definition file held in explodedMappingsDirectory - https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/MappingService.java#L111

igarashitm:
Yes It has to invoke `ADMArchiveHandler#setMappingDefinition()` explicitly before `persist()` to clear the cache of serialized one.
https://github.com/atlasmap/atlasmap/blob/main/lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java#L278

PR #3906, opened 2022-04-05
fix: UI: Refactored removeAllMappings()

pleacu:
Fixes: #3905

### DIFF
--- a/ui/packages/atlasmap-core/src/services/mapping-management.service.ts
+++ b/ui/packages/atlasmap-core/src/services/mapping-management.service.ts
@@ -160,12 +160,13 @@ export class MappingManagementService {
       const url: string =
         this.cfg.initCfg.baseAtlasServiceUrl +
         'project/' +
-        this.cfg.mappingDefinitionId;
+        this.cfg.mappingDefinitionId +
+        '/mapping';
       this.api
         .delete(url)
         .then(async (res: Response) => {
           this.cfg.logger!.debug(`Delete Mappings Response: ${res.ok}`);
-          // Re-load the mappings from the backend service.
+          // Re-load the empty mappings from the backend service.
           await this.notifyFetchMapping();
           if (this.cfg.mappings!.mappings.length > 0) {
             this.cfg.errorService.addError(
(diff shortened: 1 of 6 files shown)


ui/packages/atlasmap-core/src/services/mapping-management.service.ts
```diff
@@ -156,21 +153,30 @@ export class MappingManagementService {
...
    */
   removeAllMappings(): Promise<boolean> {
-    return new Promise<boolean>((resolve) => {
+    return new Promise<boolean>(async (resolve) => {
       const url: string =
         this.cfg.initCfg.baseAtlasServiceUrl +
         'project/' +
-        this.cfg.mappingDefinitionId +
-        '/mapping';
+        this.cfg.mappingDefinitionId;
```
igarashitm, line 163: Missing `/mapping`? the REST `DELETE /project/{mappingDefinitionId}` would go this one - https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/AtlasService.java#L392
which removes a mapping project, including Documents. In order to remove only mapping entries but keep the Documents, it needs to go to `DELETE porject/{mappingDefinitionId}/mapping` which is here - https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/MappingService.java#L106

pleacu: Ah - the `DELETE project/{mappingDefinitionId}/mapping` works - its the `fetchMappings()` in the UI that was getting the mappings based on the original JSON mappings file.  So - I assume I need to do a
`GET project/{mappingDefinitionId}/mapping` ?

igarashitm: Yes that's the one to retrieve the mapping - https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/MappingService.java#L132

pleacu: Quick tweek to backend service to `setMappingDefinition()` did the trick.

igarashitm (approved):

PR #3937, opened 2022-04-12
fix: path to react UI in README

kahboom:
There is no `ui` directory in `${ATLASMAP}/ui/packages/atlasmap`--I'm assuming this should be `src`, right?


kahboom:
will reopen once I've sorted this out :)

Issue #3938, opened 2022-04-12
[BUG] yarn build fails without prior maven build


igarashitm:
**Describe the bug**

yarn build for UI fails without prior maven build
```
 FAIL  src/services/document-management.service.spec.ts
  ● Test suite failed to run

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    src/services/document-management.service.spec.ts:29:55 - error TS2307: Cannot find module '../../../../test-resources/inspected/atlasmap-inspection-complex-object-rooted.json' or its corresponding type declarations.

    29 import atlasmapInspectionComplexObjectRootedJson from '../../../../test-resources/inspected/atlasmap-inspection-complex-object-rooted.json';
                                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    src/services/document-management.service.spec.ts:30:51 - error TS2307: Cannot find module '../../../../test-resources/inspected/atlasmap-inspection-po-example-schema.json' or its corresponding type declarations.
```
The files under `ui/test-resources/inspected` are generated through maven plugin
https://github.com/atlasmap/atlasmap/blob/206afff04b2d61db2fa471d3bb779bdc6b602f32/ui/pom.xml#L181

But it's inconvenient for UI developers that simple `yarn build` doesn't work. It might be better to avoid dynamic generation and put the static file for those.

PR #3940, opened 2022-04-12
fix: yarn build fails without prior maven build

igarashitm:
Fixes: #3938
~~Just check those file in so we can detect if there's format change.~~
Had to get the phase hook off as the property order is not fixed


pleacu (approved):
+1

igarashitm:
Had to get the phase hook off as the property order is not fixed

Issue #3935, opened 2022-04-11
[TASK] - Refactor reordering of mapped fields - pull from backend service.


pleacu:
The UI function `handleIndexChange` processes a mapped field index change.  The source field is represented by
 the current index.  The target index is represented either by a number or by a mapped field (depending on whether the index was physically modified or d&d modified).

pleacu:
@igarashitm - having trouble coming up with the simplest way to transfer the target mapping field index to the backend service with the REST API as a request payload.  Do you have an example where a simple integer is passed? Do I need a full headers definition for it in JSON?  I'd like to keep it just as a parameter (`@Parameter`)
<pre>
@Parameter(description = "Target Mapped Field Index") @PathParam("targetFieldIndex") Integer targetFieldIndex
</pre>

igarashitm:
The "RESTful" manner is to use request body but not the URL parameter for other than `GET` method. You should remove both of `@Parameter` and `@PathParam`. Here is an example of sending simple string body.
https://github.com/igarashitm/atlasmap/blob/3750/lib/service/src/main/java/io/atlasmap/service/DocumentService.java#L119-L128
https://github.com/igarashitm/atlasmap/blob/3750/ui/packages/atlasmap-core/src/services/document-management.service.ts#L990

PR #3956, opened 2022-04-22
fix: UI: Refactored handleIndexChange - handling index gaps removing …

pleacu:
…notifyMappingUpdated()

Fixes: #3935

### DIFF
--- a/lib/service/src/main/java/io/atlasmap/service/MappingService.java
+++ b/lib/service/src/main/java/io/atlasmap/service/MappingService.java
@@ -17,6 +17,8 @@ package io.atlasmap.service;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Collections;
+import java.util.List;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
@@ -45,6 +47,9 @@ import io.atlasmap.api.AtlasSession;
 import io.atlasmap.core.ADMArchiveHandler;
 import io.atlasmap.v2.AtlasMapping;
 import io.atlasmap.v2.Audits;
+import io.atlasmap.v2.BaseMapping;
+import io.atlasmap.v2.Field;
+import io.atlasmap.v2.FieldGroup;
 import io.atlasmap.v2.Mapping;
 import io.atlasmap.v2.ProcessMappingRequest;
 import io.atlasmap.v2.ProcessMappingResponse;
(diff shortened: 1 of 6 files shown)


lib/service/src/main/java/io/atlasmap/service/MappingService.java
```diff
@@ -91,6 +96,105 @@ public MappingService(AtlasService parent) {
...
+     *
+     * @param mappingDefinitionId
+     * @param mappingIndex
+     * @param sourceMapping
+     * @param currentFieldIndex
+     * @param targetFieldIndex
+     *
+     * @return empty response
+     */
+    @POST
```
igarashitm, line 148: Maybe `@PUT` as it doesn't have a response body and updating existing resource?

lib/service/src/main/java/io/atlasmap/service/MappingService.java
```diff
@@ -91,6 +96,105 @@ public MappingService(AtlasService parent) {
...
+    @Path("/{mappingIndex}/inptu/{sourceMapping}/current/{currentFieldIndex}/target/{targetFieldIndex}")
```
igarashitm, line 149: I'd suggest `/{mappingIndex}/field/{dataSourceType}/{currentFieldIndex}/index` as an URL, then send the `targetFieldIndex` in the request body. We already have `DataSourceType` enum to indicate if it's `SOURCE` or `TARGET`.

lib/service/src/main/java/io/atlasmap/service/MappingService.java
```diff
@@ -91,6 +97,126 @@ public MappingService(AtlasService parent) {
...
+     * @param mappingDefinitionId
+     * @param mappingId
+     * @param dataSourceType
+     * @param currentFieldIndex
+     * @param targetFieldIndex
+     *
+     * @return empty response
+     */
+    @PUT
+    @Path("/{mappingId}/field/{dataSourceType}/current/{currentFieldIndex}/targetFieldIndex")
```
igarashitm, line 170: Better to be `/{mappingId}/field/{dataSourceType}/{fieldIndex}/index`
URL represents the location of the resource, therefore it shouldn't contain something operation specific. The `current` in the path sounds like there is such property on a mapping field.
For example, if we were to add an operation to get/set the `FieldType` on a mapping field, the URL should be
`/{mappingId}/field/{dataSourceType}/{fieldIndex}/fieldType`
where `GET` method gets the current value and `PUT` updates the `FieldType` while URL is same. Note that only the last part of the path is different from index one. We won't have `GET` for the field index as it's self-evident, but the URL rule stays same.

lib/service/src/main/java/io/atlasmap/service/MappingService.java
```diff
@@ -91,6 +97,126 @@ public MappingService(AtlasService parent) {
...
+    @Produces({MediaType.TEXT_PLAIN})
+    @Operation(summary = "Change Mapped Field Index", description = "Change the mapped field's index value")
+    @ApiResponses(
+        @ApiResponse(responseCode = "200", description = "The mapped field formerly at index {currentFieldIndex} is now at {targetFieldIndex}."))
+    public Response changeMappedFieldIndex(
+        @Parameter(description = "Mapping Definition ID") @PathParam("mappingDefinitionId") Integer mappingDefinitionId,
+        @Parameter(description = "Mapping ID") @PathParam("mappingId") String mappingId,
+        @Parameter(description = "Source or Target Mapping") @PathParam("dataSourceType") String dataSourceType,
```
igarashitm, line 178: `DataSourceType` enum should work for the type of this parameter, to avoid String comparison at L190.

lib/service/src/main/java/io/atlasmap/service/MappingService.java
pleacu: What about `targetFieldIndex` - even though that exists in the payload - doesn't it still need to be specified in the URL?

igarashitm: `/{mappingId}/field/{dataSourceType}/{fieldIndex}/index`
so the `targetFieldIndex` should be instead `index` as a property of the mapping field.

pleacu:
closed in favor of https://github.com/atlasmap/atlasmap/pull/3966

Issue #3968, opened 2022-04-28
[TASK] UI: Refactor removeMappedField() to use backend service.


pleacu:
The mapping model `removeMappedField` function removes a field from an existing mapping.  It then calls `notifyMappingUpdated` in the mapping management service.  Refactor `removeMappedField` and remove  `notifyMappingUpdated`.

PR #3999, opened 2022-05-06
fix: UI: Refactored removeFieldMapping to backend service.  Addressed…

pleacu:
… comments.
Resubmitted PR.
Fixes: #3968

  - Didn't refactor all of the mapping service to atlas mapping handler - left TODOs
  - The MappingServiceTest is really a placeholder for now

### DIFF
--- a/lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
+++ b/lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
@@ -721,21 +721,6 @@ public class ADMArchiveHandler {
     }
 
     /**
-     * Delete the specified field from the specified mapping.
-     *
-     * @param dsType
-     * @param mappingId
-     * @param fieldIndex
-     * @throws AtlasException
-     */
-    public void deleteMappingField(DataSourceType dsType, String mappingId, Integer fieldIndex) throws AtlasException {
-        AtlasMapping def = getMappingDefinition();
-        getAtlasMappingHandler().removeMappingField(def, dsType, mappingId, fieldIndex);
-        setMappingDefinition(def);
-        persist();
-    }
-
-    /**
      * Loads ADM Archive from an exploded directory.
      * @param dir directory path.
      * @throws AtlasException unexpected error
--- a/lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
+++ b/lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
@@ -721,6 +721,21 @@ public class ADMArchiveHandler {
     }
 
     /**
+     * Delete the specified field from the specified mapping.
+     *
+     * @param dsType
+     * @param mappingId
+     * @param fieldIndex
+     * @throws AtlasException
+     */
+    public void deleteMappingField(DataSourceType dsType, String mappingId, Integer fieldIndex) throws AtlasException {
+        AtlasMapping def = getMappingDefinition();
+        getAtlasMappingHandler().removeMappingField(def, dsType, mappingId, fieldIndex);
+        setMappingDefinition(def);
+        persist();
+    }
+
+    /**
      * Loads ADM Archive from an exploded directory.
      * @param dir directory path.
      * @throws AtlasException unexpected error
--- a/lib/service/src/main/java/io/atlasmap/service/MappingService.java
+++ b/lib/service/src/main/java/io/atlasmap/service/MappingService.java
@@ -102,6 +102,7 @@ public class MappingService extends BaseAtlasService {
         return null;
     }
 
+    // TODO - move references to AtlasMappingHandler
     /**
      * Return the ordinal position of the mapped field with the specified index property.
      *
@@ -121,6 +122,7 @@ public class MappingService extends BaseAtlasService {
         return ordinalPosition;
     }
 
+    // TODO - move references to AtlasMappingHandler
     /**
      * Return the mapping element for the specified mapping uuid.
      *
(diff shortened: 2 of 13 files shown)


lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
```diff
@@ -720,6 +720,21 @@ public void deleteDocument(DataSourceType dsType, String documentId) throws Atla
...
 
+    /**
+     * Delete the specified field from the specified mapping.
+     *
+     * @param dsType
+     * @param mappingId
+     * @param fieldIndex
+     * @throws AtlasException
+     */
+    public void deleteMappingField(DataSourceType dsType, String mappingId, Integer fieldIndex) throws AtlasException {
```
igarashitm, line 731: We should directly invoke `AtlasMappingHandler#removeMappingField()` from `MappingService`, instead of adding this to `ADMArchiveHandler`.

lib/service/src/main/java/io/atlasmap/service/MappingService.java
```diff
@@ -204,6 +220,43 @@ public Response changeMappedFieldIndex(
...
+        @Parameter(description = "Mapping Definition ID") @PathParam("mappingDefinitionId") Integer mappingDefinitionId,
+        @Parameter(description = "Mapping ID") @PathParam("mappingId") String mappingId,
+        @Parameter(description = "Source or Target Mapping") @PathParam("dataSourceType") DataSourceType dataSourceType,
+        @Parameter(description = "Field Index") @PathParam("fieldIndex") Integer fieldIndex) {
+        LOG.debug("removeFieldMappingRequest: ID: {}, mappingId: {}, dataSourceType: {}, fieldIndex: {}",
+            mappingDefinitionId, mappingId, dataSourceType, fieldIndex);
+        try {
+            ADMArchiveHandler admHandler = this.atlasService.getADMArchiveHandler(mappingDefinitionId);
+            if (admHandler != null) {
+                admHandler.deleteMappingField(dataSourceType, mappingId, fieldIndex);
```
igarashitm, line 251: This should be `admHandler.getAtlasMappingHandler().removeMappingField()`

lib/service/src/test/java/io/atlasmap/service/MappingServiceTest.java
```diff
@@ -146,4 +148,11 @@ public void testProcessMappingCustomAction() throws Exception {
         assertEquals("param foo", pmr.getMapping().getOutputField().get(0).getValue());
     }
 
+    @Test
```
igarashitm, line 151: Can you add a few assertion to verify the field is removed? you can use `MappingService#getMappingRequest()` to get the current mapping JSON.

lib/core/src/main/java/io/atlasmap/core/ADMArchiveHandler.java
pleacu: done

lib/service/src/main/java/io/atlasmap/service/MappingService.java
pleacu: done

lib/service/src/test/java/io/atlasmap/service/MappingServiceTest.java
pleacu: I attempted to address the MappingServiceTest issue.  The real issue is that the processMappingRequest function doesn't actually process the URI.  I think the correct solution here is to have a generic method for executing the REST URI Info string.  I think this will require a new issue.

igarashitm: I didn't notice earlier, but shouldn't it be `mappingService.removeMappedFieldRequest()` for this test? `mappingService.processMappingRequest()` is to execute the Mapping Preview IIRC, and basically URIInfo shouldn't be such important when you invoke Java method directly.

igarashitm: Looking at the `MappingService#removeFieldMappingRequest()`
https://github.com/atlasmap/atlasmap/pull/3999/files#diff-5f102bdf1096d7b5c32dcda1c4506365f5edb58bb73ee7fd9059b1e91703751eR241
It doesn't even have URIInfo as a parameter, so you can't pass it in I think.

pleacu: Correct - so I don't have a mapping that's been digested.  I was looking for a generic way of calling into these functions with a predefined JSON mapping.  wdyt?

igarashitm: cf. `MappingService#processMappingRequest()` is for processing Mapping Preview
https://github.com/atlasmap/atlasmap/pull/3999/files#diff-5f102bdf1096d7b5c32dcda1c4506365f5edb58bb73ee7fd9059b1e91703751eL397

pleacu: yes - I get that - the issue I'm having is that `MappingService#removeFieldMappingRequest()` will fail because the `AtlasMappinmgHandler` is never set up.  So given a test like this:
<pre>
    public void testRemoveFieldMappingRequest() throws Exception {
    	Response res = mappingService.removeFieldMappingRequest(0, "mapping.xxxxxx", DataSourceType.TARGET, 1);
        assertEquals(200, res.getStatus());
    }
</pre>
Fails - I need to register the mapping first...

igarashitm: You can use `AtlasService#importADMArchiveRequest()` just like this
https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/test/java/io/atlasmap/service/DocumentServiceTest.java#L95
Or you can use Mockito to mock up ADMArchiveHandler/AtlasMappingHandler and let them handle a dummy Mapping Definition.

igarashitm: Just `MappingService#updateMappingRequest()` might also work
https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/MappingService.java#L312

pleacu:
Generated a new PR - issues with rebasing.

PR #4015, opened 2022-05-11
fix: UI: Refactored removeFieldMapping to backend service.  Addresses…

pleacu:
… comments/ unit test.

Fixes: #3968
Used a new branch - will clean up old branches after checkin.


pleacu:
Corrected unit test - all else previously reviewed.

Issue #3969, opened 2022-04-28
[BUG] Windows workflow fails


igarashitm:
The workflows on windows fail due to JDK 11.0.15/18.0.1.1 bug
https://github.com/atlasmap/atlasmap/runs/6216934755?check_suite_focus=true
https://bugs.openjdk.java.net/browse/JDK-8285445

We'd have to ignore windows workflows for now until it gets JDK 11.0.16/18.0.2.

PR #4029, opened 2022-05-16
chore: Remove husky

igarashitm:
There's not much benefit to run UI lint as a pre-commit hook separately since both of `yarn build` and `mvn install` run lint as a part of it.
Also husky is not very monorepo friendly and has been having a problem on version update - https://github.com/atlasmap/atlasmap/pull/4025


igarashitm:
@kahboom if it's about the Experimental windows build, you can just ignore it, there's a JDK bug in windows that causes a failure (#3969). We're monitoring with the latest JDK in Experimental build to see if it's fixed.

PR #4030, opened 2022-05-16
chore: Update patternfly

igarashitm:
This time of patternfly updates caused `MutationObserver is not defined` error in the standalone test. Mocking it up solved it.


kahboom (approved):
👍

Issue #3993, opened 2022-05-05
[BUG] Import Document (inspection request) should add `DataSource` into `AtlasMapping` as well


igarashitm:
**Describe the bug**

Currently Import Document action adds into `DocumentCatalog` only, but have to add `DataSource` into the Mapping Definition as well. Maybe as a part of `ModuleService#storeDocumentMetadata()`
https://github.com/atlasmap/atlasmap/blob/main/lib/modules/csv/service/src/main/java/io/atlasmap/csv/service/CsvService.java#L117-L125

Issue #3907, opened 2022-04-07
[TASK] Generate DataSource URI at backend


igarashitm:
This has to be calculated at backend and readonly in UI
https://github.com/atlasmap/atlasmap/blob/main/ui/packages/atlasmap-core/src/models/inspect/csv-inspection.model.ts#L89-L93
Once #3885, #3750 and #1476 completes, UI no longer needs to take care of `DataSource` at all, as it's an implementation detail of the mapping definition.

PR #4031, opened 2022-05-18
fix: Import Document (inspection request) should add a DataSource int…

igarashitm:
…o AtlasMapping as well

Fixes: #3993

Generate DataSource URI at backend
Fixes: #3907


pleacu (approved):
+1 looks good.

Issue #4016, opened 2022-05-12
[BUG] Reset is broken.


pleacu:
Attempting to 
`Reset all mappings and clear all imported documents`
will cause the backend service to NPE:
<pre>
Caused by: java.lang.NullPointerException: null
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:149) ~[na:na]
	at io.atlasmap.json.service.JsonService.performDocumentInspection(JsonService.java:151) ~[atlas-json-service-2.4.0-SNAPSHOT.jar:2.4.0-SNAPSHOT]
	at io.atlasmap.service.DocumentService.getDocumentInspectionResultRequest(DocumentService.java:130) ~[atlas-service-2.4.0-SNAPSHOT.jar:2.4.0-SNAPSHOT]
</pre>
The backend service `getDocumentInspectionResultRequest` should protect itself against a null spec value.  In addition, the UI document service is using the backend service to fetch the active document catalog.  The UI `reset` command should invoke the backend to clear all documents from that catalog.

pleacu:
To reproduce - just load `mockdoc.adm` and `reset`.

PR #4034, opened 2022-05-18
fix: Correct reset - NPE in backend service.

pleacu:
Fixes: #4016

### DIFF
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
@@ -209,6 +209,28 @@ public class AtlasService extends BaseAtlasService {
         return Response.ok().entity(serialized).build();
     }
 
+    @DELETE
+    @Path("/all-documents")
+    @Produces({MediaType.APPLICATION_JSON})
+    @Operation(summary = "Delete All Documents", description = "Delete all user defined documents")
+    @ApiResponses({
+        @ApiResponse(responseCode = "204", description = "Document was not found"),
+        @ApiResponse(responseCode = "500", description = "Document access error")})
+    public Response deleteAllDocuments(
+      @Parameter(description = "Mapping Definition ID") @PathParam("mappingDefinitionId") Integer mappingDefinitionId
+      ) {
+        LOG.debug("deleteAllDocumentsRequest: {}", mappingDefinitionId);
+        try {
+            ADMArchiveHandler admHandler = getADMArchiveHandler(mappingDefinitionId);
+            admHandler.deleteAllDocuments();
+        } catch (AtlasException e) {
+            throw new WebApplicationException(
+                String.format("Failed to delete all documents %s", mappingDefinitionId),
+                e);
+        }
+        return Response.ok().build();
+    }
+
     /**
      * Delete all mapping projects including Mapping Definitions and Documents saved on the server.
      * @return empty response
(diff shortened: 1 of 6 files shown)


lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -209,6 +209,28 @@ public Response listMappingDefinitionNames(@Context UriInfo uriInfo, @QueryParam
         return Response.ok().entity(serialized).build();
     }
 
+    @DELETE
```
igarashitm, line 212: This should be in `DocumentService` with `@Path("/")`

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -351,19 +373,24 @@ public Response listMappingBuilderClasses(@Context UriInfo uriInfo) {
...
-        @ApiResponse(responseCode = "204", description = "Unable to delete all user-defined JAR files and mapping projects") })
-    public Response deleteAll() {
-        LOG.debug("deleteAll");
+        @ApiResponse(responseCode = "200", description = "All user-defined libarary JARs,documents and mapping projects were deleted successfully"),
+        @ApiResponse(responseCode = "204", description = "Unable to delete all user-defined JAR files, documents and mapping projects") })
+    public Response deleteAll(
+        @Parameter(description = "Mapping Definition ID") @PathParam("mappingDefinitionId") Integer mappingDefinitionId) {
+        LOG.debug("deleteAll {} ", mappingDefinitionId);
         getLibraryLoader().clearLibraries();
         deleteAllMappingProjects();
```
igarashitm, line 392: This `deleteAllMappingProjects()` is supposed to remove all documents, if not we should fix `deleteAllMappingProjects()` instead of introducing new REST endpoint.

pleacu: So the `AtlasService` REST endpoint `deleteAll()` is what the UI reset currently calls.  It calls `deleteAllMappingProjects()`.  So - i guess I should modify `deleteAllMappingProjects` to take a `mappingDefinitionId` param and to call the admHandler's `admHandler.deleteAllDocuments().

igarashitm: Acutally because it is deleting libraries, it should delete all projects, IOW for all existing `mappingDefinitionId`, so it wouldn't take `mappingDefinitionId` as a parameter.
Looking at the `deleteAllMappingProjects()`, it aggressively deletes everything in the persistent directory. Does it happen to be just a cache problem, possibly the `deleteAllMappingProjects()` has to clear `admHandlerMap`?

pleacu: Hey @igarashitm - you're right as always - I re-instantiated the adm handler hash map and it fixed reset.  ref:
<pre>
diff --git a/lib/service/src/main/java/io/atlasmap/service/AtlasService.java b/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
index bb8565e80..9520442c7 100644
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
@@ -231,6 +231,7 @@ public class AtlasService extends BaseAtlasService {
         }
 
         AtlasUtil.deleteDirectoryContents(mappingFolderPathFile);
+        this.admHandlerMap = new ConcurrentHashMap<>();
         return Response.ok().build();
     }
</pre>

igarashitm: Cool - I could guess because that's my bug :rofl: 
maybe `this.admHandlerMap.clear()` looks better

pleacu:
Simpler PR coming

Issue #3867, opened 2022-03-25
[RFE] Add "Delete libraries"


igarashitm:
**Describe the solution you'd like**

Add "Delete libraries" menu in the UI. Currently we only have "Reset All", which removes all mapping definitions, Documents as well as libraries. If the user want to just remove Java libraries, "Delete libraries" option is more helpful.
The REST target would be
`DELETE /v2/atlas/library`
in `AtlasService`
Note that this involves deleting stale mappings due to the library removal. This should be handled at backend side.

cf. #3093 - this is to delete an individual jar, needs more detailed control

PR #4053, opened 2022-05-25
feat: Enhancement - add `Delete user-defined Java libraries` and `Del…

pleacu:
…ete documents and mappings` to the main Atlasmap pulldown menu.

Fixes: #3866

I left as a `TODO` checking if user defined libraries exist - it is default enabled in the menu.  We don't want to make a REST
backend service call on each render.  It should set something on the config model that can be checked quickly.  Needs more thought.

### DIFF
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
@@ -410,6 +410,7 @@ public class AtlasService extends BaseAtlasService {
             LOG.warn("{} is not a directory - removing anyway", mappingFolderFile.getAbsolutePath());
         }
         AtlasUtil.deleteDirectory(mappingFolderFile);
+        admHandlerMap.clear();
         return Response.ok().build();
     }
 
--- a/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
+++ b/lib/service/src/main/java/io/atlasmap/service/AtlasService.java
@@ -410,7 +410,7 @@ public class AtlasService extends BaseAtlasService {
             LOG.warn("{} is not a directory - removing anyway", mappingFolderFile.getAbsolutePath());
         }
         AtlasUtil.deleteDirectory(mappingFolderFile);
-        admHandlerMap.clear();
+        admHandlerMap.remove(mappingDefinitionId);
         return Response.ok().build();
     }
 
--- a/ui/packages/atlasmap-core/src/services/document-management.service.ts
+++ b/ui/packages/atlasmap-core/src/services/document-management.service.ts
@@ -1047,21 +1047,21 @@ export class DocumentManagementService {
   }
 
   /**
-   * Delete all user-defined documents.
+   * Delete all mappings and user-defined documents.
    *
    * @returns Promise true is all documents are successfully delete, Promise
    * false otherwise.
    */
-  deleteAllDocuments(): Promise<boolean> {
+  deleteMappingProject(): Promise<boolean> {
     const url = `${this.cfg.initCfg.baseAtlasServiceUrl}project/${this.cfg.mappingDefinitionId}`;
     this.cfg.logger!.debug(
-      `Delete All Documents Request: ID=${this.cfg.mappingDefinitionId}`
+      `Delete All Documents and Mappings Request: ID=${this.cfg.mappingDefinitionId}`
     );
     return new Promise((resolve) => {
       this.api
         .delete(url)
         .then(async (response: Response) => {
-          this.cfg.logger!.debug(`Delete All Document Response: ${response}`);
+          this.cfg.logger!.debug(`Delete All Document and Mappings Response: ${response}`);
           if (response.ok) {
             resolve(
               (await this.fetchDocuments()) &&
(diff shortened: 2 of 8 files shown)


igarashitm:
And the `deleteMappingProjectById()` would need to remove an ADMArchiveHandler cache in the `admHandlerMap`.

igarashitm:
Also for #3867, we are missing handling stale mappings and Documents in case it uses custom Java classes. It would not be trivial work at runtime side, better to be a separate fix.

ui/packages/atlasmap/src/impl/toolbarItems.tsx
```diff
@@ -113,16 +114,16 @@ export const AtlasmapToolbarItem: FunctionComponent<
...
         docsExist={docsExist!}
         onClick={runAndClose(onDeleteDocsAndMappings)}
         key="del-docs-mappings"
       />
     ),
-    false /* showDeleteResetToolbarItem #3867 */ && (
+    showDeleteResetToolbarItem && (
       <DeleteLibrariesToolbarItem
-        librariesExist={librariesExist!}
+        userLibrariesExist={librariesExist!}
```
igarashitm, line 126: Can you either 1) keep this disabled or 2) address stale mappings and Documents in https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/AtlasService.java#L294

igarashitm, line 126: Can you either 1) keep this disabled or 2) address stale mappings and Documents in https://github.com/atlasmap/atlasmap/blob/main/lib/service/src/main/java/io/atlasmap/service/AtlasService.java#L294

lib/service/src/main/java/io/atlasmap/service/AtlasService.java
```diff
@@ -410,6 +410,7 @@ public Response deleteMappingProjectById(
             LOG.warn("{} is not a directory - removing anyway", mappingFolderFile.getAbsolutePath());
         }
         AtlasUtil.deleteDirectory(mappingFolderFile);
+        admHandlerMap.clear();
```
igarashitm, line 413: This should be `admHandlerMap.remove(mappingDefinitionId)`, as `clear()` removes handlers for all existing `mappingDefinitionId`.

igarashitm, line 413: This should be `admHandlerMap.remove(mappingDefinitionId)`, as `clear()` removes handlers for all existing `mappingDefinitionId`.

igarashitm: Also can you add a test case to verify the mapping project is removed?

igarashitm: Also can you add a test case to verify the mapping project is removed?

ui/packages/atlasmap/src/impl/utils/document.ts
```diff
@@ -199,6 +199,16 @@ export async function getCustomClassNameOptions(): Promise<string[]> {
   });
 }
 
+/**
```
igarashitm, line 202: The `getCustomClassNameOptions()` above should be removed and the `DocumentManagementService.getLibraryClassNames()` should be used directly, as it doesn't add anything. Then this should be also in the `DocumentManagementService`, next to the `getLibraryClassNames()`.

igarashitm, line 202: The `getCustomClassNameOptions()` above should be removed and the `DocumentManagementService.getLibraryClassNames()` should be used directly, as it doesn't add anything. Then this should be also in the `DocumentManagementService`, next to the `getLibraryClassNames()`.

ui/packages/atlasmap/src/impl/toolbarItems.tsx
igarashitm: https://github.com/atlasmap/atlasmap/pull/4053#issuecomment-1137273863

igarashitm: https://github.com/atlasmap/atlasmap/pull/4053#issuecomment-1137273863

ui/packages/atlasmap-core/src/services/document-management.service.ts
```diff
@@ -1046,6 +1046,40 @@ export class DocumentManagementService {
     });
   }
 
+  /**
+   * Delete all user-defined documents.
+   *
+   * @returns Promise true is all documents are successfully delete, Promise
+   * false otherwise.
+   */
+  deleteAllDocuments(): Promise<boolean> {
```
igarashitm, line 1055: Maybe `deleteAllDocumentsAndMappings()` or `deleteMappingProject()`, and state that it will delete mappings as well as Documents in the comment?

igarashitm (approved):
LGTM other than the method name and comment

igarashitm:
Oops almost forgot, can you also squash the commits?

pleacu:
ref https://github.com/atlasmap/atlasmap/pull/4054

Issue #4320, opened 2022-09-07
[BUG] AtlasMap is stuck after importing CSV on latest Google Chrome 105


mkralik3:
**Describe the bug**
When I import CSV file into AtlasMap version 2.3.18, the UI is stuck on loading page. This happens only in Google Chrome. After that, refresh UI doesn't help and it is still stuck.
Also when I have already imported CSV via Firefox and then try open AtlasMap in Chrome, it is stuck until I delete CSV from in Firefox.

See video: 
https://user-images.githubusercontent.com/16251792/188913385-efebc87a-81df-41cb-a59e-57d52e71cce3.mp4

**Environment:**
Fedora 36
This issue happens only in Google Chrome v  105.0.5195.52 .   ( on version 102.0.5005.61 works )
No issue with Firefox 104.0

xref: https://issues.redhat.com/browse/ENTESB-19709

igarashitm:
Root cause
https://developer.chrome.com/articles/fetch-streaming-requests/
https://github.com/sindresorhus/ky/issues/450

Since newer ky no longer provides umd module, Jest configuration needs to be updated. Also react-scripts conflicts with newer ky.
```
/home/java11/workspace/atlasmap/ui/node_modules/ky/distribution/core/Ky.js 52:53
Module parse failed: Unexpected token (52:53)
File was processed with these loaders:
 * ../../node_modules/react-scripts/node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|         afterResponse: []
|       }, options.hooks),
>       method: normalizeRequestMethod(options.method ?? this._input.method),
|       // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
|       prefixUrl: String(options.prefixUrl || ''),
```

PR #4321, opened 2022-09-08
fix: AtlasMap is stuck after importing CSV on latest Google Chrome 105

igarashitm:
Fixes: #4320

Google Chrome 105 made a breaking change and made some case fail which used to be working fine with older Chrome. https://developer.chrome.com/articles/fetch-streaming-requests/

ky made a fix for that
https://github.com/sindresorhus/ky/issues/450

But since the newer ky no longer provides umd, we had to change the Jest configuration to deal with esm. Also we needed to introduce craco to add @babel/plugin-proposal-nullish-coalescing-operator which is required to update ky.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,2 @@
 *    @igarashitm
 
-/lib/modules/csv  @rkorytkowski
-
-/ui    @riccardo-forina
-
-/ui/packages/atlasmap-core    @pleacu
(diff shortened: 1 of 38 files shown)


igarashitm:
Hi @gashcrumb , can you have a quick look for me and see if I did something stupid or not :sweat_smile:

gashcrumb (approved):
Tested locally, LGTM :+1:

PR #4547, opened 2023-02-02
chore: Update to spring-boot 2.7.8

igarashitm:


grgrzybek:
I see you didn't give up - ありがとうございます！

igarashitm:
@grgrzybek wait, unfortunately it's not yet the end of the story - so there was no problem on main branch, but I still see an E2E test failure on 2.3.x branch... it's definitely showing a different behavior, but smells like a timing issue AtlasMap UI potentially has, and some async processing order might have changed. Still investigating to find a workaround - it could be just an issue only happens with selenium, then the last resort is to disable the E2E test.

Issue #4625, opened 2024-12-05
[BUG] AtlasMap not functioning after recent Chromium engine update in supported browsers


lyakot:
**Problem Description**
AtlasMap stopped functioning correctly after a recent update to browsers based on the Chromium engine (e.g., Google Chrome, Edge, Yandex Browser). The issue seems to be triggered by changes introduced in the new version of the browser.

**Steps to Reproduce**
Open AtlasMap in a browser with the updated Chromium version.

**Screenshot**
![photo_2024-12-05_18-24-42](https://github.com/user-attachments/assets/b7a721bc-044b-4e14-8d6f-617e2d683dbc)

mattrpav:
Work-around: Firefox has reported to work correctly.

PR #4630, opened 2024-12-21
 ky library update for atlasmap to work on chromium v55.2 base…

perfetimperfect24:
…d browsers

Fixes: #
[#4625 ] Ky library updated for atlasmap to work on chromium v55.2 based browsers


mattrpav (approved):
LGTM

